PR #10, opened 2021-07-12
Added initial files for eslint and workflow for github

anhtran-ccpo:
AT-6320

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,45 @@
+---
+name: Run Checks
+on:
+  push:
+    branches:
+      - main
+      - develop
+  pull_request:
+
+jobs:
+  lint:
+    name: Lint code base
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Config NodeJS dependency caching
+        uses: actions/cache@v2
+        with:
+          path: ~/.npm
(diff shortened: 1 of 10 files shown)


.github/workflows/test.yml
```diff
@@ -0,0 +1,45 @@
...
+        uses: actions/cache@v2
+        with:
+          path: ~/.npm
+          key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-node-
+      - name: Setup NodeJS
+        uses: actions/setup-node@v2
+        with:
+          node-version: "12"
```
zachclark-ccpo, line 25: Any reason we went with node js version 12?

anhtran-ccpo: This is the version we use in our env for this code

kylelaker-ccpo:
@anhtran-ccpo this looks good to me! I remember that you had some concerns about how the linting was working; have you had a chance to resolve those? If not, can we discuss what those particular questions are?

kylelaker-ccpo (approved):
Commented instead of approving. @anhtran-ccpo please see my previous comment and if everything here looks good to you, then go ahead and merge this.

anhtran-ccpo:
Yeah I had it linting the whole code base vs just the random_quote or api package and i was wondering if that was correct

PR #15, opened 2021-07-19
AT-6363: First cut of ATAT Internal Provisioning API. 

jeffsegal-ccpo:
This will power the Dashboard & Provisioning Wizard screens. Please take a close look at response codes and validate that we don't have any gaps in fulfilling the initial requirements around these APIs.

Does not yet include:
* email invitations
* comprehensive example data for mock services

### DIFF
--- a/provisioning/atat_provisioning_api.yaml
+++ b/provisioning/atat_provisioning_api.yaml
@@ -1,0 +1,498 @@
+openapi: 3.0.2
+info:
+  description: >-
+    This is a DRAFT version of the ATAT Portfolio Draft API
+  version: 1.0.0
+  title: ATAT Internal API - Portfolio Drafts
+  contact:
+    email: replaceme@ccpo.mil
+tags:
+  - name: portfolioDraft
+    description: >-
+      Portfolio Drafts represent instances of portfolio provisioning workloads which have not yet been submitted to
+      the CSP
+paths:
+  /portfolioDrafts:
+    get:
+      tags:
+        - portfolioDraft
+      description: Gets all Portfolio Drafts the caller has read access to
+      operationId: getPortfolioDrafts
+      responses:
+        '200':
+          description: Successful operation with empty response body
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/PortfolioDraft'
+        '404':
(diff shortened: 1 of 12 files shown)


kylelaker-ccpo:
Failures are in a file not related to this change. In `.github/workflows/test.yml` we need to change `VALIDATE_ALL_CODEBASE: true` to `VALIDATE_ALL_CODEBASE: false`.

kylelaker-ccpo:
That can either be hammered into this PR or a separate one can be opened.

provisioning/atat_provisioning_api.yaml
```diff
@@ -0,0 +1,524 @@
...
+            application/json:
+              schema:
+                $ref: '#/components/schemas/PortfolioDraftSummary'
+      requestBody:
+        description: An empty String or JSON body
+        content:
+          application/json:
+            schema:
+              type: string
+              example: '{}'
```
zachclark-ccpo, line 51: `example: '{}'`
Instead of using an empty example, should we link it to the PortfolioDraft schema?

provisioning/atat_provisioning_api.yaml
```diff
@@ -0,0 +1,524 @@
...
+          text/plain:
+            schema:
+              type: string
+              example: ''
```
zachclark-ccpo, line 55: `example: ''`
This is another empty example, could this also be linked to schema?

provisioning/atat_provisioning_api.yaml
```diff
@@ -0,0 +1,524 @@
...
+          description: Validation failed
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ValidationError'
+      requestBody:
+        description: An empty String or JSON body
+        content:
+          application/json:
+            example: '{}'
```
zachclark-ccpo, line 295: `example: '{}'`
Could this be linked to schema?

provisioning/atat_provisioning_api.yaml
```diff
@@ -0,0 +1,524 @@
...
+          text/plain:
+            example: ''
```
zachclark-ccpo, line 297: `example: ''`
Could this be linked to schema?

PR #18, opened 2021-07-20
Implement 'POST /portfolioDrafts'

jasonburkert-ccpo:
Use as foundation to move forward on other tickets.

Implements `POST /portfolioDrafts`

Creates a new JSON document to be stored in Cosmos DB in Portfolios collection.  JSON will contain the following keys:

- `id` - a generated v4 UUID
- `status` - CSP provisioning status of _not_started_
- `created_at` - audit field with date/timestamp
- `updated_at` - audit field with date/timestamp

Ticket: AT-6399


provisioning/src/handlers/portfolioDrafts/portfolioDrafts.post.ts
```diff
@@ -12,11 +12,11 @@ export async function postPortfolioDrafts (context: Context, req: HttpRequest):
     const { database } = await client.databases.createIfNotExists({ id: "atat" });
     const { container } = await database.containers.createIfNotExists({ id: "portfolios" });
 
-    // TODO: generate timestamps
+    let now = JSON.stringify(new Date());
```
jasonburkert-ccpo, line 15: Better as `let now = new Date().toISOString()`?  That's what underlying stringify.

provisioning/src/handlers/portfolioDrafts/post.ts
```diff
@@ -0,0 +1,31 @@
+import { Context, HttpRequest } from '@azure/functions'
+const { CosmosClient } = require("@azure/cosmos");
```
anhtran-ccpo, line 2: Is there a reason for using require over import?

jasonburkert-ccpo: I like the suggested ES6 syntax.  Makes it similar to the line above.

jeffsegal-ccpo: Nope, probably just copypasta that we should fix.

provisioning/src/handlers/portfolioDrafts/post.ts
```diff
@@ -0,0 +1,31 @@
...
+import { v4 as uuidv4 } from 'uuid';
+import dayjs from "dayjs";
+import {PortfolioSummary} from "../../lib/models/PortfolioSummary";
+import {ProvisioningStatus} from "../../lib/models/ProvisioningStatus";
+
+export async function handler (context: Context, req: HttpRequest): Promise<void> {
```
anhtran-ccpo, line 8: would this function break if we remove the req param? it is not being used

jeffsegal-ccpo: Nope we should be able to remove it.

anhtran-ccpo:
Will the shared infrastructure need to be copied over for all of apis we will have? I am wondering it it makes sense to keep it in provisioning if it is not unique.

provisioning/src/handlers/portfolioDrafts/post.ts
```diff
@@ -0,0 +1,31 @@
...
+    const connectionString = process.env["COSMOS-CONNECTION-STRING"];
+    context.log("Connecting to " + connectionString)
+    const client = new CosmosClient(connectionString);
+
+    // TODO: create these in terraform & fail fast
```
anhtran-ccpo, line 13: Are these TODOs still active?

jeffsegal-ccpo: That particular one is half-active I guess. The DB and Container are in Terraform, but we do still want to fail fast if this code executes and the underlying resources aren't there.

jeffsegal-ccpo:
To @anhtran-ccpo 's question on shared infrastructure - you're right that we probably ought to move it into a top-level folder in the monorepo since all Applications within it would likely use it (they may also have their own specific Terraform resources as well some day).

jasonburkert-ccpo (commented):
Deployed shared infrastructure and new function app with @zachclark-ccpo in about 45 minutes.  Triggered new function with Postman, new record is in Cosmos DB portfolios collection.  Ready to merge.

image

anhtran-ccpo (approved):

zachclark-ccpo (approved):

PR #19, opened 2021-07-20
Add JSON linting support

kylelaker-ccpo:
This should resolve scenarios where we are seeing linter differences
between local and the CI pipeline (because the CI pipeline is correctly
linting JSON files whereas the npm setup is not).

Eventually we need to converge on having a shared .eslintrc.json file
for all of our functions (as much as possible) and possibly using a tool
like lerna to manage the various projects that we have in the repo;
however, for now here are the steps to follow to get _other_ projects in
sync with this PoC app config:

1. Install the `eslint-plugin-json` package as a dev dependency
```
npm install --save-dev eslint-plugin-json
```

2. Add `"plugin:json/recommended"` to the `extends` section of
`.eslintrc.json`

3. Update `package.json` to have `,.json` in the `lint` and `lint:fix`
commands.

### DIFF
--- a/poc/random_quote/.eslintrc.json
+++ b/poc/random_quote/.eslintrc.json
@@ -7,7 +7,8 @@
   },
   "extends": [
     "standard",
-    "plugin:@typescript-eslint/recommended"
+    "plugin:@typescript-eslint/recommended",
+    "plugin:json/recommended"
   ],
   "parser": "@typescript-eslint/parser",
   "parserOptions": {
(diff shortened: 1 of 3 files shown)


anhtran-ccpo (changes requested):
can you add `coverage/**` to the `ignorePatterns`? when we execute our tests, the code coverage directory gets created and running the linter will validate that too. Something I didnt see in my original PR

jasonburkert-ccpo (approved):

PR #23, opened 2021-07-21
Reconfigured to use Lerna to manage multiple npm projects and to use a singular eslintrc config

anhtran-ccpo:
Installed lerna and moved apps into packages directory to be managed
Moved eslintrc.json to the root directory and added eslintignore to the apps


.github/workflows/test.yml
```diff
@@ -28,7 +28,7 @@ jobs:
         with:
           fetch-depth: 0
       - name: Install dependencies
-        run: npm install
+        run: npm ci
```
jasonburkert-ccpo, line 31: Good change.  Was going to suggest.

anhtran-ccpo: Yeah I completely forgot I had used CI here instead of install before

anhtran-ccpo:
Closing PR and will base off of develop after https://github.com/dod-ccpo/atat-web-api/pull/18 is merged

PR #24, opened 2021-07-21
Introduced lerna package to manage multiple packages 

anhtran-ccpo:
- Added package [Lerna](https://github.com/lerna/lerna) - _A tool for managing JavaScript projects with multiple packages._
- Moved all API projects and their microservice implementations to new `./packages` directory
- Consolidated and moved file `.eslintrc.json` to the root of the repository for use as central configuration
- Added new file `package.json` to the root of the repository that uses [lerna commands](https://lerna.js.org/#commands)
- Workflows such as linting and testing can now be executed from the repository root for multiple packages

Ticket: AT-6403

PR #27, opened 2021-07-22
Implement 'GET /portfolioDrafts' 

jasonburkert-ccpo:
Adds the capability to retrieve a collection of portfolio drafts.
Implements `GET /portfolioDrafts`

Ticket: AT-6362

### DIFF
--- a/provisioning/src/handlers/portfolioDrafts/get.ts
+++ b/provisioning/src/handlers/portfolioDrafts/get.ts
@@ -1,0 +1,18 @@
+import { Context } from '@azure/functions'
+import { CosmosClient } from '@azure/cosmos'
+
+export async function handler (context: Context): Promise<void> {
+  const connectionString = process.env['COSMOS-CONNECTION-STRING']
+  context.log('Connecting to ' + connectionString)
+  const client = new CosmosClient(connectionString)
+
+  // TODO: create these in terraform & fail fast
+  const { database } = await client.databases.createIfNotExists({ id: 'atat' })
+  const { container } = await database.containers.createIfNotExists({ id: 'portfolios' })
+
+  // TODO: add a modicum of error handling
+  const results = await container.items.readAll().fetchAll();
+  context.res = {
+    body: results
+  }
+}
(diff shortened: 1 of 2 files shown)


jasonburkert-ccpo:
Output may need some adjustment.

image

provisioning/src/handlers/portfolioDrafts/get.ts
```diff
@@ -0,0 +1,18 @@
...
+  const connectionString = process.env['COSMOS-CONNECTION-STRING']
+  context.log('Connecting to ' + connectionString)
+  const client = new CosmosClient(connectionString)
+
+  // TODO: create these in terraform & fail fast
+  const { database } = await client.databases.createIfNotExists({ id: 'atat' })
+  const { container } = await database.containers.createIfNotExists({ id: 'portfolios' })
+
+  // TODO: add a modicum of error handling
+  const results = await container.items.readAll().fetchAll()
```
jeffsegal-ccpo, line 14: We need to a slight transformation here to that the response matches the spec:

```
[
  {
      "id": "283c12eb-377f-4df5-82d0-5e553bab445c",
      "created_at": "{2021} 07-22T08:57:09 068 Z AM",
      "updated_at": "{2021} 07-22T08:57:09 068 Z AM",
      "status": "not_started"
  }
]
```

Right now it's wrapped by a `resources` property and also includes the cosmos metadata that we don't want to expose.

jeffsegal-ccpo (changes requested):
We absolutely need unit tests but I'm fine with punting since we don't have much of any conditional logic in here yet. We'll get a few tickets put together to address that, validation and error handling.

provisioning/src/handlers/portfolioDrafts/get.ts
jasonburkert-ccpo: That's exactly the type feedback I was seeking.  Thanks.

zachclark-ccpo: A for loop to map each result to the ProvisioningStatus.ts model would be a good way to implement this.

anhtran-ccpo: To expand on what zach is saying 

```
const results = await container.items.readAll().fetchAll().map((data) => {
      data.id,
      data.created_at
      data.updated_at,
      data.status
     })
```

jeffsegal-ccpo: I miss the Java Streaming API....  :'(

jasonburkert-ccpo: Very helpful. Thank you.

jasonburkert-ccpo:
Closing due file relocation from #24.  Will open new PR.

PR #32, opened 2021-07-24
Use function names from API

jasonburkert-ccpo:
File `atat_provisioning_wizard_api.yaml` contains several `operationId` values.
Function names should conform to these values.

- Added all such values to serverless.yml staged in a comment ready for implementation below a TODO marker
- Replaced the two existing function names with the corresponding `operationId` values


jasonburkert-ccpo:
image

PR #37, opened 2021-07-29
add .nvmrc and specify node 14

jasonburkert-ccpo:
Allows `nvm install` and `nvm use` without params

- same version as `atat-web-ui` repo
- [AWS Lambda supports Node.js 14](https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html)

### DIFF
--- a/.nvmrc
+++ b/.nvmrc
@@ -1,0 +1,1 @@
+lts/fermium
--- a/.nvmrc
+++ b/.nvmrc
@@ -1,1 +1,1 @@
-lts/fermium
+v14


kylelaker-ccpo:
Is there a reason to refer to this as `lts/fermium`? Other tooling (like the CDK, etc) supports the actual version number. We saw this with Azure and the Serverless Framework too. I'd like be to able to just search for `14` in a few months (or however long) and just "fix" this everywhere to update to `16` (for example). Can `v14` be specified in the file?

jasonburkert-ccpo:
We can certainly do it that way (v14).  I copied this directly from the other repo and is the way it had been done there.

jasonburkert-ccpo:
@kylelaker-ccpo Updated.

kylelaker (approved):

kylelaker-ccpo (approved):

PR #38, opened 2021-07-29
Fix build and packaging issues

kylelaker-ccpo:
We were missing types for `node` and `jest` at the root level which
broke the build. Also the AWS CDK types require `DOM` to be in the `lib`
array.

This also fixes up some wonky whitespace issues.


zachclark-ccpo:
Build is working now, thanks Kyle.
From a fresh install, do:
```
lerna bootstrap
npm install
lerna run build
```
for `cdk synth` to work.

zachclark-ccpo (approved):

PR #40, opened 2021-07-29
Implement 'deletePortfolioDraft' function

jasonburkert-ccpo:
Implements `deletePortfolioDraft` function for AWS.

Deletes entire Portfolio Draft item from DynamoDB.
Follows #44 as an example and leverages responses and errors from #46.

Ticket: AT-6364


jasonburkert-ccpo:
Camelcase config mirroring 62bf70dbd5a17adae540a68a43b0abd02ab06f1a

jasonburkert-ccpo:
Scripts mirror those in `./infrastructure/package.json`

jasonburkert-ccpo:
Happy-path working.

infrastructure/lib/atat-web-api-stack.ts
```diff
@@ -93,5 +93,14 @@ export class AtatWebApiStack extends cdk.Stack {
     // Allow the POST function to read and write (since that will be necessary to add the
     // new quotes)
     table.grantReadWriteData(postPortfolioDraftsFn);
+
+    const portfolioDrafts2 = portfolioDrafts.addResource("{portfolioDraftId}");
```
kylelaker-ccpo, line 97:
```suggestion
    const portfolioDraft = portfolioDrafts.addResource("{portfolioDraftId}");
```

Maybe since once we've used the `{portfolioDraftID}` we're dealing with _one_ portfolio? We could even call these `portfolioDraftsRoute` and `portfolioDraftRoute` if that helps?

applications/portfolioDrafts/package.json
```diff
@@ -3,12 +3,16 @@
...
+    "build": "tsc",
+    "watch": "tsc -w",
+    "test": "jest",
+    "lint": "eslint --ext .js,.ts,.json -c ../../.eslintrc.json .",
+    "lint:fix": "npm run lint -- --fix"
   },
   "dependencies": {
     "@aws-sdk/client-dynamodb": "^3.23.0",
     "@aws-sdk/lib-dynamodb": "^3.23.0",
+    "@aws-sdk/util-dynamodb": "^3.23.0",
```
kylelaker-ccpo, line 15: Do we need `@aws-sdk/util-dynamodb` for something ourselves? Is this a peer dependency? I have no problem with adding it; I just want to make sure I understand for future applications

applications/portfolioDrafts/deletePortfolioDraft.ts
```diff
@@ -0,0 +1,46 @@
...
+// const PRIMARY_KEY = process.env.PRIMARY_KEY || "";
+
+/**
+ * Deletes a Portfolio Draft
+ *
+ * @param event - The DELETE request from API Gateway
+ */
+export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayProxyResult> => {
+  const portfolioDraftId = event.pathParameters?.portfolioDraftId;
+  if (portfolioDraftId == null) {
```
kylelaker-ccpo, line 16:
```suggestion
  if (!portfolioDraftId) {
```

To catch `undefined` or `""` as well?

applications/portfolioDrafts/deletePortfolioDraft.ts
```diff
@@ -0,0 +1,46 @@
...
+    return { statusCode: 400, body: "Bad Request. A Portfolio Draft ID must be specified in URL path." };
```
kylelaker-ccpo, line 17: We should talk about defining an error result type.

```
{
  "errorCode": "InvalidRequest",
  "errorMessage": "A valid portfolio draft ID must be specified"
}
```

And maybe some helper functions? Eventually, it may be useful for the front end to parse info about errors so we'll want an expandable format for that.

applications/portfolioDrafts/deletePortfolioDraft.ts
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  if (!parse(portfolioDraftId)) {
+    return { statusCode: 400, body: "Bad Request. Portfolio Draft ID must be a UUID compliant with RFC 4122." };
+  }
```
kylelaker-ccpo, line 22: I am not sure that I like exposing the implementation detail that an ID must be a valid UUID. I think we can still perform this check but maybe return a `404` and a say that a Portfolio Draft with that ID doesn't exist (because we know it won't)?

.eslintrc.json
```diff
@@ -22,7 +22,8 @@
     ],
     "ignorePatterns": ["**/*.js", "coverage/**"],
     "rules": {
-      "no-console": "warn"
+      "no-console": "warn",
+      "camelcase": "off"
```
kylelaker-ccpo, line 26: So the only place we really want to allow non-camel-case is in the models, right? Because those are required to be snake case per the API specification? Thoughts on just exempting `models/**.ts` in the `overrides` section below? Something like this?
```
{
  "files": ["models/**.ts"],
  "rules": { "camelcase": "off" }
}
```

I'd like to avoid letting snake case sneak into the rest of the app code or the infrastructure. And we can exempt single lines as needed with `eslint-disable-line camelcase` later if we need to?

kylelaker-ccpo (changes requested):
Great to see this added! Thank you! I know this is still in draft but I did add a few questions to try to get them answered earlier in the dev process if possible

infrastructure/lib/atat-web-api-stack.ts
jasonburkert-ccpo: Thanks for looking at this.  I was pulled into meetings this afternoon.

zachclark-ccpo:
Working on changing the response, because we are receiving the metadata instead of the proper `DeleteCommandOutput`
```
{
  "$metadata": {
    "httpStatusCode": 200,
    "requestId": "OSQ1GI6APMQDOA11HPSMLLCCBJVV4KQNSO5AEMVJF66Q9ASUAAJG",
    "attempts": 1,
    "totalRetryDelay": 0
  }
}
```

applications/portfolioDrafts/package.json
kylelaker-ccpo: It's a peer dependency

zachclark-ccpo: I'll make a commit to remove it, its a peer dependency of `"node_modules/@aws-sdk/lib-dynamodb"`

.eslintrc.json
jasonburkert-ccpo: I'm going to reverse this and deal with it later.   I was trying to avoid having the linter suggest too many changes related to camelCase.  But that doesn't really belong in this PR with the function impl.

jasonburkert-ccpo: Reversed with https://github.com/dod-ccpo/atat-web-api/pull/40/commits/ecf1410cd1c61bb71bffe8dbdba0dadcf228ab9b

jasonburkert-ccpo:
Working nicely at the expected route, issuing status 204 using class `NoContentResponse`.

image

applications/portfolioDrafts/deletePortfolioDraft.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  try {
+    const data = await client.send(command);
+    if (!data.Attributes) {
+      return new ErrorResponse(
+        { code: ErrorCodes.INVALID_INPUT, message: "Portfolio Draft with the given ID does not exist" },
+        ErrorStatusCode.NOT_FOUND
+      );
+    }
+    return new NoContentResponse({ message: "Successful operation" });
```
kylelaker-ccpo, line 49: Does the constructor require this? I was expecting to work like:

```
return new NoContentResponse()
```

Since there should not be a body and the status code should be 204.

jeffsegal-ccpo (changes requested):
Needs a UT

applications/portfolioDrafts/deletePortfolioDraft.ts
kylelaker-ccpo: As-is, I think this will create a _header_ that looks like:

```
Message: Successful operation
```

That probably won't break anything but I think we should just go with an empty constructor?

applications/portfolioDrafts/deletePortfolioDraft.ts
```diff
@@ -0,0 +1,56 @@
...
+  } catch (err) {
```
zachclark-ccpo, line 50:
```suggestion
  } catch (err) {
  console.log("Database error: " + err);
```
We should log the database error, to keep it consistent with the other fns.

kylelaker-ccpo: @zachclark-ccpo I agree that is correct _for now_. I'd really like to see more abstractions around our DB interactions so that we don't have to worry about "manually" logging quite so much. We should see if we can get a ticket in the next sprint to support that.

applications/portfolioDrafts/deletePortfolioDraft.ts
jeffsegal-ccpo: I realized that I didn't do the delete responses the way I had intended (200 should return the entity being deleted and 204 should return an empty body). We can fix that in another PR though. 

And to Kyle's point - an empty constructor would be preferable to passing boilerplate JSON into the constructor.

zachclark-ccpo:
> Needs a UT

Should we do a separate PR for testing framework?

infrastructure/lib/atat-web-api-stack.ts
```diff
@@ -79,22 +79,32 @@ export class AtatWebApiStack extends cdk.Stack {
...
     // createPortfolioDraft
-    const createPortfolioDraftFn = new lambdaNodejs.NodejsFunction(this, "createPortfolioDraft", {
+    const createPortfolioDraftFn = new lambdaNodejs.NodejsFunction(this, "CreatePortfolioDraftFunction", {
       entry: "applications/portfolioDrafts/createPortfolioDraft.ts",
       ...sharedFunctionProps,
     });
     portfolioDrafts.addMethod("POST", new apigw.LambdaIntegration(createPortfolioDraftFn));
     table.grantReadWriteData(createPortfolioDraftFn);
 
+    // deletePortfolioDraft
```
jeffsegal-ccpo, line 100: Can we add a TODO to reduce the boilerplate here? Perhaps we can pass a little config object into a method like this?

```
{
  myNodejsFunction,
  parentRoute,
  "POST",
  table.grantReadWriteData
}
```

Feel free to tell me that I'm a dumb Java guy but I figured we can do something that helps us avoid the dreaded copypasta that can come with a long list of named variables and calling different methods that will all compile but not behave properly.

@kylelaker-ccpo this also brings up the question of how much value we might get out of applying a UT to our stacks here.

jeffsegal-ccpo (changes requested):

jeffsegal-ccpo:
> > Needs a UT
> 
> Should we do a separate PR for testing framework?

I'd really like to avoid the whole thing where we put off testing until later. I'd much rather see a simple UT that we have to optimize later so that we're not letting the perfect be the enemy of the good.

infrastructure/lib/atat-web-api-stack.ts
kylelaker-ccpo: Absolutely! I am actually planning to work on that today/tomorrow as well. I want to have a `registerDynamoDbInteractionFunction` or something like that with a less awful name that you just pass the API GW resource, the file name, the table, and the HTTP method to and it does some magic for you. The big comment at the top of this section is basically a `// TODO: Reduce copypasta`.

As far as how to test the infra code, working on that too. We should do it. We just need to think about what parts we _actually_ want to test here.

There's also the fact that API GW can "import" an OpenAPI spec, but I don't think that's elegant in the CDK (or CloudFormation generally) today since it's not easy to reference the Lambda integrations, etc. But I am sorta playing around with that idea a bit (but not in any form that will make it to a PR soon).

kylelaker-ccpo:
> > > Needs a UT
> > 
> > 
> > Should we do a separate PR for testing framework?
> 
> I'd really like to avoid the whole thing where we put off testing until later. I'd much rather see a simple UT that we have to optimize later so that we're not letting the perfect be the enemy of the good.

I think the issue is that we may not really have a lot of the "plumbing" for tests in place yet. If that does need to be done, the plumbing needs to be in another PR and then this PR needs to be rebased to consume that plumbing. I'd love to be wrong here though

jeffsegal-ccpo:
> > > > Needs a UT
> > > 
> > > 
> > > Should we do a separate PR for testing framework?
> > 
> > 
> > I'd really like to avoid the whole thing where we put off testing until later. I'd much rather see a simple UT that we have to optimize later so that we're not letting the perfect be the enemy of the good.
> 
> I think the issue is that we may not really have a lot of the "plumbing" for tests in place yet. If that does need to be done, the plumbing needs to be in another PR and then this PR needs to be rebased to consume that plumbing. I'd love to be wrong here though

That's fair. I don't want to hold this one up if we don't even have the plumbing in place. But @jasonburkert-ccpo can you create a new ticket for A) creating said plumbing and B) applying it to all existing routes? Once that merges we'll make it a baseline requirement for all new PRs to include tests.

applications/portfolioDrafts/deletePortfolioDraft.ts
jasonburkert-ccpo: I was trying to provide the description: Successful operation specified for the 200 in the API, but was just clumsy.  Will remove.

jasonburkert-ccpo: Fixed in https://github.com/dod-ccpo/atat-web-api/pull/40/commits/b43d75a429b5d25a08a140d4018707d5828282c7

applications/portfolioDrafts/deletePortfolioDraft.ts
jasonburkert-ccpo: Fixed in https://github.com/dod-ccpo/atat-web-api/pull/40/commits/6a7390025deab72285d0bcc32a9b31cabecb6011

jeffsegal-ccpo (approved):

infrastructure/lib/atat-web-api-stack.ts
jasonburkert-ccpo: I understood that we wish to:

- reduce boilerplate in `infrastructure/lib/atat-web-api-stack.ts`
- add unit testing against software stack infrastructure

jasonburkert-ccpo:
> > > > > Needs a UT
> > > > 
> > > > 
> > > > Should we do a separate PR for testing framework?
> > > 
> > > 
> > > I'd really like to avoid the whole thing where we put off testing until later. I'd much rather see a simple UT that we have to optimize later so that we're not letting the perfect be the enemy of the good.
> > 
> > 
> > I think the issue is that we may not really have a lot of the "plumbing" for tests in place yet. If that does need to be done, the plumbing needs to be in another PR and then this PR needs to be rebased to consume that plumbing. I'd love to be wrong here though
> 
> That's fair. I don't want to hold this one up if we don't even have the plumbing in place. But @jasonburkert-ccpo can you create a new ticket for A) creating said plumbing and B) applying it to all existing routes? Once that merges we'll make it a baseline requirement for all new PRs to include tests.

Created ticket AT-6456 for unit test plumbing.

jasonburkert-ccpo:
Change requests have been addressed.  Ready for approve and merge.

applications/portfolioDrafts/deletePortfolioDraft.ts
```diff
@@ -0,0 +1,57 @@
...
+      ErrorStatusCode.BAD_REQUEST
+    );
+  }
+
+  if (event.body) {
+    return new ErrorResponse(
+      { code: ErrorCodes.INVALID_INPUT, message: "HTTP request body must be empty" },
+      ErrorStatusCode.BAD_REQUEST
+    );
+  }
```
kylelaker-ccpo, line 29: Do we actually want to _reject_ requests with a body? Or just choose to ignore them? I can see an argument either way. The specification is unclear as it doesn't define a body (where as `POST /portfolioDrafts` explicitly specifies it _must_ be empty). Should we decide on a convention here?

I like the strategy of being liberal in what we accept and conservative in what we emit; so with that since the spec doesn't specify a body schema, we should either:
- update the spec to forbid non-empty bodies
- accept and ignore the body

And we may want to consider the possibility that we _could_ actually accept parameters to the `DELETE` operation in the future.

jasonburkert-ccpo: I could be convinced that we should be liberal in what we accept.  The DELETE in the spec does not specify a requestBody.

kylelaker-ccpo (approved):
My previous question is not a blocker. I am going to approve this with the expectation that we _do_ have the conversation on request bodies soon

applications/portfolioDrafts/deletePortfolioDraft.ts
jasonburkert-ccpo: I think it's just copy-paste from `POST /portfolioDrafts`.  Removing.

jasonburkert-ccpo: Fix in https://github.com/dod-ccpo/atat-web-api/pull/40/commits/c9487c1f822a01d4b945c5165b237515be2bc6ec

PR #42, opened 2021-07-30
Rename function 'createPortfolioDraft'

jasonburkert-ccpo:
Use operationId `createPortfolioDraft` from API spec

- implementation renamed to `createPortfolioDraft.ts` and refactored
- function definition renamed to `createPortfolioDraftFn`, with possible approach for remaining operationIds


kylelaker-ccpo (approved):
Great improvements! Thank you! I like the `import` `as client`. This structure for the stack will also be really good. There is a chance that the `console.log` on success will make too much noise but I think it's very good for now

PR #43, opened 2021-07-30
Add initial CI pipeline

kylelaker-ccpo:


kylelaker-ccpo:
@jasonburkert-ccpo some of these are the same as your PR #40. My preference would be to merge them here to simplify your PR but we can also hold this until your PR is done and I could rebase. But getting this PR in sooner rather than later would allow for both your PR and #44 to benefit from these changes.

applications/portfolioDrafts/package-lock.json
```diff
@@ -1,2038 +1,8 @@
 {
 	"name": "portfoliodrafts",
 	"version": "0.1.0",
-	"lockfileVersion": 2,
+	"lockfileVersion": 1,
```
zachclark-ccpo, line 4: Is this correct? I thought we were using npm `7.19.1`?

zachclark-ccpo: Read the comment in atat-platform, this is the correct lockfileVersion.

zachclark-ccpo (approved):
This checks out, good work.

PR #25, opened 2021-07-22
Implement POST request for portfolioStep function

zachclark-ccpo:
Adds the capability to save the user input from the portfolio step (step 1) of the portfolio draft wizard.
Implements `POST /portfolioDrafts/{portfolioDraftId}/portfolio`

Ticket: AT-6368

PR #44, opened 2021-07-30
Implement 'createPorfolioStep' function

zachclark-ccpo:
Implements `createPortfolioStep` function for AWS.
Inserts `portfolio_step` and all child elements into an existing Portfolio Draft item in DynamoDB.

- Includes resource creation for `portfolioDraftId` and `portfolio` routes.
- Includes creation of and grants ReadWrite to `createPortfolioStepFn`

This functionality was completed previously in #25 for Azure.

Ticket: AT-6368

### DIFF
--- a/applications/portfolioDrafts/models/PortfolioStep.ts
+++ b/applications/portfolioDrafts/models/PortfolioStep.ts
@@ -1,4 +1,3 @@
-import { AttributeValue } from "@aws-sdk/client-dynamodb";
 export interface PortfolioStep {
   name: string;
   description: string;
--- a/applications/portfolioDrafts/models/PortfolioStep.ts
+++ b/applications/portfolioDrafts/models/PortfolioStep.ts
@@ -1,3 +1,4 @@
+import { AttributeValue } from "@aws-sdk/client-dynamodb";
 export interface PortfolioStep {
   name: string;
   description: string;
--- a/applications/portfolioDrafts/portfolio/post.ts
+++ b/applications/portfolioDrafts/portfolio/post.ts
@@ -17,21 +17,33 @@ const CLIENT = dynamodbClient;
  * @param event - The POST request from API Gateway
  */
 export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayProxyResult> => {
+  if (!event.body) {
+    return { statusCode: 400, body: "" };
+  }
   /*
     if (event.body && !JSON.parse(event.body)) {
     return { statusCode: 400, body: "Request body must be empty" };
   }
 */
-  const portfolioId = event.pathParameters?.portfolioId;
+  const portfolioId = event.pathParameters?.portfolioDraftId;
+  /*
   if (!portfolioId) {
     return { statusCode: 400, body: "invalid request, you are missing the path parameter id:" };
   }
-  let requestJSON = "";
-  // requestJSON = JSON.parse(event.body);
+  */
+
+  const requestBody = JSON.parse(event.body);
+  const pf: PortfolioStep = {
+    name: requestBody.name,
+    description: requestBody.description,
+    dod_components: requestBody.dod_components,
+    portfolio_managers: requestBody,
+  };
+
   const now = new Date().toISOString();
 
   const pf: PortfolioStep = {
-    name: "bob",
+    name: requestBody.name,
     description: "bob's app",
     dod_components: ["air_force", "army", "marine_corps", "navy", "space_force"],
     portfolio_managers: ["joe.manager@example.com", "jane.manager@example.com"],
@@ -49,6 +61,7 @@ export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayPr
     Key: {
       portfolioId,
     },
+    UpdateExpression: "set portfolioStep = portfolioStep",
     ExpressionAttributeValues: {
       portfolioStep: pf,
     },
--- a/applications/portfolioDrafts/portfolio/post.ts
+++ b/applications/portfolioDrafts/portfolio/post.ts
@@ -20,50 +20,35 @@ export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayPr
   if (!event.body) {
     return { statusCode: 400, body: "" };
   }
-  /*
-    if (event.body && !JSON.parse(event.body)) {
-    return { statusCode: 400, body: "Request body must be empty" };
-  }
-*/
+
   const portfolioId = event.pathParameters?.portfolioDraftId;
-  /*
+
   if (!portfolioId) {
     return { statusCode: 400, body: "invalid request, you are missing the path parameter id:" };
   }
-  */
 
   const requestBody = JSON.parse(event.body);
   const pf: PortfolioStep = {
     name: requestBody.name,
     description: requestBody.description,
     dod_components: requestBody.dod_components,
-    portfolio_managers: requestBody,
+    portfolio_managers: requestBody.portfolio_managers,
   };
 
   const now = new Date().toISOString();
 
-  const pf: PortfolioStep = {
-    name: requestBody.name,
-    description: "bob's app",
-    dod_components: ["air_force", "army", "marine_corps", "navy", "space_force"],
-    portfolio_managers: ["joe.manager@example.com", "jane.manager@example.com"],
-  };
-
   console.log(pf);
-
-  const putCommand = new PutCommand({
-    TableName: TABLE_NAME,
-    Item: pf,
-  });
-
   const updateCommand = new UpdateCommand({
     TableName: TABLE_NAME,
     Key: {
-      portfolioId,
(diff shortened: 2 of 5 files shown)


zachclark-ccpo:
working request

UpdateCommand is now working, however it looks like the default post is broken, not returning the proper ID.

applications/portfolioDrafts/portfolio/post.ts
```diff
@@ -0,0 +1,55 @@
+import { UpdateCommand } from "@aws-sdk/lib-dynamodb";
+import { APIGatewayProxyEvent, APIGatewayProxyResult } from "aws-lambda";
+import { PortfolioStep } from ".././models/PortfolioStep";
+import { dynamodbClient } from ".././utils/dynamodb";
+
+const TABLE_NAME = process.env.ATAT_TABLE_NAME;
+const CLIENT = dynamodbClient;
```
kylelaker-ccpo, line 7: Follow the way this was done in #40 with the `import`

applications/portfolioDrafts/models/PortfolioStep.ts
```diff
@@ -1,3 +1,4 @@
+import { AttributeValue } from "@aws-sdk/client-dynamodb";
```
kylelaker-ccpo, line 1:
```suggestion
```

applications/portfolioDrafts/portfolio/post.ts
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * Handles requests from the API Gateway.
+ *
+ * @param event - The POST request from API Gateway
+ */
+export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayProxyResult> => {
+  if (!event.body) {
+    return { statusCode: 400, body: "" };
```
kylelaker-ccpo, line 16: Needs a real body

applications/portfolioDrafts/portfolio/post.ts
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  const portfolioId = event.pathParameters?.portfolioDraftId;
+
+  if (!portfolioId) {
+    return { statusCode: 400, body: "invalid request, you are missing the path parameter id:" };
+  }
```
kylelaker-ccpo, line 23: Is this even possible? Or can we assert that this is always there? Is it possible to omit a path parameter?

infrastructure/lib/atat-web-api-stack.ts
```diff
@@ -85,6 +87,16 @@ export class AtatWebApiStack extends cdk.Stack {
     // Prevent the GET function from being able to write to DynamoDB (it doesn't need to)
     table.grantReadData(getPortfolioDraftsFn);
 
+    // createPortfolioStep
```
kylelaker-ccpo, line 90:
```suggestion
    // operationId: createPortfolioStep
```

infrastructure/lib/atat-web-api-stack.ts
```diff
@@ -85,6 +87,16 @@ export class AtatWebApiStack extends cdk.Stack {
...
+    const createPortfolioStepFn = new lambdaNodejs.NodejsFunction(this, "CreatePortfolioSteptFunction", {
+      entry: "applications/portfolioDrafts/portfolio/post.ts",
+      ...sharedFunctionProps,
+    });
+    portfolio.addMethod("POST", new apigw.LambdaIntegration(createPortfolioStepFn));
+
+    // Allow the POST function to read and write (since that will be necessary to add the
+    // new quotes)
```
kylelaker-ccpo, line 98:
```suggestion
```

We don't need all this comment every time. it was useful once. Let's yank it

kylelaker-ccpo (changes requested):

applications/portfolioDrafts/portfolio/post.ts
jasonburkert-ccpo: a fine question

applications/portfolioDrafts/portfolio/post.ts
jasonburkert-ccpo, line 16:
```suggestion
    return { statusCode: 400, body: "“HTTP request body must not be empty" };
```

applications/portfolioDrafts/portfolio/post.ts
jasonburkert-ccpo, line 23:
```suggestion
```

applications/portfolioDrafts/portfolio/post.ts
```diff
@@ -0,0 +1,55 @@
...
+    ExpressionAttributeNames: {
+      "#portfolioVariable": "portfolioStep",
+    },
+    ExpressionAttributeValues: {
+      ":x": pf,
+    },
+  });
+
+  try {
+    await CLIENT.send(updateCommand);
```
jasonburkert-ccpo, line 49: I think this should capture any data and return it with the 201, rather than returning our input (`pf`).
```suggestion
    await client.send(command);
```

applications/portfolioDrafts/portfolio/post.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  const requestBody = JSON.parse(event.body);
+  const pf: PortfolioStep = {
+    name: requestBody.name,
+    description: requestBody.description,
+    dod_components: requestBody.dod_components,
+    portfolio_managers: requestBody.portfolio_managers,
+  };
+
+  console.log(pf);
```
jasonburkert-ccpo, line 33:
```suggestion
```

applications/portfolioDrafts/portfolio/post.ts
jasonburkert-ccpo: I've added code suggestions to assist with this change.

applications/portfolioDrafts/portfolio/post.ts
```diff
@@ -0,0 +1,55 @@
...
+  const updateCommand = new UpdateCommand({
```
jasonburkert-ccpo, line 34: Since we're only submitting a single command to be executed by the DB client, let's simplify the name until we must differentiate.
```suggestion
  const command = new UpdateCommand({
```

infrastructure/lib/atat-web-api-stack.ts
jasonburkert-ccpo: This was originally just direct output from grepping the API yaml.

jasonburkert-ccpo (changes requested):
I tried to provide explicit code suggestions that are mostly geared toward Kyle's change requests.

kylelaker-ccpo (approved):

infrastructure/lib/atat-web-api-stack.ts
```diff
@@ -85,6 +87,13 @@ export class AtatWebApiStack extends cdk.Stack {
     // Prevent the GET function from being able to write to DynamoDB (it doesn't need to)
     table.grantReadData(getPortfolioDraftsFn);
 
+    // operationId: createPortfolioStep
+    const createPortfolioStepFn = new lambdaNodejs.NodejsFunction(this, "CreatePortfolioSteptFunction", {
+      entry: "applications/portfolioDrafts/portfolio/post.ts",
```
jasonburkert-ccpo, line 92: Can we rename this file using the operation name?  It'll be easier to find later.  It's what I started doing in the `portfolioDrafts` directory.
`git mv applications/portfolioDrafts/portfolio/post.ts applications/portfolioDrafts/portfolio/createPortfolioStep.ts`
```suggestion
      entry: "applications/portfolioDrafts/portfolio/createPortfolioStep.ts",
```

jasonburkert-ccpo (changes requested):
A couple very minor things, and a file rename.

kylelaker-ccpo (changes requested):
Please adjust to comply with the new APIs from #46

kylelaker-ccpo (changes requested):
I actually have some developer's remorse over the way the generic added in #50 works. I really wish that we had a common base class for the various API models/schemas that we define that we could use for that generic at least (`<T extends BaseDocument>`) or something like that). I want to see if this is a change we could trivially make to the spec (probably because adding fields usually isn't a breaking change) and if we can change our code accordingly.

jasonburkert-ccpo (approved):
Very nice.  Really great collaborative effort.

image

jasonburkert-ccpo:
> I actually have some developer's remorse over the way the generic added in #50 works. I really wish that we had a common base class for the various API models/schemas that we define that we could use for that generic at least (`<T extends BaseDocument>`) or something like that). I want to see if this is a change we could trivially make to the spec (probably because adding fields usually isn't a breaking change) and if we can change our code accordingly.

If you still feel the same in the morning, let's address in a new PR.

This impl is looking much improved, addresses all other comments from today, and is functional so I'm going to merge.

PR #45, opened 2021-08-02
AT-6417: Update Wizard Spec to support upload flow

jeffsegal-ccpo:
As we decomposed requirements to support the Funding Step of the Portfolio Wizard, we found that the original file upload design did not support the user experience that we want (allowing users to upload files BEFORE clicking Next), so the spec has been updated to support this such that the upload step occurs before saving off form data.

### DIFF
--- a/atat_provisioning_wizard_api.yaml
+++ b/atat_provisioning_wizard_api.yaml
@@ -435,8 +435,10 @@ components:
           properties:
             size:
               type: number
+              description: file size in bytes
             name:
               type: string
+              description: name of file when originally uploaded
     PortfolioStep:
       type: object
       required:
--- a/atat_provisioning_wizard_api.yaml
+++ b/atat_provisioning_wizard_api.yaml
@@ -295,22 +295,19 @@ paths:
             example: '{}'
           text/plain:
             example: ''
-  '/taskOrderFiles/{taskOrderNumber}':
+  '/taskOrderFiles':
     post:
       tags:
         - taskOrder
       description: Uploads a Task Order PDF
       operationId: uploadTaskOrder
-      parameters:
-        - name: taskOrderNumber
-          in: path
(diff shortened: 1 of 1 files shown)


atat_provisioning_wizard_api.yaml
```diff
@@ -399,6 +426,19 @@ components:
...
+      allOf:
+        - $ref: '#/components/schemas/BaseObject'
+        - type: object
+          properties:
+            size:
+              type: number
+              description: file size in bytes
+            name:
+              type: string
+              description: name of file when originally uploaded
```
jasonburkert-ccpo, line 441: Would it be helpful to include a file hash as a property?

atat_provisioning_wizard_api.yaml
```diff
@@ -441,6 +481,9 @@ components:
         properties:
           task_order_number:
             type: string
+          task_order_file_id:
+            type: string
+            description: ID of the file associated with this task order (previously uploaded by POST /taskOrderFiles)
```
jasonburkert-ccpo, line 486: In think the word _previously_ could be misleading.

```suggestion
            description: ID of the file associated with this task order (uploaded via POST /taskOrderFiles)
```

jasonburkert-ccpo (approved):
Looks good.  See comment re: file hash.

This is the summary of changes that I see:
- Changed operation `uploadTaskOrder`
  - no longer expects taskOrderNumber as input
  - now generates a taskOrderId
  - returns JSON with file metadata
- Added operation `getTaskOrder`
- Changed operation `deleteTaskOrder`
  - now accepts taskOrderId (generated at upload) as input rather than taskOrderNumber
- Added a `BaseObject` schema
- Changed `PortfolioDraftSummary` schema to extend BaseObject
- Added `FileMetadata` schema

PR #46, opened 2021-08-02
Create an interface for responses and errors

kylelaker-ccpo:
This adds type safety around our response objects. It ensures that we provide a proper Error object that matches the API specification when returning an error response or that we provide a subclass of `BaseDocument` when returning a success response. Utilizing an `enum` for the status codes also adds type safety to ensure that we don't return a 200 when we return an error body. The less-safe abstract parent `Response` class is not exported. Possibly missing is a 30x `RedirectResponse` class but that is not currently defined in our API.

I believe that all error codes current defined in the specification and application are represented in these classes, enums, and interfaces.

### DIFF
--- a/applications/portfolioDrafts/createPortfolioDraft.ts
+++ b/applications/portfolioDrafts/createPortfolioDraft.ts
@@ -1,9 +1,11 @@
 import { PutCommand } from "@aws-sdk/lib-dynamodb";
 import { APIGatewayProxyEvent, APIGatewayProxyResult } from "aws-lambda";
 import { v4 as uuid } from "uuid";
+import { ErrorCodes } from "./models/Error";
 import { PortfolioSummary } from "./models/PortfolioSummary";
 import { ProvisioningStatus } from "./models/ProvisioningStatus";
 import { dynamodbClient as client } from "./utils/dynamodb";
+import { ErrorResponse, SuccessResponse } from "./utils/response";
 
 const TABLE_NAME = process.env.ATAT_TABLE_NAME;
 // const PRIMARY_KEY = process.env.PRIMARY_KEY || "";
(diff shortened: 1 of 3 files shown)


zachclark-ccpo:
error3
Response body isn't returning correctly.

applications/portfolioDrafts/createPortfolioDraft.ts
```diff
@@ -36,9 +41,12 @@ export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayPr
   try {
     const data = await client.send(command);
     console.log("success. created item: " + JSON.stringify(data));
-    return { statusCode: 201, body: JSON.stringify(data) };
+    return new DocumentResponse(document, SuccessStatusCode.CREATED);
```
zachclark-ccpo, line 44: Looks like this is returning a String, instead of JSON.

zachclark-ccpo:
```
"{\"id\":\"e35d5fba-6945-4b14-93d8-5cb2e49049bb\",\"created_at\":\"2021-08-02T19:46:38.475Z\",\"updated_at\":\"2021-08-02T19:46:38.475Z\",\"status\":\"not_started\"}"
```

zachclark-ccpo: This was fixed with 15afd62, here is output
```
{
  "id": "02e7d46a-5b55-4c31-8cce-59bbb9c10c53",
  "created_at": "2021-08-02T19:59:15.629Z",
  "updated_at": "2021-08-02T19:59:15.629Z",
  "status": "not_started"
}
```

zachclark-ccpo (approved):

kylelaker-ccpo:
@zachclark-ccpo before this gets merged, do you feel comfortable shepherding the remaining two open PRs for using this method to create API responses?

zachclark-ccpo:
> @zachclark-ccpo before this gets merged, do you feel comfortable shepherding the remaining two open PRs for using this method to create API responses?

Doesn't look too difficult, should be alright.

PR #48, opened 2021-08-02
Use HTTP 204 for success without body (DELETE methods)

kylelaker-ccpo:
Per the specification, HTTP status code [204 No Content](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204) should be used for successful responses without a body.

### DIFF
--- a/atat_provisioning_wizard_api.yaml
+++ b/atat_provisioning_wizard_api.yaml
@@ -88,7 +88,7 @@ paths:
           schema:
             type: string
       responses:
-        '200':
+        '204':
           description: Successful operation
         '404':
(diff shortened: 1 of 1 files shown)


jasonburkert-ccpo:
[200 OK](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/200) notes...

> The successful result of a PUT or a DELETE is often not a 200 OK but a 204 No Content (or a 201 Created when the resource is uploaded for the first time).

So this change makes sense.  There's no content we wish to display, or other location to which to navigate.

jasonburkert-ccpo (approved):

PR #52, opened 2021-08-03
Update createPortfolioStep function to add portfolioDraftId validation, fix timestamp

zachclark-ccpo:
## Validation Fix
Added a conditional expression on the `UpdateCommand` to check if the `created_at` attribute exists on the `Item`.
`created_at` was chosen because you cannot check for the PK with `attribute_exists()`.

## Timestamp Fix
Changes `updated_at` with the current time: 
```
const now = new Date().toISOString();
```

### DIFF
--- a/applications/portfolioDrafts/portfolio/createPortfolioStep.ts
+++ b/applications/portfolioDrafts/portfolio/createPortfolioStep.ts
@@ -22,12 +22,14 @@ export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayPr
   }
 
   const portfolioDraftId = event.pathParameters?.portfolioDraftId;
+
   if (!portfolioDraftId) {
     return new ErrorResponse(
       { code: ErrorCodes.INVALID_INPUT, message: "PortfolioDraftId must be specified in the URL path" },
-      ErrorStatusCode.BAD_REQUEST
+      ErrorStatusCode.NOT_FOUND
     );
   }
+
   if (!IsValidJson(event.body)) {
     return new ErrorResponse(
       { code: ErrorCodes.INVALID_INPUT, message: "Invalid request body: Invalid JSON" },
--- a/applications/portfolioDrafts/portfolio/createPortfolioStep.ts
+++ b/applications/portfolioDrafts/portfolio/createPortfolioStep.ts
@@ -56,25 +56,22 @@ export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayPr
       ":y": now,
     },
     ConditionExpression: "attribute_exists(created_at)",
-    ReturnValues: "ALL_OLD",
+    ReturnValues: "ALL_NEW",
   });
 
   try {
-    const data = await client.send(command);
-
-    /*
-    if (!data.Attributes) {
+    await client.send(command);
+    return new ApiSuccessResponse<PortfolioStep>(portfolioStep, SuccessStatusCode.CREATED);
+  } catch (error) {
+    console.log("Database error: " + error.name);
+    if (error.name === "ConditionalCheckFailedException") {
       return new ErrorResponse(
         { code: ErrorCodes.INVALID_INPUT, message: "Portfolio Draft with the given ID does not exist" },
-        ErrorStatusCode.NOT_FOUND
+        ErrorStatusCode.BAD_REQUEST
       );
     }
-    */
-    return new ApiSuccessResponse<PortfolioStep>(portfolioStep, SuccessStatusCode.CREATED);
-  } catch (err) {
-    console.log("Database error: " + err);
     return new ErrorResponse(
-      { code: ErrorCodes.OTHER, message: "Database error" },
+      { code: ErrorCodes.OTHER, message: "Database error" + error.name },
       ErrorStatusCode.INTERNAL_SERVER_ERROR
     );
   }
--- a/applications/portfolioDrafts/portfolio/createPortfolioStep.ts
+++ b/applications/portfolioDrafts/portfolio/createPortfolioStep.ts
@@ -55,13 +55,13 @@ export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayPr
     Key: {
       id: portfolioDraftId,
     },
-    UpdateExpression: "set #portfolioVariable = :x, updated_at = :y",
+    UpdateExpression: "set #portfolioVariable = :portfolio, updated_at = :now",
     ExpressionAttributeNames: {
       "#portfolioVariable": "portfolio_step",
     },
     ExpressionAttributeValues: {
-      ":x": portfolioStep,
-      ":y": now,
+      ":portfolio": portfolioStep,
+      ":now": now,
     },
     ConditionExpression: "attribute_exists(created_at)",
     ReturnValues: "ALL_NEW",
--- a/applications/portfolioDrafts/portfolio/createPortfolioStep.ts
+++ b/applications/portfolioDrafts/portfolio/createPortfolioStep.ts
@@ -79,7 +79,7 @@ export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayPr
       );
     }
     return new ErrorResponse(
-      { code: ErrorCodes.OTHER, message: "Database error" + error.name },
+      { code: ErrorCodes.OTHER, message: "Database error: " + error.name },
       ErrorStatusCode.INTERNAL_SERVER_ERROR
     );
   }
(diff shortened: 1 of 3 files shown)


zachclark-ccpo:
This is incomplete, the portfolioDraftId validation works if the requests fails, which is managed by the `try{....}catch{....}`  which is not the best way to implement this, currently working on refactoring.

applications/portfolioDrafts/portfolio/createPortfolioStep.ts
```diff
@@ -20,9 +21,28 @@ export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayPr
...
+    return new ErrorResponse(
+      { code: ErrorCodes.INVALID_INPUT, message: "Invalid request body: Invalid JSON" },
+      ErrorStatusCode.BAD_REQUEST
+    );
+  }
   const requestBody = JSON.parse(event.body);
+
+  if (!isPortfolioStep(requestBody)) {
+    return new ErrorResponse(
+      { code: ErrorCodes.INVALID_INPUT, message: "Invalid request body: Missing request attributes" },
```
kylelaker, line 41: Per the specification, you need to return an `error_map`. But maybe we handle this in a future PR with proper validation

applications/portfolioDrafts/portfolio/createPortfolioStep.ts
```diff
@@ -33,25 +53,34 @@ export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayPr
   const command = new UpdateCommand({
     TableName: TABLE_NAME,
     Key: {
-      id: portfolioId,
+      id: portfolioDraftId,
     },
-    UpdateExpression: "set #portfolioVariable = :x",
+    UpdateExpression: "set #portfolioVariable = :x, updated_at = :y",
```
kylelaker, line 58: I'd really like for this to still be readable an expression without needing to read the map later. Could we do:

```suggestion
    UpdateExpression: "set #portfolioVariable = :portfolio, updated_at = :now",
```
?

applications/portfolioDrafts/portfolio/createPortfolioStep.ts
```diff
@@ -33,25 +53,34 @@ export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayPr
...
     await client.send(command);
-  } catch (err) {
-    console.log("Database error: " + err);
+    return new ApiSuccessResponse<PortfolioStep>(portfolioStep, SuccessStatusCode.CREATED);
+  } catch (error) {
+    console.log("Database error: " + error.name);
+    if (error.name === "ConditionalCheckFailedException") {
+      return new ErrorResponse(
+        { code: ErrorCodes.INVALID_INPUT, message: "Portfolio Draft with the given ID does not exist" },
+        ErrorStatusCode.BAD_REQUEST
```
kylelaker, line 78:
```suggestion
        ErrorStatusCode.NOT_FOUND
```

Per the spec, this is a 404

applications/portfolioDrafts/portfolio/createPortfolioStep.ts
```diff
@@ -33,25 +53,34 @@ export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayPr
...
+      );
+    }
     return new ErrorResponse(
-      { code: ErrorCodes.OTHER, message: "Database error" },
+      { code: ErrorCodes.OTHER, message: "Database error" + error.name },
```
kylelaker, line 82:
```suggestion
      { code: ErrorCodes.OTHER, message: "Database error: " + error.name },
```

applications/portfolioDrafts/portfolio/createPortfolioStep.ts
```diff
@@ -20,9 +21,28 @@ export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayPr
...
-  const portfolioId = event.pathParameters?.portfolioDraftId;
-
+  const portfolioDraftId = event.pathParameters?.portfolioDraftId;
+  if (!portfolioDraftId) {
+    return new ErrorResponse(
+      { code: ErrorCodes.INVALID_INPUT, message: "PortfolioDraftId must be specified in the URL path" },
+      ErrorStatusCode.BAD_REQUEST
+    );
+  }
+  if (!IsValidJson(event.body)) {
```
kylelaker, line 31: `IsValidJson` doesn't match proper naming conventions. I'll try to remember to mention this again at it's definition

kylelaker, line 31: Could we make it so that we had a

```typescript
const requestBody = parse(event.body);
if (!requestBody) {
  // return Invalid JSON response
}
```
?

applications/portfolioDrafts/portfolio/createPortfolioStep.ts
```diff
@@ -20,9 +21,28 @@ export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayPr
...
   }
 
-  const portfolioId = event.pathParameters?.portfolioDraftId;
-
+  const portfolioDraftId = event.pathParameters?.portfolioDraftId;
+  if (!portfolioDraftId) {
+    return new ErrorResponse(
+      { code: ErrorCodes.INVALID_INPUT, message: "PortfolioDraftId must be specified in the URL path" },
+      ErrorStatusCode.BAD_REQUEST
+    );
```
kylelaker, line 29: I'd prefer that we treat this like a regular 404. The user didn't give us bad data, they navigated to the wrong URL. This response should be treated the same as if we had queried the database and there wasn't any such portfolioDraft.

The short-circuiting is kinda nice though (even if we probably don't actually need it since the later ConditionExpression should always fail for an empty input).

applications/portfolioDrafts/portfolio/createPortfolioStep.ts
```diff
@@ -20,9 +21,28 @@ export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayPr
...
   const requestBody = JSON.parse(event.body);
+
+  if (!isPortfolioStep(requestBody)) {
+    return new ErrorResponse(
+      { code: ErrorCodes.INVALID_INPUT, message: "Invalid request body: Missing request attributes" },
+      ErrorStatusCode.BAD_REQUEST
+    );
+  }
+  const now = new Date().toISOString();
   const portfolioStep: PortfolioStep = {
```
kylelaker, line 46: Will type script let you now do:

```suggestion
  const portfolioStep = requestBody as PortfolioStep;
```

Since it's seen us return if `!requestBody is PortfolioStep`?

kylelaker: Don't accept this comment btw. It'll be invalid bc of the next few lines.

applications/portfolioDrafts/portfolio/createPortfolioStep.ts
```diff
@@ -33,25 +53,34 @@ export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayPr
...
+    UpdateExpression: "set #portfolioVariable = :x, updated_at = :y",
     ExpressionAttributeNames: {
       "#portfolioVariable": "portfolio_step",
     },
     ExpressionAttributeValues: {
       ":x": portfolioStep,
+      ":y": now,
     },
+    ConditionExpression: "attribute_exists(created_at)",
+    ReturnValues: "ALL_NEW",
```
kylelaker, line 67: Do we need this? We never use the result of `client.send(command)`.

applications/portfolioDrafts/utils/validation.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Check whether a given string is valid JSON
+ *
+ * @param str - The String to check
+ * @returns true if JSON.parse is successful, otherwise returns false
+ */
+export function IsValidJson(str: string) {
```
kylelaker, line 7:
```suggestion
export function isValidJson(str: string) {
```

kylelaker: Would be cool if this could just return a valid object when we got valid JSON. But I'd actually like to follow a model where our validators either return a valid object or they raise an exception (with `error_map` if needed).

kylelaker (changes requested):

applications/portfolioDrafts/utils/validation.ts
jeffsegal-ccpo: AT-6416 is meant to provide some guide posts around this, but I do agree that a model with raised exceptions which are caught and handled in a standardized way would be nice. Or at least, that's what I would do in Java.

zachclark-ccpo: This should be included in a separate validation PR, we need to revamp how we are handling these errors.

applications/portfolioDrafts/portfolio/createPortfolioStep.ts
zachclark-ccpo: Not sure what you mean on this comment,

kylelaker-ccpo (approved):
Made some minor changes to resolve lingering comments. Thank you for your work on this for handling so many of my requests!

PR #53, opened 2021-08-04
fix misspellings in API response descriptions

jasonburkert-ccpo:


kylelaker-ccpo (approved):
I'm starting to get the feeling we'll need to add a spellchecker to the pipeline :P

PR #55, opened 2021-08-04
Update Wizard spec for Funding Step

jeffsegal-ccpo:
Fixing a few issues in the Funding Step of the spec:
1) Added support for multiple CLINs per Task Order
2) Changed `funding_step.task_order_file_id` to contain the entire `FileMetadata` object and renamed it to `task_order_file` accordingly
3) Addded `funding_step.task_order_file.status` field to indicate file scanning status for uploaded task orders

Ticket: AT-6409

### DIFF
--- a/atat_provisioning_wizard_api.yaml
+++ b/atat_provisioning_wizard_api.yaml
@@ -439,6 +439,13 @@ components:
             name:
               type: string
               description: name of file when originally uploaded
+            status:
+              type: string
+              description: virus scanning status of uploaded file
+              enum:
+                - "pending"
+                - "clean"
+                - "dirty"
     PortfolioStep:
       type: object
       required:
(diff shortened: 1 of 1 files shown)


atat_provisioning_wizard_api.yaml
```diff
@@ -439,6 +439,13 @@ components:
             name:
               type: string
               description: name of file when originally uploaded
+            status:
+              type: string
+              description: virus scanning status of uploaded file
+              enum:
+                - "pending"
+                - "clean"
+                - "dirty"
```
kylelaker-ccpo, line 448:
```suggestion
            status:
              type: string
              description: status of uploaded file scans and review
              enum:
                - "pending"
                - "accepted"
                - "rejected"
```

should we make this a little more generic and avoid `clean` or `dirty`? gives us room to do more than just virus scanning

kylelaker-ccpo (changes requested):

atat_provisioning_wizard_api.yaml
kylelaker-ccpo: we could have a "rejection reason" or something like that too if we needed

kylelaker-ccpo (approved):

PR #57, opened 2021-08-06
Implement 'getPortfolioDrafts' function

jasonburkert-ccpo:
Implements function `getPortfolioDrafts`.

_Note Well:_
Function uses expensive [Scan](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-dynamodb/classes/scancommand.html) command to return all items in table.  This should be refactored in time as authenticated users are introduced and table is optimized for 1:many User-to-PortfolioDraft access.

Also, DynamoDB implementation will not work with numeric _offset_, so a different solution is required.  I expect a future API spec change to remove query parameter _offset_ from this operation.  That will likely be replaced with an alternative cursor-type approach involving DynamoDB's `LastEvaluatedKey` element and  `ExclusiveStartKey` parameter.

TODO:
- [x] Remove _Hello World_ function.  It is occupying the GET method space in the `/portfolioDrafts` route
- [x] Change API spec for operation to return _array_ of PortfolioDraftSummary schema objects
- [x] Change API spec schema object PortfolioDraftSummary to include property _num_portfolio_managers_ to support UI design of _My Portfolios_ page
- [x] Change API spec to remove status code 404.  This will not be possible.  Empty result set will return empty array.  Table should exist and be accessible or return 5xx.
- [x] Change API spec to add status code 400 for invalid query parameters
- [x] Change `models/PortfolioDraftSummary.ts` to include property _num_portfolio_managers_ per API spec
- [x] Change `applications/portfolioDrafts/createPortfolioDraft.ts` to include property _num_portfolio_managers_ with default value at PortfolioDraft creation
- [x] Function returns status code 200 with array of PortfolioDraftSummary objects
- [x] Function returns status code 400 for invalid query parameters
- [ ] Function accepts optional query param _offset_ must be integer with minimum value of 0
- [x] Function accepts optional query param _limit_ must be integer with minimum value of 1 and maximum value of 50. Defaults to 20.
- [x] Returned result set is impacted by value of _limit_ query parameter
- [ ] Returned result set is impacted by value of _offset_ query parameter

Ticket: AT-6362

### DIFF
--- a/applications/portfolioDrafts/createPortfolioDraft.ts
+++ b/applications/portfolioDrafts/createPortfolioDraft.ts
@@ -29,6 +29,7 @@ export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayPr
     created_at: now,
     updated_at: now,
     status: ProvisioningStatus.NotStarted,
+    num_portfolio_managers: 0,
   };
 
   const params = {
--- a/applications/portfolioDrafts/getPortfolioDrafts.ts
+++ b/applications/portfolioDrafts/getPortfolioDrafts.ts
@@ -1,35 +1,11 @@
-import { GetCommand, GetCommandInput } from "@aws-sdk/lib-dynamodb";
-import { APIGatewayProxyEvent, APIGatewayProxyResult } from "aws-lambda";
-import { dynamodbClient as client } from "./utils/dynamodb";
-import { ErrorCodes } from "./models/Error";
-import { ErrorResponse, ErrorStatusCode, NoContentResponse } from "./utils/response";
-
-const TABLE_NAME = process.env.ATAT_TABLE_NAME;
+import { APIGatewayProxyResult } from "aws-lambda";
+import { NoContentResponse } from "./utils/response";
 
 /**
- * Gets all Portfolio Drafts
- * TODO: to which the user has read access
- *
+ * Gets all Portfolio Drafts, TODO: "...to which the user has read access"
+ * Revisit once authentication and authorization are in place.
  * @param event - The GET request from API Gateway
  */
-export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayProxyResult> => {
-  const params: GetCommandInput = {
-    TableName: TABLE_NAME,
-    Key: {
-      id: "",
-    },
-  };
-
-  const command = new GetCommand(params);
-
-  try {
-    const data = await client.send(command);
-    return new NoContentResponse();
-  } catch (err) {
-    console.log("Database error: " + err);
-    return new ErrorResponse(
-      { code: ErrorCodes.OTHER, message: "Database error" },
-      ErrorStatusCode.INTERNAL_SERVER_ERROR
-    );
-  }
+export const handler = async (): Promise<APIGatewayProxyResult> => {
+  return new NoContentResponse();
 };
--- a/applications/portfolioDrafts/getPortfolioDrafts.ts
+++ b/applications/portfolioDrafts/getPortfolioDrafts.ts
@@ -40,7 +40,7 @@ export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayPr
 
   // Optional query param 'limit' must be integer with minimum value of 1 and maximum value of 50. Defaults to 20.
   // Limit is the number of items to return.
-  const limit: number = getIntegerOrDefault(event.queryStringParameters?.limit, 20);
+  const limit = getIntegerOrDefault(event.queryStringParameters?.limit, 20);
   console.log("query param 'limit': " + limit);
   if (limit < 1 || limit > 50) {
     console.log("INVALID query param 'limit'");
--- a/applications/portfolioDrafts/getPortfolioDrafts.ts
+++ b/applications/portfolioDrafts/getPortfolioDrafts.ts
@@ -12,15 +12,15 @@ const QUERY_PARAM_INVALID = new ErrorResponse(
 );
 
 /**
- * Convert string to integer, return defaultInt if undefined or NaN.
- * @param str - The string value to convert
- * @param defaultInt - The default integer value to return if str undefined or NaN
+ * Evaluate query string parameter which is expected to be an integer.
+ * Return defaultInt if undefined or NaN.
+ * @param qparam - The query string parameter to evaluate
+ * @param defaultInt - The default value
  */
-function getIntegerOrDefault(str: string | undefined, defaultInt: number): number {
-  if (str === undefined) return defaultInt;
-  const int = parseInt(str);
-  if (isNaN(int)) return defaultInt;
-  return int;
+function evaluateQueryParameterInteger(qparam: string | undefined, defaultInt: number): number {
+  // assert numeric
+  if (qparam?.match(/$[0-9]+^/)) return parseInt(qparam);
+  return defaultInt;
 }
 
 /**
@@ -31,7 +31,7 @@ function getIntegerOrDefault(str: string | undefined, defaultInt: number): numbe
 export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayProxyResult> => {
   // Optional query param 'limit' must be integer with minimum value of 1 and maximum value of 50. Defaults to 20.
   // Limit is the number of items to return.
-  const limit = getIntegerOrDefault(event.queryStringParameters?.limit, 20);
+  const limit = evaluateQueryParameterInteger(event.queryStringParameters?.limit, 20);
   if (limit < 1 || limit > 50) {
     return QUERY_PARAM_INVALID;
   }
--- a/applications/portfolioDrafts/getPortfolioDrafts.ts
+++ b/applications/portfolioDrafts/getPortfolioDrafts.ts
@@ -17,6 +17,11 @@ export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayPr
   // Offset is the number of items to skip before starting to collect the result set.
   const offset = event.queryStringParameters?.offset;
 
+  // Optional query param 'limit' must be integer with minimum value of 1 and maximum value of 50. Defaults to 20.
+  // Limit is the number of items to return.
+  const limit = event.queryStringParameters?.limit;
+  const defaultLimit = 20;
+
   const params: ScanCommandInput = {
     TableName: TABLE_NAME,
   };
--- a/applications/portfolioDrafts/getPortfolioDrafts.ts
+++ b/applications/portfolioDrafts/getPortfolioDrafts.ts
@@ -19,7 +19,7 @@ const QUERY_PARAM_INVALID = new ErrorResponse(
  */
 function evaluateQueryParameterInteger(qparam: string | undefined, defaultInt: number): number {
   // assert numeric
-  if (!qparam?.match(/$[0-9]+^/)) {
+  if (!qparam?.match(/^\+?([0-9]\d*)$/)) {
     return defaultInt;
   }
   return parseInt(qparam);
@@ -34,7 +34,9 @@ export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayPr
   // Optional query param 'limit' must be integer with minimum value of 1 and maximum value of 50. Defaults to 20.
   // Limit is the number of items to return.
   const limit = evaluateQueryParameterInteger(event.queryStringParameters?.limit, 20);
+  console.log("query param 'limit': ", limit);
   if (limit < 1 || limit > 50) {
+    console.log("INVALID PARAM VALUE.");
     return QUERY_PARAM_INVALID;
   }
 
--- a/applications/portfolioDrafts/getPortfolioDrafts.ts
+++ b/applications/portfolioDrafts/getPortfolioDrafts.ts
@@ -32,18 +32,14 @@ export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayPr
   // Optional query param 'offset' must be integer with minimum value of 0.
   // Offset is the number of items to skip before starting to collect the result set.
   const offset = getIntegerOrDefault(event.queryStringParameters?.offset, 0);
-  console.log("query param 'offset': " + offset);
   if (offset < 0) {
-    console.log("INVALID query param 'offset'");
     return QUERY_PARAM_INVALID;
   }
 
   // Optional query param 'limit' must be integer with minimum value of 1 and maximum value of 50. Defaults to 20.
   // Limit is the number of items to return.
   const limit = getIntegerOrDefault(event.queryStringParameters?.limit, 20);
-  console.log("query param 'limit': " + limit);
   if (limit < 1 || limit > 50) {
-    console.log("INVALID query param 'limit'");
     return QUERY_PARAM_INVALID;
   }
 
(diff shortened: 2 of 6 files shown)


applications/portfolioDrafts/getPortfolioDrafts.ts
```diff
@@ -0,0 +1,35 @@
...
+  };
+
+  // This is an expensive command and should be replaced when
+  // portfolio draft owners and authenticated users are available
+  // so that "...to which the user has read access" can be added to the implementation.
+  const command = new ScanCommand(params);
+
+  try {
+    const data = await client.send(command);
+    return new ApiSuccessResponse<PortfolioFull[]>(data.Items as PortfolioFull[], SuccessStatusCode.OK);
```
kylelaker-ccpo, line 27: I think per the API spec, this should be a `PortfolioDraftSummary`. It's possible that you can either do this by only casting to a `PortfolioDraft` (but that's probably not good enough). You might need a `ProjectionExpression` argument as part of `ScanCommandInput` that only selects the attributes of a `PortfolioDraft`

kylelaker-ccpo (changes requested):
This is great! Thank you for the explanation of why a `Scan` operation isn't ideal but is necessary. I do want to make sure we return the right object schema per the spec, so I do have some requests there. `GET /portfolioDrafts/{portfolioDraftId}` should return the `PortfolioFull` but I think `GET /portfolioDrafts` returns a `PortfolioSummary`.

If I am misreading the spec, feel free to dismiss this review.

applications/portfolioDrafts/getPortfolioDrafts.ts
zachclark-ccpo: ProjectionExpression docs: https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-dynamodb/interfaces/scancommandinput.html#projectionexpression

jasonburkert-ccpo: Had been using PortfolioFull because UI needs a count of the portfolio managers, however, recent API changes will allow an improved approach by including such a count in the summary.

Of course we need a solution to update that, but it's another matter.  Maybe a separate function triggered by `createPortfolioStep`.

jasonburkert-ccpo:
@zachclark-ccpo I resolved a conflict just now caused by Jeff's API change.  In short, I think we need an improved example for `PortfolioDraftSummaryEx`.

applications/portfolioDrafts/getPortfolioDrafts.ts
jasonburkert-ccpo: Models renamed in #59

jasonburkert-ccpo: API spec and the function impl has been changed to return an array of PortfolioDraftSummary.

jasonburkert-ccpo:
For the remaining TODO item, I've hit an impediment.
DynamoDB does not support numeric offset-based pagination.  Instead scans and queries return a `LastEvaluatedKey` element used as input to `ExclusiveStartKey` parameter in the next scan or query.  Will have to adjust API spec and work on a solution.

References:
- https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Query.Pagination.html
- https://stackoverflow.com/questions/42765437/pagination-in-dynamodb-using-node-js

applications/portfolioDrafts/getPortfolioDrafts.ts
```diff
@@ -0,0 +1,70 @@
...
+ * Convert string to integer, return defaultInt if undefined or NaN.
+ * @param str - The string value to convert
+ * @param defaultInt - The default integer value to return if str undefined or NaN
+ */
+function getIntegerOrDefault(str: string | undefined, defaultInt: number): number {
+  if (str === undefined) return defaultInt;
+  const int = parseInt(str);
+  if (isNaN(int)) return defaultInt;
+  return int;
+}
```
kylelaker-ccpo, line 24: This may not necessarily do what we want. It seems like this is being used primary to evaluate query string parameters. The issue here is that users pass bad data all the time and `parseInt` doesn't actually fail on non-integer inputs. For example `parseInt("123abc") == 123`. We also would accept an offset of `234.123` as being `234`. What we _may_ want to do is instead:

```suggestion
function getIntegerOrDefault(str?: string, defaultInt: number): number {
  // ?. should protect against undefined and null while asserting the input value is numeric
  if (!str?.match(/$[0-9]+^/)) {
    return defaultInt;
  }
  // NaN should not be a possible for `parseInt` with a numeric input.
  return parseInt(str);
}
```

applications/portfolioDrafts/getPortfolioDrafts.ts
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * Gets all Portfolio Drafts, TODO: "...to which the user has read access"
+ * Revisit once authentication and authorization are in place.
+ * @param event - The GET request from API Gateway
+ */
+export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayProxyResult> => {
+  // Optional query param 'offset' must be integer with minimum value of 0.
+  // Offset is the number of items to skip before starting to collect the result set.
+  const offset: number = getIntegerOrDefault(event.queryStringParameters?.offset, 0);
```
kylelaker-ccpo, line 34:
```suggestion
  const offset = getIntegerOrDefault(event.queryStringParameters?.offset, 0);
```

TypeScript should be able to infer the type here, right?

applications/portfolioDrafts/getPortfolioDrafts.ts
```diff
@@ -0,0 +1,70 @@
...
+  console.log("query param 'offset': " + offset);
+  if (offset < 0) {
+    console.log("INVALID query param 'offset'");
+    return QUERY_PARAM_INVALID;
+  }
+
+  // Optional query param 'limit' must be integer with minimum value of 1 and maximum value of 50. Defaults to 20.
+  // Limit is the number of items to return.
+  const limit: number = getIntegerOrDefault(event.queryStringParameters?.limit, 20);
```
kylelaker-ccpo, line 43:
```suggestion
  const limit = getIntegerOrDefault(event.queryStringParameters?.limit, 20);
```

TypeScript should be able to infer the type here, right?

applications/portfolioDrafts/getPortfolioDrafts.ts
jeffsegal-ccpo, line 34:
```suggestion
  const offset: number = parseInt(typeof event.queryStringParameters?.offset !== 'undefined') ?  event.queryStringParameters?.offset: "0");
```

applications/portfolioDrafts/getPortfolioDrafts.ts
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * Gets all Portfolio Drafts, TODO: "...to which the user has read access"
+ * Revisit once authentication and authorization are in place.
+ * @param event - The GET request from API Gateway
+ */
+export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayProxyResult> => {
+  // Optional query param 'offset' must be integer with minimum value of 0.
+  // Offset is the number of items to skip before starting to collect the result set.
+  const offset: number = getIntegerOrDefault(event.queryStringParameters?.offset, 0);
+  console.log("query param 'offset': " + offset);
```
kylelaker-ccpo, line 35: We probably don't want to log these verbose messages in the normal case. This will generate quite a bit of noise (especially since this is likely to be a "hot" API).

applications/portfolioDrafts/getPortfolioDrafts.ts
```diff
@@ -0,0 +1,70 @@
...
+  if (offset < 0) {
+    console.log("INVALID query param 'offset'");
```
kylelaker-ccpo, line 37: Without a value being included, this may not be a particularly useful log message.

applications/portfolioDrafts/getPortfolioDrafts.ts
jeffsegal-ccpo, line 34: Another option which relies on ECMAScript 2015 default variables:

```
function getOffset(offsetParam = "0") : number {
  return parseInt(offsetParam);
}
```

```
  const params: ScanCommandInput = {
    TableName: TABLE_NAME,
    Limit: getOffset(event.queryStringParameters?. offset),
  };
```

And finally, the same as above but with a one-liner that uses an anonymous function: 

```
    const params: ScanCommandInput = {
        TableName: TABLE_NAME,
        Limit: function (offset: string = "0") {return parseInt(offset);}(event.queryStringParameters?.offset),
    };
```

kylelaker-ccpo (changes requested):

atat_provisioning_wizard_api.yaml
```diff
@@ -16,23 +16,26 @@ paths:
...
         '200':
-          description: Successful operation with empty response body
+          description: Should return up to {limitParam} items
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/PortfolioDraftSummary'
-              examples:
-                PortfolioDraftSummaryEx:
-                  $ref: '#/components/examples/PortfolioDraftSummaryEx'
```
jeffsegal-ccpo, line 33: We still want examples here right?

jeffsegal-ccpo (changes requested):

applications/portfolioDrafts/getPortfolioDrafts.ts
jeffsegal-ccpo: This is probably too cute since it doesn't actually validate the parameter appropriately. But worth noting the existence of default function parameters.

jasonburkert-ccpo: yep

applications/portfolioDrafts/getPortfolioDrafts.ts
jasonburkert-ccpo: yep

applications/portfolioDrafts/getPortfolioDrafts.ts
jasonburkert-ccpo: Removed.

applications/portfolioDrafts/getPortfolioDrafts.ts
jasonburkert-ccpo: Removed.

atat_provisioning_wizard_api.yaml
jasonburkert-ccpo: Yes.  I seem to have clobbered them during a rebase.  I've restored the examples.

applications/portfolioDrafts/getPortfolioDrafts.ts
kylelaker-ccpo: This may still return `NaN` for totally invalid inputs. And you can't do arithmetic comparisons on `NaN`.

kylelaker-ccpo: ECMAScript 2015 should be fine since TypeScript should transpile anyway.

applications/portfolioDrafts/getPortfolioDrafts.ts
jasonburkert-ccpo: Attempting to combine this advice with @jeffsegal-ccpo observations about default params.

jasonburkert-ccpo:
Current state is non-functional.  Working to correct.

applications/portfolioDrafts/getPortfolioDrafts.ts
jasonburkert-ccpo: I like the inclusion of the regex.  Smart.
Caret (^) marks the beginning of the string and dollar sign ($) marks the end, however.

Attempting to change the type for `str` as prescribed results in _A required parameter cannot follow an optional parameter.ts(1016)_ so will retrain `string | undefined`.

applications/portfolioDrafts/getPortfolioDrafts.ts
jeffsegal-ccpo: Yeah - realized that afterwards. I have a habit of trying to do one-liners to keep my code as compact as possible but that's probably flying too close to the sun here.

applications/portfolioDrafts/getPortfolioDrafts.ts
jasonburkert-ccpo: I've incorporated this suggestion with some minor slight changes.

jasonburkert-ccpo:
All requested changes have been incorporated.  Ready for further review/commit.

applications/portfolioDrafts/getPortfolioDrafts.ts
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * Evaluate query string parameter which is expected to be an integer.
+ * Return defaultInt if undefined or NaN.
+ * @param qparam - The query string parameter to evaluate
+ * @param defaultInt - The default value
+ */
+function evaluateQueryParameterInteger(qparam: string | undefined, defaultInt: number): number {
+  // assert numeric
+  if (!qparam?.match(/^\+?([0-9]\d*)$/)) {
```
kylelaker-ccpo, line 22:
```suggestion
  if (!qparam?.match(/^\d+$/)) {
```

simplifies this to accept any string of 0 or more digits (`0-9`). it will _reject_ a leading `+`.

jasonburkert-ccpo: That's nice.

kylelaker-ccpo: @zachclark-ccpo Just want to highlight this could be something really good to get unit tests for early on as well, if you need something as a good place to get started. Especially if it got pulled into `utils/` somewhere and `export`ed. But if this is the only place we'll ever end up needing to do this, then maybe it's not appropriate for `utils/`. Something to consider.

applications/portfolioDrafts/createPortfolioDraft.ts
```diff
@@ -29,6 +29,7 @@ export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayPr
     created_at: now,
     updated_at: now,
     status: ProvisioningStatus.NotStarted,
+    num_portfolio_managers: 0,
```
kylelaker-ccpo, line 32: Something to consider for future refactoring: having a `default()` or `empty()` or something factory for these items.

kylelaker-ccpo: Actually, this feels like a computed/derviced (not stored) property

kylelaker-ccpo (approved):
Good stuff! I think that parsing and the logical validation are kept as two separate steps (1. Get the integer, 2. Make sure it's in range). Thank you!

applications/portfolioDrafts/createPortfolioDraft.ts
jasonburkert-ccpo: Yes, Jeff had suggested that perhaps another function might perform this computation.  When we have users the creating function should set it to one and add the creator as the default manager.

jasonburkert-ccpo: And that might be overkill as a separate function, but perhaps rolled into function that updates the managers.

jeffsegal-ccpo (approved):

PR #58, opened 2021-08-06
Update Wizard spec to:

jeffsegal-ccpo:
1) modify the /taskOrderFile routes to provide one GET for the binary file and another GET for the file metadata
2) simplify the data that the UI needs to supply in the task_order_file property of the Funding Step

Ticket: AT-6409

### DIFF
--- a/atat_provisioning_wizard_api.yaml
+++ b/atat_provisioning_wizard_api.yaml
@@ -348,7 +348,7 @@ paths:
     get:
       tags:
         - taskOrder
-      description: Gets a Task Order PDF
+      description: Gets File Metadata for a Task Order
       operationId: getTaskOrder
       parameters:
         - name: taskOrderId
@@ -361,10 +361,9 @@ paths:
         '200':
           description: Successful operation
           content:
-            application/pdf:
+            application/json:
               schema:
-                type: string
-                format: binary
+                $ref: '#/components/schemas/FileMetadata'
         '404':
           description: Task Order with the given ID does not exist
     delete:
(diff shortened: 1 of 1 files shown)


atat_provisioning_wizard_api.yaml
```diff
@@ -384,6 +383,29 @@ paths:
...
+      responses:
+        '200':
+          description: Successful operation
+          content:
+            application/pdf:
+              schema:
+                type: string
+                format: binary
+        '404':
+          description: Task Order with the given ID does not exist or has not completed virus scanning
```
jasonburkert-ccpo, line 408: Perhaps something less specific?  Not sure I like my own wording here.

```suggestion
          description: Task Order with the given ID does not exist or is pending processing
```

atat_provisioning_wizard_api.yaml
```diff
@@ -455,19 +477,26 @@ components:
...
               description: 'Maps form input IDs to validation error messages so that clients can display in-line errors'
+    FileMetadataSummary:
+      type: object
+      description: Metadata describing an uploaded file
+      properties:
+        id:
+          type: string
+        name:
+          type: string
+          description: name of file when originally uploaded
```
jasonburkert-ccpo, line 488:
```suggestion
          description: name of file when uploaded
```

atat_provisioning_wizard_api.yaml
```diff
@@ -455,19 +477,26 @@ components:
             error_map:
               type: object
               description: 'Maps form input IDs to validation error messages so that clients can display in-line errors'
+    FileMetadataSummary:
+      type: object
+      description: Metadata describing an uploaded file
+      properties:
+        id:
```
jasonburkert-ccpo, line 484: Does this have any prescribed format?

jasonburkert-ccpo (changes requested):
I suggest a slight change to the routes
- GET `/taskOrderFiles/{taskOrderId}/metadata`
- GET `/taskOrderFiles/{taskOrderId}/file`

Keep this as it is
- DELETE `/taskOrderFiles/{taskOrderId}`

atat_provisioning_wizard_api.yaml
jeffsegal-ccpo: I'll add that it's a UUID. In practice, it will be the S3 Object Key name, which can be anything unique within a Bucket, although we'll want to use the same UUIDs across all three Buckets.

jasonburkert-ccpo (approved):

PR #63, opened 2021-08-11
Add /taskOrderFiles

jasonburkert-ccpo:
Adds 'taskOrderFiles' parallel to existing 'portfolioDrafts'.
Configures routes in API Gateway.

Includes GET method for "Hello World" function.
image


jasonburkert-ccpo:
This was mostly a way to arrive at a decision about organization, but if this is used as a base for the ticket...
Dependencies in `package.json` could be whittled down and the scripts filled out.

jasonburkert-ccpo:
This has served its purpose for discussion and decision making.

PR #64, opened 2021-08-11
Reorganize microservices

jasonburkert-ccpo:
Reorganizes implementation files for microservices for increased clarity.
The catalyst for this was a discussion about where files for route `/taskOrderFiles` will live in support of ticket AT-6417.
This change enacts the following structure envisioned by today's discussion:

```
./infrastructure/
./microservices/package.json
./microservices/utils/
./microservices/models/
./microservices/api/portfolioDrafts/
./microservices/api/taskOrderFiles/
```

- [x] rename `applications` dir to `microservices`
- [x] add `api` dir to contain `portfolioDrafts` and `taskOrderFiles`
- [x] relocate `utils` and `models` dirs parallel to `api` dir
- [x] adjust lerna configuration
- [x] adjust existing file content (import references, etc)
- [x] ensure transpiles with tsc
- [x] adjust github workflow
- [x] adjust package scripts 
- [x] deploy
- [x] test thoroughly


jasonburkert-ccpo:
Deployed stack and tested all existing functions.   Verified behavior is as expected.

jeffsegal-ccpo (changes requested):
I would say that right now we are really deploying a single microservice, which is the entire internal API. So you could even say that this makes `atat-web-api` a monorepo. If we're in agreement there, then the only change I might make is `s/microservices/microservice` since we are not planning on adding another one here.

jasonburkert-ccpo:
Captured this direction during today's conversation.

TARGET STRUCTURE TODAY
```
./infrastructure/
./infrastructure/package.json

./packages/api/
./packages/api/package.json
./packages/api/portfolioDrafts/
./packages/api/taskOrderFiles/
./packages/api/utils/
./packages/api/models/

./packages/filescanner/
./packages/filescanner/package.json
```

FUTURE MODIFICATIONS - Move `models` to it's own shareable package
```
./packages/models/
./packages/models/package.json
```

jasonburkert-ccpo:
Will replace with new PR.

PR #71, opened 2021-08-13
Create unit tests for validation

zachclark-ccpo:
## Overview
Using `jest` to test the business logic of the API endpoints.
Moved all of the validation to `utils/validation.ts`, each microservice will use these validation fns.
`npm test` output:
```
zclark@EDC-H6K8BH2:~/test-kyle-pr/atat-web-api/applications/portfolioDrafts$ npm test

> portfoliodrafts@0.1.0 test
> jest --config=jest.config.js

 PASS  test/requestbody.test.ts
  Testing validation of request body
    ✓ should return true because request body is present (2 ms)
    ✓ should return false because request body is empty JSON (1 ms)
    ✓ should return false when body is empty object (1 ms)
    ✓ should return false because request body is an empty String (1 ms)
    ✓ should return false because request body is empty whitespace String
    ✓ should return false because request body is null
    ✓ should return false because request body JSON is invalid (2 ms)
    ✓ should return true because request body JSON is valid
  Validation tests for createPortfolioStep function
    ✓ should map body to portfolioStep object
    ✓ should fail to map body to portfolioStep object due to missing attribute (2 ms)
    ✓ should fail to map body to portfolioStep object because request body is null
  Testing validation of path parameter
    ✓ should return false because path parameter is an empty string (1 ms)
    ✓ should return true because path parameter is a valid string (1 ms)
    ✓ should return false because path parameter is undefined (1 ms)

---------------|---------|----------|---------|---------|-------------------
File           | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s 
---------------|---------|----------|---------|---------|-------------------
All files      |     100 |      100 |     100 |     100 |                   
 validation.ts |     100 |      100 |     100 |     100 |                   
---------------|---------|----------|---------|---------|-------------------
Test Suites: 1 passed, 1 total
Tests:       14 passed, 14 total
Snapshots:   0 total
Time:        2.934 s, estimated 3 s
Ran all test suites.
```
### TODO

- [x] Ensure all files are being collected with code coverage (Im certain that some of the fns outside of validation.ts).
- [x] Split up tests into smaller files
- [x] Fix lerna errors

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 *.d.ts
 node_modules
 lerna-debug.log
+applications/portfolioDrafts/coverage
 
 # CDK asset staging directory
 .cdk.staging
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,7 @@
 *.d.ts
 node_modules
 lerna-debug.log
-applications/portfolioDrafts/coverage
+applications/portfolioDrafts/coverage/
 
 # CDK asset staging directory
 .cdk.staging
--- a/applications/portfolioDrafts/jest.config.js
+++ b/applications/portfolioDrafts/jest.config.js
@@ -1,0 +1,9 @@
+module.exports = {
+    testEnvironment: 'node',
+    roots: ['<rootDir>/test'],
+    testMatch: ['**/*.test.ts'],
+    transform: {
+      '^.+\\.tsx?$': 'ts-jest'
+    }
+  };
+  
--- a/applications/portfolioDrafts/jest.config.js
+++ b/applications/portfolioDrafts/jest.config.js
@@ -4,6 +4,7 @@ module.exports = {
     testMatch: ['**/*.test.ts'],
     transform: {
       '^.+\\.tsx?$': 'ts-jest'
-    }
+    },
+    collectCoverage: true,
   };
   
--- a/applications/portfolioDrafts/portfolio/createPortfolioStep.ts
+++ b/applications/portfolioDrafts/portfolio/createPortfolioStep.ts
@@ -5,7 +5,7 @@ import { ErrorCodes } from "../models/Error";
 import { PortfolioStep } from "../models/PortfolioStep";
 import { dynamodbClient as client } from "../utils/dynamodb";
 import { ApiSuccessResponse, ErrorResponse, ErrorStatusCode, SuccessStatusCode } from "../utils/response";
-import { isPortfolioStep, isValidJson, isBodyEmpty } from "../utils/validation";
+import { isPortfolioStep, isValidJson, isBodyPresent, isPathParameterPresent } from "../utils/validation";
 
 const TABLE_NAME = process.env.ATAT_TABLE_NAME;
 const NO_SUCH_PORTFOLIO = new ErrorResponse(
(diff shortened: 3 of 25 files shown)


zachclark-ccpo:
Lerna errors:
```
npm ERR! Missing: @types/jest@^26.0.24
npm ERR! Missing: jest@^27.0.6
npm ERR! Missing: ts-jest@^27.0.4
npm ERR! 
```

applications/portfolioDrafts/utils/commands.ts
```diff
@@ -0,0 +1,27 @@
+import { UpdateCommand } from "@aws-sdk/lib-dynamodb";
+import { PortfolioStep } from "../models/PortfolioStep";
+
+export function portfolioStepCommand(
+  table: string | undefined,
```
kylelaker-ccpo, line 5: This doesn't handle undefined, it'd be invalid to pass undefined to the API, and so we should not allow undefined as a type.

applications/portfolioDrafts/utils/validation.ts
```diff
@@ -15,6 +16,30 @@ export function isValidJson(str: string): boolean {
   return true;
 }
 
+/**
+ * Check whether a given string body is empty
+ *
+ * @param body - The body of the request
+ * @returns true if the string is empty or null
+ */
+export function isBodyPresent(body: string | null): body is string {
```
kylelaker-ccpo, line 25: This is may be the wrong use of a type guard. It does accomplish the goals we have and it does in fact assert that the input is a string. But it fails to capture the additional restrictions imposed within the body.

applications/portfolioDrafts/jest.config.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
+    testEnvironment: 'node',
+    roots: ['<rootDir>/test'],
+    testMatch: ['**/*.test.ts'],
+    transform: {
+      '^.+\\.tsx?$': 'ts-jest'
+    },
+    collectCoverage: true,
```
zachclark-ccpo, line 8: There is a function invoked createPortfolioStep.ts, `const command = portfolioStepCommand(TABLE_NAME, portfolioDraftId, portfolioStep);` and some others that are not being included in the code coverage report. 
```
collectCoverageFrom: [
"**/*.ts"
]
```
yields the same results --> only validtions.ts is tested.
`collectCoverageFrom=<glob>` uses `glob` which I don't have the most experience with. Docs here: https://jestjs.io/docs/cli#--collectcoveragefromglob

zachclark-ccpo: This is now resolved.

applications/portfolioDrafts/portfolio/createPortfolioStep.ts
```diff
@@ -1,12 +1,14 @@
...
 import { APIGatewayProxyEvent, APIGatewayProxyResult } from "aws-lambda";
 import { ErrorCodes } from "../models/Error";
 import { PortfolioStep } from "../models/PortfolioStep";
 import { dynamodbClient as client } from "../utils/dynamodb";
 import { ApiSuccessResponse, ErrorResponse, ErrorStatusCode, SuccessStatusCode } from "../utils/response";
-import { isPortfolioStep, isValidJson } from "../utils/validation";
+import { isPortfolioStep, isValidJson, isBodyPresent, isPathParameterPresent } from "../utils/validation";
 
-const TABLE_NAME = process.env.ATAT_TABLE_NAME;
+const TABLE_NAME = process.env.ATAT_TABLE_NAME ?? "";
```
kylelaker-ccpo, line 11: If this isn't set, we need to totally abort. We can't work with a default value. This would be an unexpected infrastructure issue. We probably want a helper that does something like:

```
if (!process.env.ATAT_TABLE_NAME) {
  throw new Error("Oh no! The table name isn't set.")
}
```

of course with a better message and maybe a subclass of `Error`? But this really is an exceptional case that should never happen and if it does, we need to return a 500 response with a really good log message ASAP.

.gitignore
```diff
@@ -3,6 +3,7 @@
 *.d.ts
 node_modules
 lerna-debug.log
+applications/portfolioDrafts/coverage
```
kylelaker-ccpo, line 6:
```suggestion
coverage/
```

PR #81, opened 2021-08-13
Implement 'createFundingStep' function

jasonburkert-ccpo:
Implements core functionality of function.  Lacks validation and error map.

- Define function `CreateFundingStepFunction` and add to stack
- Add `funding` route for _/portfolioDrafts/{portfolioDraftId}/funding_
- Add models `FundingStep`, `FileMetadataSummary`, and `Clin`
- Add enum `CloudServiceProvider`
- Add validation function isFundingStep()
- Some PR review advice was applied to implementation of other functions

Ticket: AT-6409

### DIFF
--- a/packages/api/models/Clin.ts
+++ b/packages/api/models/Clin.ts
@@ -1,0 +1,8 @@
+export interface Clin {
+  clin_number: string;
+  idiq_clin: string;
+  total_clin_value: number;
+  obligated_funds: number;
+  pop_start_date: string;
+  pop_end_date: string;
+}
--- a/packages/api/models/CloudServiceProvider.ts
+++ b/packages/api/models/CloudServiceProvider.ts
@@ -1,4 +1,4 @@
-export const enum CloudServiceProvider {
+export enum CloudServiceProvider {
   AWS = "aws",
   AZURE = "azure",
 }
--- a/packages/api/models/CloudServiceProvider.ts
+++ b/packages/api/models/CloudServiceProvider.ts
@@ -1,4 +1,4 @@
 export const enum CloudServiceProvider {
-  AmazonWebServices = "aws",
-  MicrosoftAzure = "azure",
+  AWS = "aws",
+  AZURE = "azure",
 }
--- a/packages/api/models/CloudServiceProvider.ts
+++ b/packages/api/models/CloudServiceProvider.ts
@@ -1,0 +1,4 @@
+export const enum CloudServiceProvider {
+  AmazonWebServices = "aws",
+  MicrosoftAzure = "azure",
+}
--- a/packages/api/portfolioDrafts/funding/createFundingStep.ts
+++ b/packages/api/portfolioDrafts/funding/createFundingStep.ts
@@ -68,7 +68,7 @@ export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayPr
     if (error.name === "ConditionalCheckFailedException") {
       return NO_SUCH_PORTFOLIO_DRAFT;
     }
-    console.log("Database error: " + error.name);
+    console.error("Database error: " + error);
     return new ErrorResponse(
       { code: ErrorCodes.OTHER, message: "Database error: " + error.name },
       ErrorStatusCode.INTERNAL_SERVER_ERROR
--- a/packages/api/portfolioDrafts/funding/createFundingStep.ts
+++ b/packages/api/portfolioDrafts/funding/createFundingStep.ts
@@ -70,7 +70,7 @@ export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayPr
     }
     console.error("Database error: " + error);
     return new ErrorResponse(
-      { code: ErrorCodes.OTHER, message: "Database error: " + error.name },
+      { code: ErrorCodes.OTHER, message: "Database error" },
       ErrorStatusCode.INTERNAL_SERVER_ERROR
     );
   }
--- a/packages/api/portfolioDrafts/funding/createFundingStep.ts
+++ b/packages/api/portfolioDrafts/funding/createFundingStep.ts
@@ -1,0 +1,78 @@
+import { UpdateCommand } from "@aws-sdk/lib-dynamodb";
+import { APIGatewayProxyEvent, APIGatewayProxyResult } from "aws-lambda";
+import { ErrorCodes } from "../../models/Error";
+import { FundingStep } from "../../models/FundingStep";
+import { dynamodbClient as client } from "../../utils/dynamodb";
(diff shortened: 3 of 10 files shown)


packages/api/models/CloudServiceProvider.ts
```diff
@@ -0,0 +1,4 @@
+export const enum CloudServiceProvider {
+  AmazonWebServices = "aws",
+  MicrosoftAzure = "azure",
+}
```
kylelaker-ccpo, line 4: Enum values should usually be ALL_UPPER_CASE_WITH_UNDERSCORES. I think these names get a little long for that. Thoughts on `CloudServiceProvider.AWS` or CloudServiceProvider.AZURE`?

kylelaker-ccpo: Noticed we're not always consistent between `enum` and `const enum`. Can we add guidance on when to pick one or another to our development docs?

packages/api/utils/validation.ts
```diff
@@ -25,9 +26,25 @@ export function isValidJson(str: string): boolean {
...
+ * that the object is a valid {@link FundingStep}.
+ *
+ * @param object - The object to check
+ * @returns true if the object has all the attributes of a {@link FundingStep}
+ */
+export function isFundingStep(object: unknown): object is FundingStep {
+  if (typeof object !== "object" || object === null) {
+    return false;
+  }
+  return ["task_order_number", "task_order_file", "csp", "clins"].every((item) => item in object);
```
kylelaker-ccpo, line 49: Good for now! (In another ticket) we'll need to add support for more advanced validation than just "does the field exist".

packages/api/portfolioDrafts/funding/createFundingStep.ts
```diff
@@ -0,0 +1,78 @@
...
+    ConditionExpression: "attribute_exists(created_at)",
+  });
+
+  try {
+    await client.send(command);
+  } catch (error) {
+    if (error.name === "ConditionalCheckFailedException") {
+      return NO_SUCH_PORTFOLIO_DRAFT;
+    }
+    console.log("Database error: " + error.name);
```
kylelaker-ccpo, line 71:
```suggestion
    console.error("Database error: " + error);
```

Let's log all the error details. And it might be good to move to using an error-level print statement. Let's make a note to the team to look into this in other places.

packages/api/portfolioDrafts/funding/createFundingStep.ts
```diff
@@ -0,0 +1,78 @@
...
+    return new ErrorResponse(
+      { code: ErrorCodes.OTHER, message: "Database error: " + error.name },
```
kylelaker-ccpo, line 73:
```suggestion
      { code: ErrorCodes.OTHER, message: "Database error" },
```

I don't love that we're exposing this detail to the front end. We'll want to find a better way to correlate requests to log messages on our end.

kylelaker-ccpo (commented):
I know you're not ready for review yet but I wanted to drop this feedback in while I'm here in GitHub.

packages/api/models/CloudServiceProvider.ts
jasonburkert-ccpo: I think this would help differentiate the enums and should be applied to `ProvisioningStatus` as well.

packages/api/portfolioDrafts/funding/createFundingStep.ts
jasonburkert-ccpo: Good suggestion.

packages/api/models/CloudServiceProvider.ts
jasonburkert-ccpo: Changed `CloudServiceProvider` and `ProvisioningStatus` in 5026b90f6e1cd6004a8715cc1647ed180a72bb3e.

jasonburkert-ccpo: Investigated `enum` vs `const enum`.  Removed `const` from enum declarations in 37cc2026f05a428af571abe399bbaffbdf4ff2bd.

packages/api/portfolioDrafts/funding/createFundingStep.ts
jasonburkert-ccpo: Applied this change here and in a similar case in 25cb3e936d1006855e6f8d7af4fda6f8047458cf.

packages/api/models/Clin.ts
```diff
@@ -0,0 +1,8 @@
+export interface Clin {
+  clin_number: string;
+  idiq_clin: string;
+  total_clin_value: number;
+  obligated_funds: number;
+  pop_start_date: string;
+  pop_end_date: string;
```
kylelaker-ccpo, line 7: It feels like it might be a little awkward to have to repeat the word `clin` so often. `clin.clin_number` and `clin.total_clin_value` and so on` since it's very likely that this is either going to be accessed either as a variable named clin or as an attribute where we've referenced `.clins[]` or something like that. Perhaps `number`, `total_value`? I am unsure about `idiq_clin`. Is this going to require an API change?

We should consider adding TSDoc for the interface(s) and their members too (across the board).

kylelaker-ccpo (commented):
One last question and feel free to tell me it's either too late to fix or that we need to move the convo to an internal forum

packages/api/models/Clin.ts
jasonburkert-ccpo: This comes from an unnamed object that is a collection of named properties directly from the API spec, so I used the exact names as specified.

```
          clins:
            type: array
            items:
              type: object
              properties:
                clin_number:
                  type: string
                idiq_clin:
                  type: string
                total_clin_value:
                  type: "integer"
                obligated_funds:
                  type: "integer"
                pop_start_date:
                  type: string
                  format: date
                  example: '2021-07-01'
                pop_end_date:
                  type: string
                  format: date
                  example: '2022-07-01'
```

jasonburkert-ccpo: Perhaps code reads awkwardly.  Let's discuss outside the PR whether that is deserving of a named schema object Clin, and whether names should be adjusted.

jasonburkert-ccpo: @jeffsegal-ccpo See the above.

kylelaker-ccpo (approved):

packages/api/models/Clin.ts
jeffsegal-ccpo: Not a bad idea to have a named object but we should maybe go back to make sure we're 100% on the functional requirements, e.g. the relationship between the "parent" CLIN and the IDIQ CLIN that it maps to.

PR #80, opened 2021-08-13
Bump and fix dependencies

kylelaker-ccpo:
Some dependencies are using outdated versions and others have been
missing peer dependencies. This adds all of them.

PR #86, opened 2021-08-16
First cut of POST /taskOrderFiles

jeffsegal-ccpo:
Implementation will take a file (with Content-Type: multipart/form) and save it off to the Pending S3 Bucket. A few notes:

1) There's a hard limit per request of 10MB because we have to go through API Gateway,. This should be OK given that our average PDF size is 100KB and max is 2MB.
2) We have raised the timeout from the default of 3s to 300s to accommodate very slow connections. No need to prematurely kill our lambdas here.

Ticket: AT-6417

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -22,7 +22,7 @@
     ],
     "ignorePatterns": ["**/*.js", "coverage/**"],
     "rules": {
-      "no-console": "off"
+      "no-console": "warn"
     },
     "overrides": [
       {
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -22,7 +22,7 @@
     ],
     "ignorePatterns": ["**/*.js", "coverage/**"],
     "rules": {
-      "no-console": "warn"
+      "no-console": "off"
     },
     "overrides": [
       {
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,7 @@ tags-private.json
 
 # Editor configuration files
 .vscode/
+.idea
+
+# OSX
+.DS_Store
--- a/infrastructure/lib/atat-web-api-stack.ts
+++ b/infrastructure/lib/atat-web-api-stack.ts
@@ -156,7 +156,7 @@ export class AtatWebApiStack extends cdk.Stack {
   }
 }
 
-const addCreateTaskOrderFiles = (scope: cdk.Stack, resource: apigw.Resource) => {
+function addCreateTaskOrderFiles(scope: cdk.Stack, resource: apigw.Resource) {
   const bucket = new s3.Bucket(scope, "PendingBucket", {
     publicReadAccess: false,
     removalPolicy: cdk.RemovalPolicy.RETAIN,
@@ -175,4 +175,4 @@ const addCreateTaskOrderFiles = (scope: cdk.Stack, resource: apigw.Resource) => 
   });
   resource.addMethod("POST", new apigw.LambdaIntegration(fn));
   bucket.grantPut(fn);
-};
+}
--- a/infrastructure/lib/atat-web-api-stack.ts
+++ b/infrastructure/lib/atat-web-api-stack.ts
@@ -1,8 +1,11 @@
 import * as apigw from "@aws-cdk/aws-apigateway";
-import { UserPool } from "@aws-cdk/aws-cognito";
+import {ContentHandling} from "@aws-cdk/aws-apigateway";
+import {UserPool} from "@aws-cdk/aws-cognito";
 import * as dynamodb from "@aws-cdk/aws-dynamodb";
+import * as s3 from "@aws-cdk/aws-s3";
 import * as lambdaNodejs from "@aws-cdk/aws-lambda-nodejs";
 import * as cdk from "@aws-cdk/core";
+import {Duration} from "@aws-cdk/core";
 
 // This is a suboptimal solution to finding the relative directory to the
 // package root. This is necessary because it is possible for this file to be
@@ -52,6 +55,12 @@ export class AtatWebApiStack extends cdk.Stack {
     // Ideally we'd define different stages for dev, test, and staging. For now, a single
     // stage for everything being dev is good enough for a proof of concept
     const restApi = new apigw.RestApi(this, "AtatWebApi", {
+      binaryMediaTypes: ["multipart/form-data"],
+      deployOptions: {
+        loggingLevel: apigw.MethodLoggingLevel.INFO,
+        tracingEnabled: true,
+        dataTraceEnabled: true
+      },
       endpointConfiguration: {
         types: [apigw.EndpointType.REGIONAL],
       },
@@ -88,6 +97,7 @@ export class AtatWebApiStack extends cdk.Stack {
     // We definitely want to improve the ergonomics of this and doing so is a high priority; however, following
     // these examples and steps should be a good start to allow progress while that work is happening.
     const portfolioDrafts = restApi.root.addResource("portfolioDrafts");
+    const taskOrderFiles = restApi.root.addResource("taskOrderFiles");
     const portfolioDraftId = portfolioDrafts.addResource("{portfolioDraftId}");
     const portfolio = portfolioDraftId.addResource("portfolio");
     const funding = portfolioDraftId.addResource("funding");
@@ -147,7 +157,28 @@ export class AtatWebApiStack extends cdk.Stack {
     // TODO: getApplicationStep
     // TODO: createApplicationStep
     // TODO: submitPortfolioDraft
-    // TODO: uploadTaskOrder
+    addCreateTaskOrderFiles(this, taskOrderFiles, sharedFunctionProps);
     // TODO: deleteTaskOrder
   }
+}
+
+function addCreateTaskOrderFiles(scope: cdk.Stack,
+                                 resource: apigw.Resource) {
+  const bucket = new s3.Bucket(scope, "PendingBucket", {
+    publicReadAccess: false,
+    removalPolicy: cdk.RemovalPolicy.RETAIN,
+    autoDeleteObjects: false
+  });
+  const fn = new lambdaNodejs.NodejsFunction(scope, "CreateTaskOrderFileFunction", {
+    entry: packageRoot() + "/api/taskOrderFiles/createTaskOrderFile.ts",
+    timeout: Duration.seconds(300),
+    environment: {
+      PENDING_BUCKET: bucket.bucketName
+    },
+    bundling: {
+      externalModules: ["aws-sdk"],
+    },
+  })
+  resource.addMethod("POST", new apigw.LambdaIntegration(fn));
+  bucket.grantPut(fn);
 }
--- a/infrastructure/lib/atat-web-api-stack.ts
+++ b/infrastructure/lib/atat-web-api-stack.ts
@@ -1,11 +1,10 @@
 import * as apigw from "@aws-cdk/aws-apigateway";
-import {ContentHandling} from "@aws-cdk/aws-apigateway";
-import {UserPool} from "@aws-cdk/aws-cognito";
+import { UserPool } from "@aws-cdk/aws-cognito";
 import * as dynamodb from "@aws-cdk/aws-dynamodb";
 import * as s3 from "@aws-cdk/aws-s3";
 import * as lambdaNodejs from "@aws-cdk/aws-lambda-nodejs";
 import * as cdk from "@aws-cdk/core";
-import {Duration} from "@aws-cdk/core";
+import { Duration } from "@aws-cdk/core";
 
 // This is a suboptimal solution to finding the relative directory to the
 // package root. This is necessary because it is possible for this file to be
@@ -56,11 +55,6 @@ export class AtatWebApiStack extends cdk.Stack {
     // stage for everything being dev is good enough for a proof of concept
     const restApi = new apigw.RestApi(this, "AtatWebApi", {
       binaryMediaTypes: ["multipart/form-data"],
-      deployOptions: {
-        loggingLevel: apigw.MethodLoggingLevel.INFO,
-        tracingEnabled: true,
-        dataTraceEnabled: true
-      },
       endpointConfiguration: {
         types: [apigw.EndpointType.REGIONAL],
       },
@@ -157,28 +151,28 @@ export class AtatWebApiStack extends cdk.Stack {
     // TODO: getApplicationStep
     // TODO: createApplicationStep
     // TODO: submitPortfolioDraft
-    addCreateTaskOrderFiles(this, taskOrderFiles, sharedFunctionProps);
+    addCreateTaskOrderFiles(this, taskOrderFiles);
     // TODO: deleteTaskOrder
   }
 }
 
-function addCreateTaskOrderFiles(scope: cdk.Stack,
-                                 resource: apigw.Resource) {
+const addCreateTaskOrderFiles = (scope: cdk.Stack, resource: apigw.Resource) => {
   const bucket = new s3.Bucket(scope, "PendingBucket", {
     publicReadAccess: false,
     removalPolicy: cdk.RemovalPolicy.RETAIN,
-    autoDeleteObjects: false
+    autoDeleteObjects: false,
   });
   const fn = new lambdaNodejs.NodejsFunction(scope, "CreateTaskOrderFileFunction", {
     entry: packageRoot() + "/api/taskOrderFiles/createTaskOrderFile.ts",
     timeout: Duration.seconds(300),
     environment: {
-      PENDING_BUCKET: bucket.bucketName
+      PENDING_BUCKET: bucket.bucketName,
     },
     bundling: {
       externalModules: ["aws-sdk"],
     },
-  })
+    memorySize: 256,
+  });
   resource.addMethod("POST", new apigw.LambdaIntegration(fn));
   bucket.grantPut(fn);
-}
+};
--- a/packages/api/models/Error.ts
+++ b/packages/api/models/Error.ts
@@ -9,5 +9,5 @@ export interface Error {
 }
 
 export interface ValidationError extends Error {
-  errorMap: Record<string, any>;
+  errorMap: Record<string, never>;
 }
--- a/packages/api/models/Error.ts
+++ b/packages/api/models/Error.ts
@@ -9,5 +9,5 @@ export interface Error {
 }
 
 export interface ValidationError extends Error {
-  errorMap: Record<string, never>;
+  errorMap: Record<string, unknown>;
 }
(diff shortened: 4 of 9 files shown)


infrastructure/lib/atat-web-api-stack.ts
```diff
@@ -1,8 +1,11 @@
 import * as apigw from "@aws-cdk/aws-apigateway";
-import { UserPool } from "@aws-cdk/aws-cognito";
+import {ContentHandling} from "@aws-cdk/aws-apigateway";
```
kylelaker-ccpo, line 2:
```suggestion
```

infrastructure/lib/atat-web-api-stack.ts
```diff
@@ -1,8 +1,11 @@
...
+import {UserPool} from "@aws-cdk/aws-cognito";
 import * as dynamodb from "@aws-cdk/aws-dynamodb";
+import * as s3 from "@aws-cdk/aws-s3";
 import * as lambdaNodejs from "@aws-cdk/aws-lambda-nodejs";
 import * as cdk from "@aws-cdk/core";
+import {Duration} from "@aws-cdk/core";
```
kylelaker-ccpo, line 8:
```suggestion
import { Duration } from "@aws-cdk/core";
```

infrastructure/lib/atat-web-api-stack.ts
```diff
@@ -40,6 +43,12 @@ export class AtatWebApiStack extends cdk.Stack {
     // Ideally we'd define different stages for dev, test, and staging. For now, a single
     // stage for everything being dev is good enough for a proof of concept
     const restApi = new apigw.RestApi(this, "AtatWebApi", {
+      binaryMediaTypes: ["multipart/form-data"],
+      deployOptions: {
+        loggingLevel: apigw.MethodLoggingLevel.INFO,
+        tracingEnabled: true,
+        dataTraceEnabled: true
```
kylelaker-ccpo, line 50:
```suggestion
        dataTraceEnabled: true,
```

infrastructure/lib/atat-web-api-stack.ts
```diff
@@ -135,7 +145,28 @@ export class AtatWebApiStack extends cdk.Stack {
...
     // TODO: deleteTaskOrder
   }
 }
+
+function addCreateTaskOrderFiles(scope: cdk.Stack,
+                                 resource: apigw.Resource) {
+  const bucket = new s3.Bucket(scope, "PendingBucket", {
+    publicReadAccess: false,
+    removalPolicy: cdk.RemovalPolicy.RETAIN,
+    autoDeleteObjects: false
```
kylelaker-ccpo, line 158:
```suggestion
    autoDeleteObjects: false,
    blockPublicAccess: s3.BlockPublicAccess.BLOCK_ALL,
```

packages/api/models/FileMetadata.ts
```diff
@@ -0,0 +1,13 @@
...
+export interface FileMetadata extends BaseDocument {
+    name: string;
+    size: number;
+    status: FileScanStatus;
+}
+
+export const enum FileScanStatus {
+    Pending = "pending",
+    Accepted = "accepted",
+    Rejected = "rejected"
```
kylelaker-ccpo, line 12:
```suggestion
    PENDING = "pending",
    ACCEPTED = "accepted",
    REJECTED = "rejected"
```

packages/api/models/FileMetadata.ts
```diff
@@ -0,0 +1,13 @@
+import { BaseDocument } from "./BaseDocument";
+
+export interface FileMetadata extends BaseDocument {
+    name: string;
+    size: number;
+    status: FileScanStatus;
+}
+
+export const enum FileScanStatus {
```
kylelaker-ccpo, line 9:
```suggestion
export enum FileScanStatus {
```

packages/api/taskOrderFiles/createTaskOrderFile.ts
```diff
@@ -0,0 +1,66 @@
...
+
+    } catch (err) {
+        console.log("Unexpected error: " + err);
+        return new ErrorResponse(
+            { code: ErrorCodes.OTHER, message: "Unexpected error" },
+            ErrorStatusCode.INTERNAL_SERVER_ERROR
+        );
+    }
+
+};
```
kylelaker-ccpo, line 66:
```suggestion
};

```

packages/api/taskOrderFiles/createTaskOrderFile.ts
```diff
@@ -0,0 +1,66 @@
...
+const bucketName = process.env.PENDING_BUCKET;
+
+/**
+ * Creates a new Task Order File
+ *
+ * @param event - The POST request from API Gateway
+ */
+export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayProxyResult> => {
+    const key = uuid();
+    const client = new S3Client({});
```
kylelaker-ccpo, line 18: Eventually we may need client configs (for endpoints and such). We're going to need to move this to another function at some point. DynamoDB will need the same. Not a request. Just identifying something we need to plan for.

packages/api/taskOrderFiles/createTaskOrderFile.ts
```diff
@@ -0,0 +1,66 @@
...
+
+    try {
+        const result = await parser.parse(event);
+        if (result.files.length != 1) {
+            return new ErrorResponse(
+                { code: ErrorCodes.INVALID_INPUT, message: "Expected exactly one file." },
+                ErrorStatusCode.BAD_REQUEST
+            );
+        }
+        else {
```
kylelaker-ccpo, line 28: Else isn't needed (we just returned).

packages/api/taskOrderFiles/createTaskOrderFile.ts
```diff
@@ -0,0 +1,66 @@
...
+            let file = result.files[0];
+            if (file.contentType != 'application/pdf') {
+                return new ErrorResponse(
+                    { code: ErrorCodes.INVALID_INPUT, message: "Expected a PDF file" },
+                    ErrorStatusCode.BAD_REQUEST
+                );
+            }
+            else {
```
kylelaker-ccpo, line 36: Else isn't needed (we just returned)

packages/api/taskOrderFiles/createTaskOrderFile.ts
```diff
@@ -0,0 +1,66 @@
...
+                    ContentType: file.contentType
+                });
+                const response = await client.send(command);
+                console.log(response);
+                const now = new Date().toISOString();
+                const metadata: FileMetadata = {
+                    id: key,
+                    created_at: now,
+                    updated_at: now,
+                    status: FileScanStatus.Pending,
```
kylelaker-ccpo, line 50:
```suggestion
                    status: FileScanStatus.PENDING,
```

packages/api/taskOrderFiles/createTaskOrderFile.ts
```diff
@@ -0,0 +1,66 @@
...
+            }
+            else {
+                const command = new PutObjectCommand({
+                    Bucket: bucketName,
+                    Key: key,
+                    Body: file.content,
+                    ContentType: file.contentType
+                });
+                const response = await client.send(command);
+                console.log(response);
```
kylelaker-ccpo, line 44: We probably don't actually want to do this?

infrastructure/lib/atat-web-api-stack.ts
```diff
@@ -40,6 +43,12 @@ export class AtatWebApiStack extends cdk.Stack {
     // Ideally we'd define different stages for dev, test, and staging. For now, a single
     // stage for everything being dev is good enough for a proof of concept
     const restApi = new apigw.RestApi(this, "AtatWebApi", {
+      binaryMediaTypes: ["multipart/form-data"],
+      deployOptions: {
```
kylelaker-ccpo, line 47: Are we ready for this?

packages/api/taskOrderFiles/createTaskOrderFile.ts
```diff
@@ -0,0 +1,66 @@
...
+            }
+        }
+
+    } catch (err) {
+        console.log("Unexpected error: " + err);
+        return new ErrorResponse(
+            { code: ErrorCodes.OTHER, message: "Unexpected error" },
+            ErrorStatusCode.INTERNAL_SERVER_ERROR
+        );
+    }
```
kylelaker-ccpo, line 64: I am really starting to dislike this pattern in our handlers where we have basically the whole body in a `try` and then a `catch` that just logs and returns a 500. We probably need to look into having our handler just invoke another function so that the handler can be:

```
try {
  return doTheRealWork(event);
} catch (err) {
  console.error("Unexpected error: " + err);
  ...
}
```

This is a trivial pattern to adopt, we can do it one-function-at-a-time, and we should do it soon. This doesn't necessarily need to be rewritten but I wouldn't hate it if it was :)

.gitignore
```diff
@@ -10,3 +10,9 @@ cdk.out
 
 # Private tag data
 tags-private.json
+
+# Intellij
+.idea
+
+# OSX
+.DS_Store
```
kylelaker-ccpo, line 18:
```suggestion
.DS_Store

```

kylelaker-ccpo (changes requested):
Can you check your editor config to make sure linting is happening correctly? In VS Code, the `eslint` plugin can be configured to automatically will fix to comply with the eslintrc. Not sure about WebStorm but I have to imagine JetBrains or a plugin has a similar feature.

infrastructure/lib/atat-web-api-stack.ts
```diff
@@ -135,7 +145,28 @@ export class AtatWebApiStack extends cdk.Stack {
...
 }
+
+function addCreateTaskOrderFiles(scope: cdk.Stack,
+                                 resource: apigw.Resource) {
+  const bucket = new s3.Bucket(scope, "PendingBucket", {
+    publicReadAccess: false,
+    removalPolicy: cdk.RemovalPolicy.RETAIN,
+    autoDeleteObjects: false
+  });
+  const fn = new lambdaNodejs.NodejsFunction(scope, "CreateTaskOrderFileFunction", {
```
kylelaker-ccpo, line 160: We may need to consider bumping the memory limit for this function to 256MB. I am concerned we're close enough to 128MB that if we actually did get a 10MB PDF that we'd OOM crash. Can you peek at the CWL data at the `REPORT` lines to see if we're within 10MB of 128MB of memory utilization?

kylelaker-ccpo (changes requested):

packages/api/taskOrderFiles/createTaskOrderFile.ts
jeffsegal-ccpo: yep, removed

infrastructure/lib/atat-web-api-stack.ts
jeffsegal-ccpo: I enabled it to help with my own development but happy to leave it out for now

packages/api/taskOrderFiles/createTaskOrderFile.ts
jeffsegal-ccpo: No you're 100% right. UTs would keep us way more honest here in terms of writing small internal functions. I'll take a quick stab at it.

jeffsegal-ccpo: I won't go overboard since we have another ticket to handle this. E.g. I'd prefer a model where we throw errors rather than return them since that will allow for a cleaner control flow, but I won't do that here.

infrastructure/lib/atat-web-api-stack.ts
jeffsegal-ccpo: Probably too close for comfort. Here's one example for a 1.6 MB file: 

```
Duration: 1389.71 ms Billed Duration: 1390 ms Memory Size: 128 MB Max Memory Used: 106 MB Init Duration: 328.10 ms
```

I'll go ahead and raise it to 256MB.

infrastructure/lib/atat-web-api-stack.ts
```diff
@@ -135,7 +139,28 @@ export class AtatWebApiStack extends cdk.Stack {
     // TODO: getApplicationStep
     // TODO: createApplicationStep
     // TODO: submitPortfolioDraft
-    // TODO: uploadTaskOrder
+    addCreateTaskOrderFiles(this, taskOrderFiles);
     // TODO: deleteTaskOrder
   }
 }
+
+const addCreateTaskOrderFiles = (scope: cdk.Stack, resource: apigw.Resource) => {
```
kylelaker-ccpo, line 147: I really want to get away from this declaration syntax unless we have a good reason to use it.

```suggestion
function addCreateTaskOrderFiles(scope: cdk.Stack, resource: apigw.Resource) {
```

and then drop the semicolon at the end?

packages/api/models/Error.ts
```diff
@@ -9,5 +9,5 @@ export interface Error {
 }
 
 export interface ValidationError extends Error {
-  errorMap: Record<string, any>;
+  errorMap: Record<string, never>;
```
kylelaker-ccpo, line 12:
```suggestion
  errorMap: Record<string, never>;
```

I wasn't entirely convinced that `never` was appropriate here. I don't disagree with it's use but can we talk through why?

packages/api/taskOrderFiles/createTaskOrderFile.ts
```diff
@@ -0,0 +1,64 @@
...
+        return new ErrorResponse(
+          { code: ErrorCodes.OTHER, message: "Unexpected error" },
+          ErrorStatusCode.INTERNAL_SERVER_ERROR
+        );
+      }
+    }
+  }
+};
+
+const uploadFile = async (file: parser.MultipartFile) => {
```
kylelaker-ccpo, line 44:
```suggestion
async function uploadFile(file: parser.MultipartFile): Promise<APIGatewayProxyResult> {
```

packages/api/taskOrderFiles/createTaskOrderFile.ts
```diff
@@ -0,0 +1,64 @@
...
+import { ErrorCodes } from "../models/Error";
+
+const bucketName = process.env.PENDING_BUCKET;
+
+/**
+ * Creates a new Task Order File
+ *
+ * @param event - The POST request from API Gateway
+ */
+export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayProxyResult> => {
```
kylelaker-ccpo, line 16:
```suggestion
export async function handler(event: APIGatewayProxyEvent): Promise<APIGatewayProxyResult> {
```

packages/api/models/FileMetadata.ts
```diff
@@ -0,0 +1,13 @@
+import { BaseDocument } from "./BaseDocument";
+
+export const enum FileScanStatus {
```
kylelaker-ccpo, line 3:
```suggestion
export enum FileScanStatus {
```

.eslintrc.json
```diff
@@ -22,7 +22,7 @@
     ],
     "ignorePatterns": ["**/*.js", "coverage/**"],
     "rules": {
-      "no-console": "warn"
+      "no-console": "off"
```
kylelaker-ccpo, line 25:
```suggestion
      "no-console": "warn"
```

We've talked about a logging library. I'd be open to this if it were an override applied only to `infrastructure/` but for now, I'd like to have it keep warning at least in the app code as an indication of the fact there's still work to be done here.

packages/api/taskOrderFiles/createTaskOrderFile.ts
```diff
@@ -0,0 +1,64 @@
...
+  const result = await parser.parse(event);
+  if (result.files.length !== 1) {
+    return new ErrorResponse(
+      { code: ErrorCodes.INVALID_INPUT, message: "Expected exactly one file." },
+      ErrorStatusCode.BAD_REQUEST
+    );
+  } else {
```
kylelaker-ccpo, line 23:
```
if (condition) {
  return;
}
else {
  // stuff
}
```

Can be cleaned up to:

```
if (condition) {
  return;
}
// stuff
```

There are a few places to make this improvement here.

kylelaker-ccpo (changes requested):
There are a few things that I think either I lost or GitHub did in the first round of review. Or if these are repetitive and I'm saying things you intentionally didn't do just lmk

packages/api/taskOrderFiles/createTaskOrderFile.ts
jeffsegal-ccpo: Ah good call. Fixed that as well. Ultimately something like 

```
validate(event);  // this will throw errors
return await doTheThing(event);
```

would be nice to make it more readable and testable. But as mentioned I'll avoid any further optimizations in validation for now.

infrastructure/lib/atat-web-api-stack.ts
jeffsegal-ccpo: Fair enough - probably good to avoid any confusion with usage of `this`. Will change.

packages/api/models/Error.ts
jeffsegal-ccpo: That was me taking the first IDE suggestion and then subsequently getting an education into the three options there. Looks like `unknown` is more appropriate, if not just making it `string` outright. And actually, I believe we settled on an in-line type approach when we discussed it with @tonysumner-ccpo. E.g. 

Input Form:
```
{
  "pf_name": "this is too long",
  "description": "a description",
  "apps": [
    {
      "app_name": "app1",
      "price": 100
    },
    {
      "app_name": "app2",
      "price": 1000000
    }
  ]
}
```

HTTP 400 / Validation Error Response Body:
```
{
  "pf_name": {
      "input": "this is too long",
      "message": "Must be fewer than 250 characters."
    },
  "description": "a description",
  "apps": [
    {
      "app_name": "app1",
      "price": 100
    },
    {
      "app_name": "app2",
      "price": {
        "input": 1000000,
        "message": "Must be less than 1000."
    }
  ]
}
```

Note that this is totally notional and we can certainly tweak it to make life easier on both the function and UI side. But I am mentioning here mostly just to say that it probably doesn't matter what we settle on in this particular PR since I think it'll end up getting blown away. :)

.eslintrc.json
jeffsegal-ccpo: OK that's fine. I'll create a ticket to standardize logging and then we can revisit this.

kylelaker-ccpo (approved):

zachclark-ccpo (approved):

kylelaker-ccpo:
Rebased to resolve merge conflicts after merging #80. Assuming tests still come back green, we'll merge

PR #87, opened 2021-08-17
Add reusable instances of ErrorResponse for use across functions

jasonburkert-ccpo:
- Added reusable instances of `ErrorResponse` for use across functions to new `errors` module
- Applied same to `createFundingStep` function

Ticket: AT-6409

### DIFF
--- a/packages/api/utils/response.ts
+++ b/packages/api/utils/response.ts
@@ -1,5 +1,5 @@
 import { APIGatewayProxyResult } from "aws-lambda";
-import { Error, ErrorCodes } from "../models/Error";
+import { Error } from "../models/Error";
 
 type Headers = { [header: string]: string | number | boolean } | undefined;
 type MultiValueHeaders = { [header: string]: (string | number | boolean)[] } | undefined;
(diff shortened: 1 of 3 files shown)


jasonburkert-ccpo:
Making this available for review due to addition of shareable code.

packages/api/utils/response.ts
```diff
@@ -137,3 +137,84 @@ export class ErrorResponse extends Response {
...
+  ErrorStatusCode.NOT_FOUND
+);
+
+/**
+ * To be used when a Application Step is not found for a specified Portfolio Draft
+ */
+export const NO_SUCH_APPLICATION_STEP = new ErrorResponse(
+  { code: ErrorCodes.INVALID_INPUT, message: "Application Step not found for this Portfolio Draft" },
+  ErrorStatusCode.NOT_FOUND
+);
```
kylelaker-ccpo, line 220: I am tempted to recommend moving these outside of this file. `utils/response.ts` largely defines more general structures to be used, mostly free of specific application text or data. Maybe these can go in a `utils/errors.ts`? I could be talked out of that recommendation if we found a way to stop exporting `ErrorResponse` to force usage of these constants but I am not sure that we're ready for that.

In any case, I think it'd be good to try to keep this file more "abstract" and free from the specific application logic. I can envision a scenario where we might want to "share" this `reponse.ts` file (and perhaps `models/errors.ts`) across multiple applications.

kylelaker-ccpo (changes requested):

packages/api/utils/response.ts
jasonburkert-ccpo: My first inclination was to put it in `errors.ts`, but I found most of the supporting code in `response.ts`.  I think the recommendation makes sense.

jasonburkert-ccpo: New code relocated to `utils/errors.ts` in 162053ae3e3909bde7f2640a6b70125d8a47a8e5

kylelaker-ccpo (approved):

PR #92, opened 2021-08-18
Bump AWS CDK to 1.119.0

kylelaker-ccpo:
This bumps all dependencies which #91 fails to do


jasonburkert-ccpo (approved):
Successfully ran `cdk synth` and `cdk deploy` with version `1.119.0 (build 2921d64)`

PR #93, opened 2021-08-18
Implement 'deleteTaskOrderFile` function

zachclark-ccpo:
### Overview
Ticket: AT-6418

### Problems
There is no way to check if an object has been deleted with a single `DeleteObjectCommand`. This means we will need to make a `GetObjectCommand` or  `HeadObjectCommand` to check if it exists as part of the function. 
[HeadObjectCommand docs](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-s3/classes/headobjectcommand.html)
However, this solution is a problem because S3 has eventually consistent deletes, and a `GetObjectCommand` or  `HeadObjectCommand` may return a false positive.
A solution with versioning may be viable if we rework how we are creating the S3 bucket, but for now this PR only deletes the object in the bucket and **does not** check if it exists before doing so.

### DIFF
--- a/packages/api/taskOrderFiles/deleteTaskOrderFile.ts
+++ b/packages/api/taskOrderFiles/deleteTaskOrderFile.ts
@@ -7,9 +7,9 @@ import * as parser from "lambda-multipart-parser";
 import { ErrorCodes } from "../models/Error";
 import { isPathParameterPresent } from "../utils/validation";
 
-const bucketName = process.env.PENDING_BUCKET;
+const bucketName = process.env.ACCEPTED_BUCKET;
 export const NO_SUCH_TASK_ORDER_FILE = new ErrorResponse(
-  { code: ErrorCodes.INVALID_INPUT, message: "TaskOrderNumber must be specified in the URL path" },
+  { code: ErrorCodes.INVALID_INPUT, message: "TaskOrderId must be specified in the URL path" },
   ErrorStatusCode.BAD_REQUEST
 );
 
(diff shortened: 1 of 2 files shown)


packages/api/taskOrderFiles/deleteTaskOrderFile.ts
```diff
@@ -0,0 +1,47 @@
...
+  ErrorStatusCode.BAD_REQUEST
+);
+
+/**
+ * Delete a Task Order File
+ *
+ * @param event - The DELETE request from API Gateway
+ */
+export async function handler(event: APIGatewayProxyEvent): Promise<APIGatewayProxyResult> {
+  const taskOrderNumber = event.pathParameters?.taskOrderNumber;
```
jeffsegal-ccpo, line 22: We updated the API a little while back to generate this ID ourselves, so it should be `taskOrderId` instead.

packages/api/taskOrderFiles/deleteTaskOrderFile.ts
```diff
@@ -0,0 +1,47 @@
...
+import { ApiSuccessResponse, ErrorResponse, ErrorStatusCode, NoContentResponse } from "../utils/response";
+import { S3Client, DeleteObjectCommand, DeleteObjectCommandOutput } from "@aws-sdk/client-s3";
+import { FileMetadata, FileScanStatus } from "../models/FileMetadata";
+import * as parser from "lambda-multipart-parser";
+import { ErrorCodes } from "../models/Error";
+import { isPathParameterPresent } from "../utils/validation";
+
+const bucketName = process.env.PENDING_BUCKET;
+export const NO_SUCH_TASK_ORDER_FILE = new ErrorResponse(
+  { code: ErrorCodes.INVALID_INPUT, message: "TaskOrderNumber must be specified in the URL path" },
```
jeffsegal-ccpo, line 12: `taskOrderId`

packages/api/taskOrderFiles/deleteTaskOrderFile.ts
```diff
@@ -0,0 +1,47 @@
...
+    );
+  }
+  return new NoContentResponse();
+}
+
+async function deleteFile(taskOrderNumber: string): Promise<DeleteObjectCommandOutput> {
+  const client = new S3Client({});
+  const command = new DeleteObjectCommand({
+    Bucket: bucketName,
+    Key: taskOrderNumber,
```
jeffsegal-ccpo, line 43: `taskOrderId`

jeffsegal-ccpo (changes requested):

kylelaker:
What value would be provided by trying to have a strongly consistent API ourselves over an eventually consistent API from AWS? Do we _really_ need to care about returning a 404 here anyway? Fundamentally, you just want to know whether the delete completed successfully or not where that is defined by whether object is expected to still exist or not. Seems weird for two DELETEs in a row to have to return different status codes when the user probably doesn't really care. They just want to make sure it's not a 400 or 500.

jeffsegal-ccpo (approved):

kylelaker-ccpo (approved):

PR #107, opened 2021-08-20
Use CDK constructs

kylelaker-ccpo:
This is a first stab at using CDK constructs which will be important for refactoring that is necessary in upcoming infrastructure work, possibly including the `cdk-nag` remediation.


zachclark-ccpo (approved):
`cdk deploy` had no problems, everything seems to be working fine.

PR #124, opened 2021-08-30
Add Cognito integration

kylelaker-ccpo:
This configures Cognito to integrate with an external OIDC IdP.

Things to complete before ready-for-review:

- [x] Actually pull in secrets (right now, secrets are hardcoded as empty string `""`)
   - This is probably going to be through AWS Secrets Manager but we may need to discuss strategy
- [x] Clean up the way we handle the Cognito domain
- [x] Figure out how we're going to "know" the App URL

Opening this now for feedback on design and approach.

Ticket: AT-6531


kylelaker-ccpo:
This is ready for review. Basically, this sets up a `CognitoAuthentication` construct which sets up the "backend" for Cognito to authenticate against Azure AD. When the time comes, we'll configure the UserPoolClient for the front-end application in its repository. It is deployed as its own stack at the moment and I am not sure whether I like that or not so I am happy to hear that adds more complexity than we need and to be asked to move it back into the main stack. My thought process for that involved some thought that we might have to handle dependencies between Cognito, the API, and the front end but those don't feel apparent to me at this point. Keeping it as a high-level construct will remain no matter what.

This also starts moving some of the stack configuration to props instead of trying to resolve them inside the Stack. Trying to resolve environment variables, context, and other things inside stacks and constructs is considered an anti-pattern so I tried to prevent more tech debt in that regard here.

We also start using Secrets Manager here! These are some of our first real secrets. The naming here is subject to change in a future ticket but this works for now.

infrastructure/lib/constructs/authentication.ts
```diff
@@ -0,0 +1,249 @@
...
+        })
+    );
+  }
+
+  private processOidcIdps(
+    idps: OIDCIdentityProviderProps[],
+    props: CognitoAuthenticationProps
+  ): cognito.CfnUserPoolIdentityProvider[] {
+    return idps.map(
+      (idp) =>
```
jeffsegal-ccpo, line 233: I have done this exact sort of thing in Java (creating the generic interfaces to capture similar configs then mapping them to some library object or whatever) so you won't see any eye-rolling about this being overkill from me. :)

packages/cognito/preTokenGeneration/groups.test.ts
```diff
@@ -0,0 +1,160 @@
+import { PreTokenGenerationTriggerEvent } from "aws-lambda";
```
jeffsegal-ccpo, line 1: Really love having tests now that we have sufficiently complex logic here

jeffsegal-ccpo (approved):

packages/cognito/preTokenGeneration/groups.test.ts
kylelaker-ccpo: Yeah! I was really happy with `jest`'s `each` function too! I was able to use the SAME test for SAML and OIDC without any duplication! I like that it's capable of decomposing the given objects into parameters to the test function cleanly too.

PR #136, opened 2021-08-31
Update TypeScript to 4.4.2

kylelaker-ccpo:
This binds typescript to a particular minor version. Since we're using
the `strict` flag, it seems like new updates may break our code so we
should do this in a more controlled way.

In this case, we should probably come back through and reconsider
whether to be more strict in our type checking on errors but that should
be done as part of a larger error handling ticket.


jasonburkert-ccpo:
Following the completion of the larger error handling ticket, we'd set `useUnknownInCatchVariables` to true?

jeffsegal-ccpo (approved):

PR #137, opened 2021-08-31
Rename 'taskOrderFiles' function definitions and implementation files using OperationIds from API spec

jasonburkert-ccpo:
This is the approach is used for 'portfolioDrafts' functions.

I suggest API operation `getTaskOrder` be renamed `getTaskOrderMetadata`.  The current name is too easily confused with `downloadTaskOrder`.

Note that...
- we'd previously listed only three (3) of the four (4) operations in the stack definition file
- the old placeholder `getTaskOrderFiles` was clearly being conflated with `downloadTaskOrder`


kylelaker-ccpo:
@zachclark-ccpo be aware that this is likely to result in merge conflicts with AT-6416. You may want to be prepared to rebase when it is merged.

kylelaker-ccpo (approved):
Good stuff! Makes this far more clear

PR #147, opened 2021-08-31
API Gateway Validation for atat-web-api-stack

zachclark-ccpo:
### Description
Uses [SpecRestApi](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-apigateway.SpecRestApi.html) to generate the API directly from the `atat_provisioning_wizard_api.yaml`, and validates the requests based on the schema.

This required major refactoring of the stack and constructs, but will future proof request schema validation.

#### Ticket
[AT-6416](https://ccpo.atlassian.net/browse/AT-6416)

### TODO

- [x] Refactor api-function construct to add fn permissions and overrideLogicalId in Cfn
- [x] Refactor ApiDynamoDBFunctions in atat-web-api-stack, and in the yaml
- [x] Refactor S3 task order functions in atat-web-api-stack, and in the yaml
- [x] Clean up comments
- [x] Ensure validation works for all requests (test each endpoint)
- [x] Fix merge conflicts
~~Make a default error response model?~~ Not possible with Proxy Integration, but could consider it for the future if switch to custom integration OR if there is a ticket for ErrorResponse


backup_atat_provisioning_wizard_api.yaml
```diff
@@ -0,0 +1,737 @@
+openapi: 3.0.2
```
jeffsegal-ccpo, line 1: I assume we don't want this file right?

infrastructure/lib/atat-web-api-stack.ts
```diff
@@ -34,125 +34,91 @@ export class AtatWebApiStack extends cdk.Stack {
...
     });
 
+    // The API spec, which just so happens to be a valid CloudFormation snippet (with some actual CloudFormation
+    // in it) gets uploaded to S3
+    const apiAsset = new s3asset.Asset(this, "ApiSpecAsset", {
+      path: "./atat_provisioning_wizard_api.yaml",
+    });
+
+    // And now we include that snippet as an actual part of the template using the AWS::Include Transform. This
+    // is where the two previous pieces come together. Now, all the Fn::Sub and other functions will be resolved
```
jeffsegal-ccpo, line 86: I think we may need a small explanation here for the uninitiated among us. If the first piece is the spec in a S3 bucket, what's the other previous piece? The overall goal here is to connect the Lambdas to the APIGW Routes right?

infrastructure/lib/atat-web-api-stack.ts
```diff
@@ -34,125 +34,91 @@ export class AtatWebApiStack extends cdk.Stack {
...
+    // which means that we can reference the ARNs of the various functions.
+    const data = cdk.Fn.transform("AWS::Include", { Location: apiAsset.s3ObjectUrl });
```
jeffsegal-ccpo, line 88: Can we get a better name for this variable? :)

jeffsegal-ccpo (changes requested):
So this validates the requests but doesn't actually give us the generated code? Maybe I misunderstood that part. Do we still need to define our own types or is there follow-on work that would let us avoid that?

infrastructure/lib/atat-web-api-stack.ts
zachclark-ccpo: Those comments are a bit confusing because I incorporated the logicalId into the api-function construct. We don't need to define the lambda integration because we are linking to it in the `uri` section of the `x-amazon-api-gateway-integration`.

backup_atat_provisioning_wizard_api.yaml
zachclark-ccpo: We need a backup because there are still functions that don't have `x-amazon-api-gateway-integration` (because they aren't defined in the atat-web-api. For example, there is AT-6494 (my TODO task), which is to create the POST for submitting the portfolio: this is defined in the yaml spec, but not in the atat-web-api, so it doesn't have a lambda function to reference. We can't compile it because it doesn't have a x-amazon-api-gateway-integration, so I removed it from the yaml, and made a backup for when we implement it. This probably isn't the best way to do this.

infrastructure/lib/atat-web-api-stack.ts
zachclark-ccpo: Before there were 2 extra lines, extracting the logicalId so it could play nice with Cloudformation, but this logic was incorporated into the api-function construct to avoid repetition.

infrastructure/lib/atat-web-api-stack.ts
zachclark-ccpo: This is the data loaded into the api spec yaml, living on s3. We could name it `apiData` , `functionData`, `resolvedLambdaArnData`, `lambdaData`, `lambdaFnNames`, `functionalDataForS3`?

backup_atat_provisioning_wizard_api.yaml
kylelaker-ccpo: What if we just don't provide `x-amazon-api-gateway-*` attributes until there's an implementation ready?

infrastructure/lib/atat-web-api-stack.ts
kylelaker-ccpo: `apiSpecAsTemplateInclude`

infrastructure/lib/constructs/api-function.ts
```diff
@@ -45,13 +41,22 @@ export abstract class ApiFunction extends cdk.Construct {
...
     this.method = props.method;
     this.fn = new lambdaNodeJs.NodejsFunction(this, "Function", {
       entry: props.handlerPath,
       ...props.functionPropsOverride,
     });
-    this.route = props.resource.addMethod(props.method, new apigw.LambdaIntegration(this.fn));
+    this.fn.addPermission("AllowApiGatwewayInvoke", {
+      principal: new iam.ServicePrincipal("apigateway.amazonaws.com"),
+    });
+    this.logicalId = this.fn.node.defaultChild as lambda.CfnFunction;
```
kylelaker-ccpo, line 59: I am hesitant to call this `logicalId` because it isn't. I also very much don't want to make it publicly accessible since it can be derived from `this.fn`. Can we call it `const functionL1` (for Layer 1 construct) or `const cfnFunction` or `const rawCloudFormationFunction`? `Cfn` is often used for "CloudFormation" in the CDK and I see no reason to diverge from that pattern.

infrastructure/lib/constructs/api-function.ts
```diff
@@ -45,13 +41,22 @@ export abstract class ApiFunction extends cdk.Construct {
...
   protected constructor(scope: cdk.Construct, id: string, props: ApiFunctionProps) {
     super(scope, id);
     this.method = props.method;
     this.fn = new lambdaNodeJs.NodejsFunction(this, "Function", {
       entry: props.handlerPath,
       ...props.functionPropsOverride,
     });
-    this.route = props.resource.addMethod(props.method, new apigw.LambdaIntegration(this.fn));
+    this.fn.addPermission("AllowApiGatwewayInvoke", {
+      principal: new iam.ServicePrincipal("apigateway.amazonaws.com"),
```
kylelaker-ccpo, line 57: Thoughts on making this a private const in the module (top level) instead?

```
const APIGW_SERVICE_PRINCIPAL = new iam.ServicePrincipal("apigateway.amazonaws.com");
```

is this better or worse?

infrastructure/lib/constructs/api-function.ts
```diff
@@ -45,13 +41,22 @@ export abstract class ApiFunction extends cdk.Construct {
    */
   public readonly route: apigw.Method;
 
+  /**
+   * The CfnFunction logicalId, used for templating the api yaml.
+   */
+  public readonly logicalId: lambda.CfnFunction;
```
kylelaker-ccpo, line 47: See other comments on rewording this

infrastructure/lib/constructs/api-function.ts
kylelaker-ccpo: While you're changing class attributes, we may want to change `this.fn`'s type to be `lambda.IFunction` instead of `lambda.Function`.

infrastructure/lib/atat-web-api-stack.ts
```diff
@@ -34,125 +34,91 @@ export class AtatWebApiStack extends cdk.Stack {
...
-  const taskOrderId = taskOrderFiles.addResource("{taskOrderId}");
+function addTaskOrderRoutes(scope: cdk.Stack) {
   const taskOrderManagement = new TaskOrderLifecycle(scope, "TaskOrders");
-  const createTaskOrderFile = new ApiS3Function(scope, "CreateTaskOrderFile", {
-    resource: taskOrderFiles,
+  const uploadTaskOrder = new ApiS3Function(scope, "UploadTaskOrderFunction", {
     bucket: taskOrderManagement.pendingBucket,
     method: HttpMethod.POST,
-    handlerPath: packageRoot() + "/api/taskOrderFiles/createTaskOrderFile.ts",
+    handlerPath: packageRoot() + "/api/taskOrderFiles/uploadTaskOrder.ts",
```
kylelaker-ccpo, line 111: Make sure these names are correct based on the current state of Jason's PR to rename them. I don't remember if that has been merged yet or not.

infrastructure/package.json
```diff
@@ -43,6 +43,7 @@
     "@aws-cdk/aws-lambda": "1.119.0",
     "@aws-cdk/aws-lambda-nodejs": "1.119.0",
     "@aws-cdk/aws-s3": "1.119.0",
+    "@aws-cdk/aws-s3-assets": "1.119.0",
```
kylelaker-ccpo, line 46: Make sure to rebase/merge `develop` and bump this to `1.120.0`

infrastructure/lib/constructs/api-function.ts
```diff
@@ -45,13 +41,22 @@ export abstract class ApiFunction extends cdk.Construct {
...
     this.fn = new lambdaNodeJs.NodejsFunction(this, "Function", {
       entry: props.handlerPath,
       ...props.functionPropsOverride,
     });
-    this.route = props.resource.addMethod(props.method, new apigw.LambdaIntegration(this.fn));
+    this.fn.addPermission("AllowApiGatwewayInvoke", {
+      principal: new iam.ServicePrincipal("apigateway.amazonaws.com"),
+    });
+    this.logicalId = this.fn.node.defaultChild as lambda.CfnFunction;
+    this.logicalId.overrideLogicalId(id);
```
kylelaker-ccpo, line 60: Using the ID here is clever and I see what you're doing but I don't know if it's actually super safe to do. I'd prefer to add it as an element of the `ApiFunctionProps` and to use that instead. First, it makes it far more explicit what it's being used for so there's less chance of someone forgetting that it has to match the spec file. Second, this opens up the possibility to a naming collision since the name of this attribute is now specifically set to `id` but the construct itself has the same name. Luckily our construct itself isn't raw CloudFormation. But it still makes me a little nervous.

If you've seen examples from AWS or other code bases where they will override the logical ID using the ID of the parent construct, then I will accept this as idiomatic; however, barring that, this is better as a required field in the prop. Perhaps named `functionLogicalId` or something like that.

infrastructure/lib/atat-web-api-stack.ts
```diff
@@ -34,125 +34,91 @@ export class AtatWebApiStack extends cdk.Stack {
...
+    const data = cdk.Fn.transform("AWS::Include", { Location: apiAsset.s3ObjectUrl });
+
+    // And with the data now loaded from the template, we can use ApiDefinition.fromInline to parse it as real
+    // OpenAPI spec (because it was!) and now we've got all our special AWS values and variables interpolated.
+    const restApi = new apigw.SpecRestApi(this, "AtatSpecTest", {
+      apiDefinition: apigw.ApiDefinition.fromInline(data),
+      endpointTypes: [apigw.EndpointType.REGIONAL],
+      parameters: {
+        endpointConfigurationTypes: apigw.EndpointType.REGIONAL,
+      },
```
kylelaker-ccpo, line 97: We really ought to have a comment for why it's necessary to be repetitive here. And the answer to that is that when you're actually loading in these values from a `SpecRestApi`, API Gateway ignores the `endpointTypes` field in favor of `parameters.endpointConfigurationTypes`. We should first validate this is actually still a required workaround. If it is, then it should get a clear comment so that someone helpful trying to clean up duplication (probably future me or future you) doesn't remove this and break things. It will also be helpful when this needs to change to `.PRIVATE`.

kylelaker-ccpo (changes requested):

kylelaker-ccpo:
I think what I really want to see with `atat_provisioning_wizard_api.yaml` is that we have added the `x-amazon-apigateway-*` properties where we need to and ideally have made no other modifications. So any of these deletes/reorderings may need to be explained to me.

infrastructure/lib/atat-web-api-stack.ts
jeffsegal-ccpo: I'm still a bit lost there. We can take this conversation to standup and hopefully give us the context we need (and possibly update these comments but I don't want to slow down the trains).

infrastructure/lib/constructs/api-function.ts
jeffsegal-ccpo: I don't mind that because it calls it out as a special object which may not be clear to everyone.

backup_atat_provisioning_wizard_api.yaml
zachclark-ccpo: In that case, it won't work with `cdk deploy`. It throws a `missing amazon integration for ......` the endpoint.

infrastructure/lib/atat-web-api-stack.ts
zachclark-ccpo: These are the renamed files, don't worry.

backup_atat_provisioning_wizard_api.yaml
zachclark-ccpo: We could have 2 yaml files, one with the integrations, and the other just openapi 3.0 yaml. We coan point the web stack at the one with integrations.

infrastructure/lib/constructs/api-function.ts
zachclark-ccpo: I am confused on what to change. The `id` is a string passed in:
```
const createPortfolioStep = new ApiDynamoDBFunction(this, "CreatePortfolioStepFunction", {
```

```
protected constructor(scope: cdk.Construct, id: string, props: ApiFunctionProps) {
```
^The id for this implementation would be `CreatePortfolioStepFunction`

infrastructure/lib/constructs/api-function.ts
zachclark-ccpo:
```
lib/constructs/api-dynamodb-function.ts(30,13): error TS2339: Property 'addEnvironment' does not exist on type 'IFunction'.
lib/constructs/api-s3-function.ts(18,13): error TS2339: Property 'addEnvironment' does not exist on type 'IFunction'.
```
cdk isn't playing well with `lambda.IFunction`

kylelaker-ccpo: Good catch. There's a solution to this but it's not worth it. Continue using `lambda.Function`

infrastructure/lib/atat-web-api-stack.ts
kylelaker-ccpo: I think we just need to rewrite the flow of these comments. Let me know when you're done with the merge work and I'll work on that since I wrote the bad comment in the first place.

zachclark-ccpo:
Need to fix some stuff on the spec, should be resolved in 10-20 minutes
```
Errors found during import:
        Unable to put integration on 'POST' for resource at path '/portfolioDrafts/{portfolioDraftId}/application': Invalid lambda function
        Unable to put integration on 'GET' for resource at path '/taskOrderFiles/{taskOrderId}/file': Invalid lambda function
        Unable to put integration on 'POST' for resource at path '/portfolioDrafts/{portfolioDraftId}/submit': Invalid lambda function
        Unable to put integration on 'GET' for resource at path '/taskOrderFiles/{taskOrderId}': Invalid lambda function
Additionally, these warnings were found:
        Unsupported model type 'Content' in request schema for 'POST /portfolioDrafts'. Ignoring.
        Unsupported model type 'Content' in request schema for 'POST /portfolioDrafts'. Ignoring.
        Unsupported model type 'BinarySchema' in 200 response to method 'GET /taskOrderFiles/{taskOrderId}/file'. Ignoring.
        Schema for content type application/json for 'POST /portfolioDrafts/{portfolioDraftId}/submit' is not specified. Ignoring.
        Schema for content type text/plain fo

File: atat_provisioning_wizard_api.yaml [hunk 11]
```diff
@@ -298,7 +357,13 @@ paths:
           application/json:
             schema:
               $ref: '#/components/schemas/ApplicationStep'
-  '/portfolioDrafts/{portfolioDraftId}/submit':
+      x-amazon-apigateway-integration:
+          httpMethod: "POST"
+          uri: 
+            Fn::Sub: "arn:${AWS::Partition}:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${GetPortfolioDraftsFunction.Arn}/invocations"
```
kylelaker-ccpo, line 363: This and all unimplemented functions need to use a new function that returns a "Not Implemented" response (either 400 or 404. see if @jeffsegal-ccpo has a recommendation). That function should probably go ahead and use our ErrorResponse and can be reused.

Make sure when doing the find-and-replace that you don't accidentally replace the actual GetPortfolioDraftsFunction with the new NotImplementedFunction

kylelaker-ccpo (dismissed):

atat_provisioning_wizard_api.yaml
zachclark-ccpo: Done.

jeffsegal-ccpo (approved):

PR #152, opened 2021-09-01
Implement function 'GetPortfolioDraft'

jeffsegal-ccpo:
Includes unit test with mock DynamoDB client. Nothing fancy here.

Ticket: AT-6487

### DIFF
--- a/packages/api/portfolioDrafts/getPortfolioDraft.test.ts
+++ b/packages/api/portfolioDrafts/getPortfolioDraft.test.ts
@@ -1,0 +1,52 @@
+import { DynamoDBDocumentClient, GetCommand, GetCommandOutput } from "@aws-sdk/lib-dynamodb";
+import { APIGatewayProxyEvent } from "aws-lambda";
+import { mockClient } from "aws-sdk-client-mock";
+import { handler, NO_PORTFOLIO_PATH_PARAM, NO_SUCH_PORTFOLIO } from "./getPortfolioDraft";
+import { DATABASE_ERROR } from "../utils/errors";
+import { ErrorStatusCode, SuccessStatusCode } from "../utils/response";
+
+const ddbMock = mockClient(DynamoDBDocumentClient);
+beforeEach(() => {
+  ddbMock.reset();
+});
+
+const validRequest: APIGatewayProxyEvent = {
+  pathParameters: { portfolioDraftId: "1234" },
+} as any;
+
+describe("Validation tests", function () {
+  it("should require path param", async () => {
+    const emptyRequest: APIGatewayProxyEvent = {} as any;
+    expect(await handler(emptyRequest)).toEqual(NO_PORTFOLIO_PATH_PARAM);
+  });
+
+  it("should return error if portfolio ID does not exist", async () => {
+    const response: GetCommandOutput = {} as any;
+    ddbMock.on(GetCommand).resolves(response);
+    expect(await handler(validRequest)).toEqual(NO_SUCH_PORTFOLIO);
+  });
+
+  it("should return generic Error if exception caught", async () => {
+    ddbMock.on(GetCommand).rejects("Some error occurred");
+    expect(await handler(validRequest)).toEqual(DATABASE_ERROR);
+  });
+});
+
+describe("Happy Path tests", function () {
+  const item = { foo: "bar" };
+
+  beforeEach(() => {
+    ddbMock.on(GetCommand).resolves({
+      Item: item,
+    });
+  });
+
+  it("should return Item", async () => {
+    const result = await handler(validRequest);
+    expect(result.body).toEqual(JSON.stringify(item));
+  });
+  it("should return status code 200", async () => {
+    const result = await handler(validRequest);
+    expect(result.statusCode).toEqual(SuccessStatusCode.OK);
+  });
+});
--- a/packages/api/portfolioDrafts/getPortfolioDraft.test.ts
+++ b/packages/api/portfolioDrafts/getPortfolioDraft.test.ts
@@ -14,7 +14,7 @@ const validRequest: APIGatewayProxyEvent = {
   pathParameters: { portfolioDraftId: "1234" },
 } as any;
 
-describe("Validation tests", function () {
+describe("Validation tests", () => {
   it("should require path param", async () => {
     const emptyRequest: APIGatewayProxyEvent = {} as any;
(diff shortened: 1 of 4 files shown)


packages/api/portfolioDrafts/getPortfolioDraft.test.ts
```diff
@@ -0,0 +1,52 @@
...
+const ddbMock = mockClient(DynamoDBDocumentClient);
+beforeEach(() => {
+  ddbMock.reset();
+});
+
+const validRequest: APIGatewayProxyEvent = {
+  pathParameters: { portfolioDraftId: "1234" },
+} as any;
+
+describe("Validation tests", function () {
```
kylelaker-ccpo, line 17:
```suggestion
describe("Validation tests", () => {
```

We keep being inconsistent in our test code because we'll do `async () => { }` for the `it`s but `function () { }` for the `describe`s. But we can just do `() => { }` for the `describes` too. The `describe` can't be `async` but it'd be nice to use the same syntax.

kylelaker-ccpo, line 17:
```suggestion
describe("Validation tests", () => {
```

We keep being inconsistent in our test code because we'll do `async () => { }` for the `it`s but `function () { }` for the `describe`s. But we can just do `() => { }` for the `describes` too. The `describe` can't be `async` but it'd be nice to use the same syntax for this and the next `describe`.

This is a nitpick and the first thing I am seeing. If this is the only comment I leave then it's not worth spending time cleaning up.

packages/api/portfolioDrafts/getPortfolioDraft.ts
```diff
@@ -0,0 +1,51 @@
...
+          id: portfolioDraftId,
+        },
+      })
+    );
+
+    if (!result.Item) {
+      return NO_SUCH_PORTFOLIO;
+    } else {
+      return new ApiSuccessResponse<PortfolioDraft>(result.Item as PortfolioDraft, SuccessStatusCode.OK);
+    }
```
kylelaker-ccpo, line 46:
```suggestion
    }
    return new ApiSuccessResponse<PortfolioDraft>(result.Item as PortfolioDraft, SuccessStatusCode.OK);
```

Unnecessary `else` due to a `return`

kylelaker-ccpo (approved):
Nitpicks but nothing actually wrong here. If there are more comments then it'd be cool to see these cleaned up but I am going to approve because there's no reason to slow this down for these comments.

packages/api/portfolioDrafts/getPortfolioDraft.test.ts
jeffsegal-ccpo: All good - I'm more than happy to nip this stuff in the bud before copypasta kills us.

packages/api/portfolioDrafts/getPortfolioDraft.test.ts
```diff
@@ -0,0 +1,52 @@
...
+});
+
+const validRequest: APIGatewayProxyEvent = {
+  pathParameters: { portfolioDraftId: "1234" },
+} as any;
+
+describe("Validation tests", () => {
+  it("should require path param", async () => {
+    const emptyRequest: APIGatewayProxyEvent = {} as any;
+    expect(await handler(emptyRequest)).toEqual(NO_PORTFOLIO_PATH_PARAM);
```
jasonburkert-ccpo, line 20: Suggest using existing response from `errors.ts`
```suggestion
    expect(await handler(emptyRequest)).toEqual(PATH_VARIABLE_REQUIRED_BUT_MISSING);
```

packages/api/portfolioDrafts/getPortfolioDraft.ts
```diff
@@ -0,0 +1,50 @@
...
+import { DATABASE_ERROR } from "../utils/errors";
+
+export const NO_PORTFOLIO_PATH_PARAM = new ErrorResponse(
+  { code: ErrorCodes.INVALID_INPUT, message: "PortfolioDraftId must be specified in the URL path." },
+  ErrorStatusCode.BAD_REQUEST
+);
+export const NO_SUCH_PORTFOLIO = new ErrorResponse(
+  { code: ErrorCodes.INVALID_INPUT, message: "Portfolio Draft with the given ID does not exist" },
+  ErrorStatusCode.NOT_FOUND
+);
```
jasonburkert-ccpo, line 17: Please use existing NO_SUCH_PORTFOLIO_DRAFT

packages/api/portfolioDrafts/getPortfolioDraft.ts
```diff
@@ -0,0 +1,50 @@
...
+import { ApiSuccessResponse, ErrorResponse, ErrorStatusCode, SuccessStatusCode } from "../utils/response";
+import { PortfolioDraft } from "../models/PortfolioDraft";
+import { isPathParameterPresent } from "../utils/validation";
+import { dynamodbDocumentClient as client } from "../utils/dynamodb";
+import { DATABASE_ERROR } from "../utils/errors";
+
+export const NO_PORTFOLIO_PATH_PARAM = new ErrorResponse(
+  { code: ErrorCodes.INVALID_INPUT, message: "PortfolioDraftId must be specified in the URL path." },
+  ErrorStatusCode.BAD_REQUEST
+);
```
jasonburkert-ccpo, line 13: Please use existing PATH_VARIABLE_REQUIRED_BUT_MISSING

packages/api/portfolioDrafts/getPortfolioDraft.ts
```diff
@@ -0,0 +1,50 @@
...
+ * @param event - The GET request from API Gateway
+ */
+export async function handler(event: APIGatewayProxyEvent): Promise<APIGatewayProxyResult> {
+  const portfolioDraftId = event.pathParameters?.portfolioDraftId;
+
+  // TODO: validate that we get a valid ATAT_TABLE_NAME env variable
+  if (!isPathParameterPresent(portfolioDraftId)) {
+    return NO_PORTFOLIO_PATH_PARAM;
+  }
+
```
jasonburkert-ccpo, line 31: After checking for presence, suggest checking param is a valid UUIDv4 and issuing PATH_VARIABLE_INVALID if necessary.

jasonburkert-ccpo: After team discussion, checking the param for validity would be fine for the purposes of avoiding a database call, but not for returning information in the response.

PR #157, opened 2021-09-03
Refactor all the things

jasonburkert-ccpo:
Attempt to make the implementations from different sprints by different developers have more similarities


jasonburkert-ccpo:
There's more that could be done, but this is a good chunk.

zachclark-ccpo (approved):

kylelaker-ccpo (approved):
Definitely agree that there's more to do but there always is :D Thanks for your work on this. Hopefully we can all find time to do some refactoring and tech debt work soon to keep making this better.

PR #165, opened 2021-09-07
Remove all the things

kylelaker-ccpo:
On stack deletion, delete all the resources. There should not be
anything left behind at this stage of development. DynamoDB tables, S3
buckets, and Cognito User Pools will all be destroyed.


zachclark-ccpo (approved):
These changes are necessary, and will save us money.

jasonburkert-ccpo (approved):
Removed actual deployed resources with the feature branch.

PR #170, opened 2021-09-07
Implement function 'SubmitPortfolioDraft'

zachclark-ccpo:
## Overview
Implement SubmitPortfolioDraft function, which _currently_ adds a `submit_id` to an existing portfolioDraft.
It returns `202 Accepted` along with the entire PortfolioDraft (this may need to be changed).
The ACs aren't especially clear if the response is an empty body, or the entire portfolioDraft.

### Ticket
[AT-6494](https://ccpo.atlassian.net/browse/AT-6494)

### DIFF
--- a/atat_provisioning_wizard_api.yaml
+++ b/atat_provisioning_wizard_api.yaml
@@ -557,10 +557,10 @@ components:
               type: string
               description: CSP Provisioning Status
               enum:
-                - "NOT_STARTED"
-                - "IN_PROGRESS"
-                - "COMPLETE"
-                - "FAILED"
+                - "not_started"
+                - "in_progress"
+                - "complete"
+                - "failed"
     Error:
       type: object
       description: Generic error model
@@ -735,15 +735,15 @@ components:
     PortfolioDraftSummaryEx:
       value:
         - id: "dc2bbee6-8cdb-477e-a363-f9f1593a0a9b"
-          status: "NOT_STARTED"
+          status: "not_started"
           created_at: "2021-08-03T16:19:51.686Z"
           updated_at: "2021-08-03T16:19:51.686Z"
         - id: "c3e8c444-a971-4515-bc43-3cdfef072ccc"
-          status: "IN_PROGRESS"
+          status: "in_progress"
           created_at: "2021-08-03T16:19:24.362Z"
           updated_at: "2021-08-03T16:19:24.362Z"
         - id: "3f44ef42-2322-4a42-a381-6770e18eceba"
-          status: "COMPLETED"
+          status: "complete"
           created_at: "2021-08-03T16:21:07.978Z"
           updated_at: "2021-08-03T16:21:07.978Z"
     PortfolioDraftEx:
--- a/atat_provisioning_wizard_api.yaml
+++ b/atat_provisioning_wizard_api.yaml
@@ -377,8 +377,8 @@ paths:
           schema:
             type: string
       responses:
-        '201':
-          description: Successful operation
+        '202':
+          description: Accepted
         '400':
           description: Validation failed
           content:
@@ -396,10 +396,11 @@ paths:
             example: '{}'
           text/plain:
             example: ''
+      x-amazon-apigateway-request-validator: "request-params-validator"
       x-amazon-apigateway-integration:
         httpMethod: "POST"
         uri: 
-          Fn::Sub: "arn:${AWS::Partition}:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${NotImplementedFunction.Arn}/invocations"
+          Fn::Sub: "arn:${AWS::Partition}:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${SubmitPortfolioDraftFunction.Arn}/invocations"
         passthroughBehavior: "when_no_match"
         type: "aws_proxy"
   /taskOrderFiles:
--- a/atat_provisioning_wizard_api.yaml
+++ b/atat_provisioning_wizard_api.yaml
@@ -557,10 +557,10 @@ components:
               type: string
               description: CSP Provisioning Status
               enum:
-                - "not_started"
-                - "in_progress"
-                - "complete"
-                - "failed"
+                - "NOT_STARTED"
+                - "IN_PROGRESS"
+                - "COMPLETE"
+                - "FAILED"
     Error:
       type: object
       description: Generic error model
@@ -735,15 +735,15 @@ components:
     PortfolioDraftSummaryEx:
       value:
         - id: "dc2bbee6-8cdb-477e-a363-f9f1593a0a9b"
-          status: "not_started"
+          status: "NOT_STARTED"
           created_at: "2021-08-03T16:19:51.686Z"
           updated_at: "2021-08-03T16:19:51.686Z"
         - id: "c3e8c444-a971-4515-bc43-3cdfef072ccc"
-          status: "in_progress"
+          status: "IN_PROGRESS"
           created_at: "2021-08-03T16:19:24.362Z"
           updated_at: "2021-08-03T16:19:24.362Z"
         - id: "3f44ef42-2322-4a42-a381-6770e18eceba"
-          status: "complete"
+          status: "COMPLETED"
           created_at: "2021-08-03T16:21:07.978Z"
           updated_at: "2021-08-03T16:21:07.978Z"
     PortfolioDraftEx:
--- a/packages/api/portfolioDrafts/submit/submitPortfolioDraft.test.ts
+++ b/packages/api/portfolioDrafts/submit/submitPortfolioDraft.test.ts
@@ -1,0 +1,69 @@
+import { GetCommand, DynamoDBDocumentClient, UpdateCommand } from "@aws-sdk/lib-dynamodb";
+import { APIGatewayProxyEvent } from "aws-lambda";
+import { mockClient } from "aws-sdk-client-mock";
+import { handler } from "./submitPortfolioDraft";
+import {
+  NO_SUCH_PORTFOLIO_DRAFT,
+  REQUEST_BODY_NOT_EMPTY,
+  PORTFOLIO_ALREADY_SUBMITTED,
+  PATH_PARAMETER_REQUIRED_BUT_MISSING,
+  DATABASE_ERROR,
+} from "../../utils/errors";
+
+const ddbMock = mockClient(DynamoDBDocumentClient);
+beforeEach(() => {
+  ddbMock.reset();
+});
+
+const validRequest: APIGatewayProxyEvent = {
+  pathParameters: { portfolioDraftId: "1234" },
+} as any;
+
+describe("Validation tests", () => {
+  it("should require path param", async () => {
+    const emptyRequest: APIGatewayProxyEvent = {} as any;
+    expect(await handler(emptyRequest)).toEqual(PATH_PARAMETER_REQUIRED_BUT_MISSING);
+  });
+
+  it("should return generic Error if exception caught", async () => {
+    ddbMock.on(GetCommand).rejects("Some error occurred");
+    expect(await handler(validRequest)).toEqual(DATABASE_ERROR);
+  });
+
+  it("should return REQUEST_BODY_NOT_EMPTY when request body is not empty", async () => {
+    const requestBodyShouldBeEmpty = {
+      hi: "This should be empty",
+    };
+    const request: APIGatewayProxyEvent = {
+      body: JSON.stringify(requestBodyShouldBeEmpty),
+      pathParameters: { portfolioDraftId: "1234" },
+    } as any;
+
+    const response = await handler(request);
+    expect(response).toEqual(REQUEST_BODY_NOT_EMPTY);
+  });
+});
+describe("Handler response with mock dynamodb", function () {
+  it("should return error when the submitPortfolioDraftCommand fails, and portfolioDraft exists", async () => {
+    ddbMock.on(UpdateCommand).rejects({ name: "ConditionalCheckFailedException" });
+    const item = { foo: "bar" };
+    ddbMock.on(GetCommand).resolves({
+      Item: item,
+    });
+    // setting up new request
+    const request: APIGatewayProxyEvent = {
+      pathParameters: { portfolioDraftId: "1234" },
+    } as any;
+
+    expect(await handler(request)).toEqual(PORTFOLIO_ALREADY_SUBMITTED);
+  });
+  it("should return error when the submitPortfolioDraftCommand fails, and portfolioDraft doesn't exist", async () => {
+    ddbMock.on(UpdateCommand).rejects({ name: "ConditionalCheckFailedException" });
+    ddbMock.on(GetCommand).resolves({}); // resolves, but doesn't include an Item (!results.Item)
+    // setting up new request
+    const request: APIGatewayProxyEvent = {
+      pathParameters: { portfolioDraftId: "1234" },
+    } as any;
+    expect(await handler(request)).toEqual(NO_SUCH_PORTFOLIO_DRAFT);
+  });
+});
--- a/packages/api/portfolioDrafts/submit/submitPortfolioDraft.test.ts
+++ b/packages/api/portfolioDrafts/submit/submitPortfolioDraft.test.ts
@@ -35,8 +35,8 @@ describe("Validation tests", () => {
       hi: "This should be empty",
     };
     const request: APIGatewayProxyEvent = {
+      ...validRequest,
       body: JSON.stringify(requestBodyShouldBeEmpty),
-      pathParameters: { portfolioDraftId: "1234" },
     } as any;
 
     const response = await handler(request);
--- a/packages/api/portfolioDrafts/submit/submitPortfolioDraft.test.ts
+++ b/packages/api/portfolioDrafts/submit/submitPortfolioDraft.test.ts
@@ -50,20 +50,11 @@ describe("Handler response with mock dynamodb", function () {
     ddbMock.on(GetCommand).resolves({
       Item: item,
     });
-    // setting up new request
-    const request: APIGatewayProxyEvent = {
-      pathParameters: { portfolioDraftId: "1234" },
-    } as any;
-
-    expect(await handler(request)).toEqual(PORTFOLIO_ALREADY_SUBMITTED);
+    expect(await handler(validRequest)).toEqual(PORTFOLIO_ALREADY_SUBMITTED);
   });
   it("should return error when the submitPortfolioDraftCommand fails, and portfolioDraft doesn't exist", async () => {
     ddbMock.on(UpdateCommand).rejects({ name: "ConditionalCheckFailedException" });
     ddbMock.on(GetCommand).resolves({}); // resolves, but doesn't include an Item (!results.Item)
-    // setting up new request
-    const request: APIGatewayProxyEvent = {
-      pathParameters: { portfolioDraftId: "1234" },
-    } as any;
-    expect(await handler(request)).toEqual(NO_SUCH_PORTFOLIO_DRAFT);
+    expect(await handler(validRequest)).toEqual(NO_SUCH_PORTFOLIO_DRAFT);
   });
 });
--- a/packages/api/portfolioDrafts/submit/submitPortfolioDraft.ts
+++ b/packages/api/portfolioDrafts/submit/submitPortfolioDraft.ts
@@ -1,0 +1,88 @@
+import { DynamoDBClient } from "@aws-sdk/client-dynamodb";
+import { DynamoDBDocumentClient, UpdateCommand, UpdateCommandOutput } from "@aws-sdk/lib-dynamodb";
+import { APIGatewayProxyEvent, APIGatewayProxyResult } from "aws-lambda";
+import { ErrorCodes } from "../../models/Error";
+import { ApiSuccessResponse, ErrorResponse, ErrorStatusCode, SuccessStatusCode } from "../../utils/response";
+import { isBodyPresent, isPathParameterPresent, isPortfolioStep, isValidJson } from "../../utils/validation";
+import { DATABASE_ERROR } from "../../utils/errors";
+import { PortfolioDraft } from "../../models/PortfolioDraft";
+import { v4 as uuidv4 } from "uuid";
+
+const TABLE_NAME = process.env.ATAT_TABLE_NAME ?? "";
+export const NO_SUCH_PORTFOLIO = new ErrorResponse(
+  { code: ErrorCodes.INVALID_INPUT, message: "Portfolio Draft with the given ID does not exist" },
+  ErrorStatusCode.NOT_FOUND
+);
+export const EMPTY_REQUEST_BODY = new ErrorResponse(
+  { code: ErrorCodes.INVALID_INPUT, message: "Request body must be empty" },
+  ErrorStatusCode.BAD_REQUEST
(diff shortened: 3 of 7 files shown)


packages/api/portfolioDrafts/submit/submitPortfolioDraft.ts
```diff
@@ -0,0 +1,79 @@
...
+  const result = await ddb.send(
+    new UpdateCommand({
+      TableName: table,
+      Key: {
+        id: portfolioDraftId,
+      },
+      UpdateExpression: "set #portfolioVariable = :portfolio, updated_at = :now",
+      ExpressionAttributeNames: {
+        "#portfolioVariable": "submit_id",
+      },
```
kylelaker-ccpo, line 69: I wonder if we should more specifically track the datetime when it was submitted? Should we also have logic to prevent duplicate submissions? Is that a `400 Bad Request`?

Going to cc @jeffsegal-ccpo for the intent of the API here. Should the spec specify?

packages/api/portfolioDrafts/submit/submitPortfolioDraft.ts
```diff
@@ -0,0 +1,79 @@
...
+
+const TABLE_NAME = process.env.ATAT_TABLE_NAME ?? "";
+export const NO_SUCH_PORTFOLIO = new ErrorResponse(
+  { code: ErrorCodes.INVALID_INPUT, message: "Portfolio Draft with the given ID does not exist" },
+  ErrorStatusCode.NOT_FOUND
+);
+export const EMPTY_REQUEST_BODY = new ErrorResponse(
+  { code: ErrorCodes.INVALID_INPUT, message: "Request body must be empty" },
+  ErrorStatusCode.BAD_REQUEST
+);
```
kylelaker-ccpo, line 19: Some of the recent refactoring from @jasonburkert changed this in other functions. Can we make sure this stays consistent?

packages/api/portfolioDrafts/submit/submitPortfolioDraft.ts
kylelaker-ccpo, line 69: Confused about why one of these gets a name (`#portfolioVariable`->`submit_id`) but the other (`updated_at`) does not. Can we be consistent, Also `#portfolioVariable` is not a particularly descriptive name. If we were to make both of these variables, `#portfolioVariable` would be equally valid for both.

kylelaker-ccpo (changes requested):

zachclark-ccpo:
Should we be updating the CSP provisioning status?
```
status:
              type: string
              description: CSP Provisioning Status
              enum:
                - "not_started"
                - "in_progress"
                - "complete"
                - "failed"

File: packages/api/portfolioDrafts/submit/submitPortfolioDraft.ts (hunk 1 above)
Comment (zachclark-ccpo) [reply packages/api/portfolioDrafts/submit/submitPortfolioDraft.ts:69]: That is because submit_id is a **new attribute** added to the object, for example:
```
{
  "created_at": "2021-09-07T20:34:13.647Z",
  "id": "c54d1d10-faab-4870-9e7c-9d33dfa6743e",
  "num_portfolio_managers": 0,
  "status": "not_started",
  "submit_id": "f22872c2-ef76-4be4-a124-54e366b8bbf4",
  "updated_at": "2021-09-07T20:55:30.337Z"
}
```
`updated_at` already exists on the object, so we don't have to set a new expression attribute for it.

As for why its named `#portfolioVariable` : [dynamodb docs: An expression attribute name must begin with a pound sign (#)](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.ExpressionAttributeNames.html)

We are doing the same thing for the createPortfolioStep, and createFundingStep:
```
 UpdateExpression: "set #portfolioVariable = :portfolio, updated_at = :now",
    ExpressionAttributeNames: {
      "#portfolioVariable": FUNDING_STEP,
    },
    ExpressionAttributeValues: {
      ":portfolio": fundingStep,
      ":now": now,
    },
```
`#portfolioVariable` is replaced with `FUNDING_STEP` in this instance, and `:portfolio` is set to the fundingStep object.
In the dynamodb docs, they usually use `#x = :y` for these replaced variables.

If we wanted to be cleaner, I guess we could do `#submitPortfolioVariable` instead.

File: packages/api/portfolioDrafts/submit/submitPortfolioDraft.ts (hunk 2 above)
Comment (jasonburkert-ccpo) [reply packages/api/portfolioDrafts/submit/submitPortfolioDraft.ts:19]: See `NO_SUCH_PORTFOLIO_DRAFT` and `REQUEST_BODY_EMPTY` in `errors.ts`

File: packages/api/portfolioDrafts/submit/submitPortfolioDraft.ts (hunk 1 above)
Comment (kylelaker-ccpo) [reply packages/api/portfolioDrafts/submit/submitPortfolioDraft.ts:69]: Hmm well I still don't like the name `#portfolioVariable`. Can we pick a better name here? I'd like `#submitId` honestly. Or `#newAttribute`. Or something like that.

File: packages/api/portfolioDrafts/submit/submitPortfolioDraft.ts (hunk 2 above)
Comment (zachclark-ccpo) [reply packages/api/portfolioDrafts/submit/submitPortfolioDraft.ts:19]: I will update these to use the ones in `errors.ts`

File: packages/api/portfolioDrafts/submit/submitPortfolioDraft.ts (hunk 1 above)
Comment (jeffsegal-ccpo) [reply packages/api/portfolioDrafts/submit/submitPortfolioDraft.ts:69]: Sure, not a bad idea to track the submission time in a new attribute. We should also update the `status` attribute of the Portfolio Item. It's initially set as 
```
    status: ProvisioningStatus.NOT_STARTED,
```
We should update it to ProvisioningStatus.IN_PROGRESS.

Along those lines, if the existing `status` is anything other than `NOT_STARTED`, then we should return a `400 Bad Request`. We can add a blurb to the `description` of the response in the spec clarifying this behavior if we want.

File: packages/api/portfolioDrafts/submit/submitPortfolioDraft.ts (hunk 2 above)
Comment (jeffsegal-ccpo) [reply packages/api/portfolioDrafts/submit/submitPortfolioDraft.ts:19]: We also have these constants defined in errors.ts so we should just re-use those.

File: packages/api/portfolioDrafts/submit/submitPortfolioDraft.ts (hunk 1 above)
Comment (jeffsegal-ccpo) [reply packages/api/portfolioDrafts/submit/submitPortfolioDraft.ts:69]: It's definitely confusing and it's hard to see why DynamoDB would require you to use a variable name for a variable itself regardless (seems like more of a fringe edge case when you want a dynamically generated attribute name). But if that's indeed required then so be it.

**Review by jeffsegal-ccpo — CHANGES_REQUESTED** [review]

**Review by jasonburkert-ccpo — APPROVED** [review]
Approved with the assumption that other reviewers' comments are addressed.

File: packages/api/portfolioDrafts/submit/submitPortfolioDraft.ts (hunk 1 above)
Comment (jasonburkert-ccpo) [reply packages/api/portfolioDrafts/submit/submitPortfolioDraft.ts]: Why add `submit_id` and not say, a datetime `submitted_at`?  Is it because something is needed to accompany `updated_at` which will have that datetime?  Does the presence of `submit_id` act as a flag to indicate this is no longer a Portfolio _Draft_?

Comment (jasonburkert-ccpo) [reply packages/api/portfolioDrafts/submit/submitPortfolioDraft.ts]: Captured these decisions in a ticket comment for @hannahposeyscholl-ccpo and @elsanegash-ccpo

Comment (jasonburkert-ccpo) [reply packages/api/portfolioDrafts/submit/submitPortfolioDraft.ts]: I wrestled with some of the issues discussed above in `feature/AT-6409`.  Here was my attempt at a consistent approach that could be used across steps.  The value of `stepKey` can even reference [one of these](https://github.com/dod-ccpo/atat-web-api/blob/d5cb5256f228c36ec4f9c57ae28d24906306ee60/packages/api/models/PortfolioDraft.ts#L7).
```
      new UpdateCommand({
        TableName: process.env.ATAT_TABLE_NAME ?? "",
        Key: {
          id: portfolioDraftId,
        },
        UpdateExpression: "set #stepKey = :step, #updateAtKey = :now",
        ExpressionAttributeNames: {
          // values are JSON keys
          "#stepKey": "funding_step",
          "#updateAtKey": "updated_at",
        },
        ExpressionAttributeValues: {
          ":step": fundingStep,
          ":now": new Date().toISOString(),
        },
        ConditionExpression: "attribute_exists(created_at)",
        ReturnValues: "ALL_NEW",
      })
```

Comment (zachclark-ccpo) [reply packages/api/portfolioDrafts/submit/submitPortfolioDraft.ts]:
```
export async function submitPortfolioDraftCommand(
  table: string,
  portfolioDraftId: string
): Promise<UpdateCommandOutput> {
  const dynamodb = new DynamoDBClient({});
  const ddb = DynamoDBDocumentClient.from(dynamodb);
  const now = new Date().toISOString();
  const submitId = uuidv4();
  const result = await ddb.send(
    new UpdateCommand({
      TableName: table,
      Key: {
        id: portfolioDraftId,
      },
      UpdateExpression: "set #submitId = :submitIdValue, updated_at = :now",
      ExpressionAttributeNames: {
        "#submitId": "submit_id",
      },
      ExpressionAttributeValues: {
        ":now": now,
        ":submitIdValue": submitId,
      },
      ConditionExpression: "attribute_exists(created_at)",
      ReturnValues: "ALL_NEW",
    })
  );
  return result;
}
```
Is this better?

Comment (jeffsegal-ccpo) [reply packages/api/portfolioDrafts/submit/submitPortfolioDraft.ts]: @jasonburkert , the `submit_id` field is a placeholder that we'll leave in for now but is subject to change. We may end up nixing it given how we're going to track provisioning status, but that's something to deal with in M3.

Comment (jasonburkert-ccpo) [reply packages/api/portfolioDrafts/submit/submitPortfolioDraft.ts]: @jeffsegal-ccpo Thanks for clarification.

Comment (jasonburkert-ccpo) [reply packages/api/portfolioDrafts/submit/submitPortfolioDraft.ts]: @zachclark-ccpo That looks pretty good to me.  Only changes I would make:

1. eliminate `const submitId` and instead `":submitIdValue": uuidv4(),`
2. move table value into the Update Command creation unless table is needed elsewhere
`TableName: process.env.ATAT_TABLE_NAME ?? "",`

General question:  In the future would we want to log this as an important INFO-level event _"Jane Manager submitted Portfolio Draft ABC, submitId: XYZ"_

File: packages/api/portfolioDrafts/submit/submitPortfolioDraft.ts [hunk 3]
```diff
@@ -0,0 +1,104 @@
...
+        ":submitIdValue": uuidv4(),
+        ":statusUpdate": ProvisioningStatus.IN_PROGRESS,
+      },
+      ConditionExpression: "attribute_exists(created_at) AND attribute_not_exists(submit_id)",
+      ReturnValues: "ALL_NEW",
+    })
+  );
+  return result;
+}
+// We cannot use ConditionExpression to differentiate between a 404 and 400 response, so we need
```
jeffsegal-ccpo, line 87: Too bad but good to know.

jeffsegal-ccpo:
> Should we be updating the CSP provisioning status?
> 
> ```
> status:
>               type: string
>               description: CSP Provisioning Status
>               enum:
>                 - "not_started"
>                 - "in_progress"
>                 - "complete"
>                 - "failed"
> ```

To fix the capitalization? Yeah let's go ahead and do that.

zachclark-ccpo:
> To fix the capitalization? Yeah let's go ahead and do that.

So submitting now updates the status to `in_progress` and you can't submit a portfolioDraft that has already been submitted.
We need another blanket spec update I think.

jeffsegal-ccpo (approved):
Looks good as long as you fix the enum capitalization thing. Thanks!

packages/api/portfolioDrafts/submit/submitPortfolioDraft.ts
```diff
@@ -0,0 +1,104 @@
...
+      ExpressionAttributeNames: {
+        "#submitId": "submit_id",
+        "#status": "status",
+      },
+      ExpressionAttributeValues: {
+        ":now": now,
+        ":submitIdValue": uuidv4(),
+        ":statusUpdate": ProvisioningStatus.IN_PROGRESS,
+      },
+      ConditionExpression: "attribute_exists(created_at) AND attribute_not_exists(submit_id)",
```
kylelaker-ccpo, line 81:
```suggestion
      ConditionExpression: "attribute_exists(created_at) AND attribute_not_exists(submit_id) AND status = not_started",
```

Or some code that accomplishes this goal correctly? And I agree that it _should_ be an invariant that when `submit_id` exists that `status = "not_started"` but do we want to guard against this anyway? I believe that the true guarantee is that we'll only allow submitting a portfolio that's `not_started` and so it feels weird to rely on an implied invariant of our database schema here.

packages/api/portfolioDrafts/submit/submitPortfolioDraft.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  try {
+    const result = await submitPortfolioDraftCommand(TABLE_NAME, portfolioDraftId);
+    return new ApiSuccessResponse(result.Attributes as PortfolioDraft, SuccessStatusCode.ACCEPTED);
+  } catch (error) {
+    console.log("Error: " + error);
+    // Check if the portfolioDraft exists
+    if (error.name === "ConditionalCheckFailedException") {
+      const getResult = await doesPortfolioDraftExistCommand(TABLE_NAME, portfolioDraftId);
+      console.log("getResult: " + JSON.stringify(getResult));
```
kylelaker-ccpo, line 48: We probably shouldn't log this information every time we hit this path. If we do, it needs to provide far more useful context about where we are and what has happened.

packages/api/portfolioDrafts/submit/submitPortfolioDraft.test.ts
```diff
@@ -0,0 +1,69 @@
...
+  it("should return error when the submitPortfolioDraftCommand fails, and portfolioDraft exists", async () => {
+    ddbMock.on(UpdateCommand).rejects({ name: "ConditionalCheckFailedException" });
+    const item = { foo: "bar" };
+    ddbMock.on(GetCommand).resolves({
+      Item: item,
+    });
+    // setting up new request
+    const request: APIGatewayProxyEvent = {
+      pathParameters: { portfolioDraftId: "1234" },
+    } as any;
```
kylelaker-ccpo, line 56: Why aren't we re-using `validRequest` defined on L18?

packages/api/portfolioDrafts/submit/submitPortfolioDraft.test.ts
```diff
@@ -0,0 +1,69 @@
...
+  });
+
+  it("should return REQUEST_BODY_NOT_EMPTY when request body is not empty", async () => {
+    const requestBodyShouldBeEmpty = {
+      hi: "This should be empty",
+    };
+    const request: APIGatewayProxyEvent = {
+      body: JSON.stringify(requestBodyShouldBeEmpty),
+      pathParameters: { portfolioDraftId: "1234" },
+    } as any;
```
kylelaker-ccpo, line 40:
```suggestion
    const request: APIGatewayProxyEvent = {
      body: JSON.stringify(requestBodyShouldBeEmpty),
      ...validRequest,
    } as any;
```

?

packages/api/portfolioDrafts/submit/submitPortfolioDraft.ts
```diff
@@ -0,0 +1,104 @@
...
+    console.log("Database error: " + error);
+    return DATABASE_ERROR;
+  }
+}
+export async function submitPortfolioDraftCommand(
+  table: string,
+  portfolioDraftId: string
+): Promise<UpdateCommandOutput> {
+  const dynamodb = new DynamoDBClient({});
+  const ddb = DynamoDBDocumentClient.from(dynamodb);
```
kylelaker-ccpo, line 63: Here, and elsewhere in this file, re-use the stuff in `../utils/dynamodb.ts` or justify why we can't. You should `import { dynamodbDocumentClient as client } from "../utils/dynamodb"`

packages/api/portfolioDrafts/submit/submitPortfolioDraft.ts
```diff
@@ -0,0 +1,104 @@
...
+  const result = await ddb.send(
+    new GetCommand({
+      TableName: table,
+      Key: {
+        id: portfolioDraftId,
+      },
+    })
+  );
+  return result;
```
kylelaker-ccpo, line 103:
```suggestion
  return ddb.send(
    new GetCommand({
      TableName: table,
      Key: {
        id: portfolioDraftId,
      },
    })
  );
```

Do we really need to `await` and store this as a temp var? Even if we do need to `await` storing to a variable named `result` that we don't use doesn't feel helpful.

packages/api/portfolioDrafts/submit/submitPortfolioDraft.ts
kylelaker-ccpo: And to clarify why I'm hitting to hard on reusing that, we're going to need it when we start actually setting up private networking and specifying actual attributes to pass to the constructor of `DynamoDBClient`. I'd far rather do that in one place than 100 and we need to understand why that code is not being used today.

kylelaker-ccpo (changes requested):
There's definitely some improvement and I appreciate the work you've done. We still need to clean a few things up before this is ready to be merged.

atat_provisioning_wizard_api.yaml
```diff
@@ -556,10 +557,10 @@ components:
...
               description: CSP Provisioning Status
               enum:
-                - "not_started"
-                - "in_progress"
-                - "complete"
-                - "failed"
+                - "NOT_STARTED"
+                - "IN_PROGRESS"
+                - "COMPLETE"
+                - "FAILED"
```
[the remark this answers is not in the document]
jasonburkert-ccpo: Agreed.  Enum _values_ are implemented as lower case and specified as such in the API spec.  The enum _names_ we've been capitalizing.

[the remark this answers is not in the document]
jasonburkert-ccpo: Values for `ErrorCodes` seems to be the exception -- implementation and spec are both capitalized.  All other enums in API spec are lower case values.

packages/api/portfolioDrafts/submit/submitPortfolioDraft.ts
jasonburkert-ccpo: @kylelaker-ccpo Good catch.  I've not been on the lookout for this.

packages/api/portfolioDrafts/submit/submitPortfolioDraft.ts
jeffsegal-ccpo: And if we're going to have to make two trips anyway, we might as well put each condition in its own request right? Not saying to do that in this PR, but if that's the optimal path then we can at least note it in a comment.

atat_provisioning_wizard_api.yaml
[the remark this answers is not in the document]
jeffsegal-ccpo: That's my fault then - I thought we had already been in the process of making the actual values capitalized. Ultimately I want us to be consistent but I'll defer to you guys for the best way to get there.

packages/api/portfolioDrafts/submit/submitPortfolioDraft.ts
jeffsegal-ccpo: Agreed - in my Java PR review days, I was constantly on the lookout for unnecessary local variables. After all, removing them means we don't have to tackle the hardest problem in programming - naming variables.

packages/api/portfolioDrafts/submit/submitPortfolioDraft.test.ts
zachclark-ccpo: This change results in a typescript error 
```
'body' is specified more than once, so this usage will be overwritten.ts(2783)
submitPortfolioDraft.test.ts(39, 7): This spread always overwrites this property.
```

kylelaker-ccpo: I am confused when/where `validRequest` gets a body attribute but that's fine. Just flip them.

```
{
  ...validRequest,
  body: JSON.stringify(requestBodyShouldBeEmpty)
}
```

packages/api/portfolioDrafts/submit/submitPortfolioDraft.ts
```diff
@@ -0,0 +1,97 @@
...
+        "#status": "status",
+      },
+      ExpressionAttributeValues: {
+        ":now": now,
+        ":submitIdValue": uuidv4(),
+        ":statusUpdate": ProvisioningStatus.IN_PROGRESS,
+        ":currentStatus": ProvisioningStatus.NOT_STARTED,
+      },
+      ConditionExpression:
+        "attribute_exists(created_at) AND attribute_not_exists(submit_id) AND #status = :currentStatus",
```
kylelaker-ccpo, line 73:
```suggestion
        ":expectedStatus": ProvisioningStatus.NOT_STARTED,
      },
      ConditionExpression:
        "attribute_exists(created_at) AND attribute_not_exists(submit_id) AND #status = :expectedStatus",
```

maybe? We're expecting/requiring it to be `NOT_STARTED`. The current status could be anything.

packages/api/portfolioDrafts/submit/submitPortfolioDraft.ts
```diff
@@ -0,0 +1,97 @@
...
+      ReturnValues: "ALL_NEW",
+    })
+  );
+  return result;
+}
+// The `ConditionExpression` that we're using performs two different types of validation. The first is that
```
kylelaker-ccpo, line 79: As mentioned, I think this would be better at the `if` check in the main handler function that here.

kylelaker-ccpo (changes requested):

packages/api/portfolioDrafts/submit/submitPortfolioDraft.ts
```diff
@@ -0,0 +1,96 @@
...
+        ":submitIdValue": uuidv4(),
+        ":statusUpdate": ProvisioningStatus.IN_PROGRESS,
+        ":expectedStatus": ProvisioningStatus.NOT_STARTED,
+      },
+      ConditionExpression:
+        "attribute_exists(created_at) AND attribute_not_exists(submit_id) AND #status = :expectedStatus",
+      ReturnValues: "ALL_NEW",
+    })
+  );
+  return result;
```
kylelaker-ccpo, line 82:
```suggestion
  return client.send(
    new UpdateCommand({
      TableName: table,
      Key: {
        id: portfolioDraftId,
      },
      UpdateExpression: "set #submitId = :submitIdValue, updated_at = :now, #status = :statusUpdate",
      ExpressionAttributeNames: {
        "#submitId": "submit_id",
        "#status": "status",
      },
      ExpressionAttributeValues: {
        ":now": new Date().toISOString(),
        ":submitIdValue": uuidv4(),
        ":statusUpdate": ProvisioningStatus.IN_PROGRESS,
        ":expectedStatus": ProvisioningStatus.NOT_STARTED,
      },
      ConditionExpression:
        "attribute_exists(created_at) AND attribute_not_exists(submit_id) AND #status = :expectedStatus",
      ReturnValues: "ALL_NEW",
    })
  );
```

kylelaker-ccpo (approved):
Assuming that I didn't just break the tests with that, this looks good to me!

PR #177, opened 2021-09-09
Make an SSM parameter with User Pool ID available

kylelaker-ccpo:
To support the UI repo's CDK stack to be able to create a User Pool
Client, we need to actually make the pool's ID available somewhere. This
does a few things:

1. Renames `TicketId` to `EnvironmentId` to support names like 'dev',
'staging', or 'prod' more clearly.
2. Adds the SSM parameter for the User Pool ID with a
good-enough-for-now naming convention until we have time to formalize a
process for this (if that ever becomes necessary)
3. Updates existing code to use `EnvironmentId` naming

### DIFF
--- a/infrastructure/lib/atat-auth-stack.ts
+++ b/infrastructure/lib/atat-auth-stack.ts
@@ -1,13 +1,13 @@
 import * as secretsmanager from "@aws-cdk/aws-secretsmanager";
+import * as ssm from "@aws-cdk/aws-ssm";
 import * as cdk from "@aws-cdk/core";
-import { v4 as uuidv4 } from "uuid";
 import { CognitoAuthentication } from "./constructs/authentication";
 import { HttpMethod } from "./http";
 
 export interface AtatAuthStackProps extends cdk.StackProps {
   secretName: string;
   providerName: string;
-  ticketId?: string;
+  environmentId: string;
   adminsGroupName?: string;
   usersGroupName?: string;
   removalPolicy?: cdk.RemovalPolicy;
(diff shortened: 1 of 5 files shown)


infrastructure/lib/atat-auth-stack.ts
```diff
@@ -36,5 +35,10 @@ export class AtatAuthStack extends cdk.Stack {
         },
       ],
     });
+    const ssmParam = new ssm.StringParameter(this, "UserPoolIdParameter", {
+      description: "Cognito User Pool ID",
+      stringValue: cognitoAuthentication.userPool.userPoolId,
+      parameterName: `/atat/${props.environmentId}/cognito/userpool/id`,
```
jeffsegal-ccpo, line 41: Is this future proofing for a time when environments are sharing a single SSM store?

jeffsegal-ccpo (approved):

infrastructure/lib/atat-auth-stack.ts
kylelaker-ccpo: Yeah so this is actually a little unfortunate. It's actually now-proofing rather than future-proofing. We currently have multiple sandbox environments in an account. Eventually we'll only have one. But we already have other Cognito-related resources that _have_ to be unique globally (the domain specifically) so this is an extension of that. I'd _like_ to find a creative way to make it optional but it's already required and I really want to avoid something like:

```typescript
if (['prod', 'staging', 'dev'].contains(props.environmentId)) {
  // drop the ID from this one use case;
} else {
  // keep it
}
```

jeffsegal-ccpo: Got it. Definitely agree with applying KISS here so as to avoid dangerous bugs for special cases around environment names.

PR #179, opened 2021-09-09
Avoid invalid uses of INVALID_INPUT

kylelaker-ccpo:
Using the INVALID_INPUT error code is supposed to be an indication that
the response is an actual ValidationError response. Unfortunately, we
have seen it used incorrectly for any time a 4xx error is being returned
and this has been inconsistent.

This introduces a change to the `utils/response.ts` to ensure that the
only time that INVALID_INPUT responses can be issued is by creating a
ValidationErrorResponse object and providing the error map and a
message. For all other error responses, OTHER will be used as is correct
per the API spec.

### DIFF
--- a/packages/api/portfolioDrafts/getPortfolioDraft.ts
+++ b/packages/api/portfolioDrafts/getPortfolioDraft.ts
@@ -1,18 +1,17 @@
 import { GetCommand } from "@aws-sdk/lib-dynamodb";
 import { APIGatewayProxyEvent, APIGatewayProxyResult } from "aws-lambda";
-import { ErrorCodes } from "../models/Error";
-import { ApiSuccessResponse, ErrorResponse, ErrorStatusCode, SuccessStatusCode } from "../utils/response";
 import { PortfolioDraft } from "../models/PortfolioDraft";
-import { isPathParameterPresent } from "../utils/validation";
 import { dynamodbDocumentClient as client } from "../utils/dynamodb";
 import { DATABASE_ERROR } from "../utils/errors";
+import { ApiSuccessResponse, ErrorResponse, ErrorStatusCode, SuccessStatusCode } from "../utils/response";
+import { isPathParameterPresent } from "../utils/validation";
 
 export const NO_PORTFOLIO_PATH_PARAM = new ErrorResponse(
-  { code: ErrorCodes.INVALID_INPUT, message: "PortfolioDraftId must be specified in the URL path." },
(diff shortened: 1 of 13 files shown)


packages/api/portfolioDrafts/getPortfolioDraft.ts
```diff
@@ -1,18 +1,17 @@
...
-import { ErrorCodes } from "../models/Error";
-import { ApiSuccessResponse, ErrorResponse, ErrorStatusCode, SuccessStatusCode } from "../utils/response";
 import { PortfolioDraft } from "../models/PortfolioDraft";
-import { isPathParameterPresent } from "../utils/validation";
 import { dynamodbDocumentClient as client } from "../utils/dynamodb";
 import { DATABASE_ERROR } from "../utils/errors";
+import { ApiSuccessResponse, ErrorResponse, ErrorStatusCode, SuccessStatusCode } from "../utils/response";
+import { isPathParameterPresent } from "../utils/validation";
 
 export const NO_PORTFOLIO_PATH_PARAM = new ErrorResponse(
```
jeffsegal-ccpo, line 9: Definitely some more cleanup and consolidation to do here but I'll leave it to @zachclark-ccpo to do in 6413 (e.g. moving this and other errors up to a errors.ts).

jeffsegal-ccpo (approved):

packages/api/portfolioDrafts/getPortfolioDraft.ts
kylelaker-ccpo: Totally agree. I was trying to fight the urge of scope creeping to fixing _all_ of our error handling stuff. Wanted to try to stay targeted to updating the constructor API and class hierarchy wherever they currently were used (and I think IDE autoformatting betrayed me and caught a few import reorderings as part of that).

Though, there's also likely an argument to make that errors that only have one consumer should not be global/shared errors. `NO_PORTFOLIO_PATH_PARAM` specifically isn't one of those though. But AT-6413 is a good home for those things.

kylelaker-ccpo:
Okay so it was fun to learn that you can actually refer to a particular variant of an enum as a type. Which means we can enforce that `ValidationError`'s `code` is an `INVALID_INPUT`.

Also there were a few other class name fixups added. I don't know if I like `OtherErrorResponse` but if you understand our API then it does make a bit more sense. And renaming `ErrorCodes` -> `ErrorCode` didn't touch many files that weren't being touched anyway.

zachclark-ccpo (approved):

PR #187, opened 2021-09-11
Add summary data to GET /portfolioDrafts

juliusfitzhugh-ccpo:
The UX design for the Dashboard added some new data to the Portfolio cards after the initial API design was complete. As a result, we need to update the API design and implementation of the Portfolio Summary objects to include:

- Name of Portfolio (currently exists in Portfolio Step metadata)
- Number of Portfolio Managers
- Number of Applications
- Number of Environments
- Number of Task Orders

### TODO
- [x] The API design for GET /portfolioDrafts includes the attributes listed in the description.
- [x] Update POST ```/portfolioDrafts/:id/portfolio``` to denormalize portfolio name & number of portfolio managers.
- [x] Update POST ```/portfolioDrafts/:id/applications``` to denormalize the number of Applications & number of Environments.
- [x] Update POST ```/portfolioDrafts/:id/funding``` to denormalize the number of Task Orders.
- [x] The API implementation of GET /portfolioDrafts includes each attribute (names must match the spec).

Ticket: AT-6575


packages/api/portfolioDrafts/createPortfolioDraft.ts
```diff
@@ -23,7 +23,11 @@ export async function handler(event: APIGatewayProxyEvent): Promise<APIGatewayPr
     created_at: now,
     updated_at: now,
     status: ProvisioningStatus.NOT_STARTED,
+    name: "",
     num_portfolio_managers: 0,
+    num_applications: 0,
+    num_environments: 0,
+    num_task_orders: 0,
```
kylelaker-ccpo, line 30: Do we really want to store this in the database? These are computed attributes so we're going to have to make specific efforts to keep them in sync. Not suggesting that that is wrong; just curious what our design is here.

juliusfitzhugh-ccpo: @kylelaker-ccpo These are initial values that will be updated in the portfolio, funding, and application steps to denormalize the data for the final portfolio draft summary (similar to how `num_portfolio_managers` was there initially). I was trying to follow the convention that was already in place, and also to ensure the type interface aligns since a test failed due to the interface shape being different. Should they be removed or we take a different approach?

jasonburkert-ccpo: The intent is to provide a light weight solution to satisfy the My Portfolios page design.
image
@jeffsegal-ccpo spent a good bit of time considering a design that normalized some of this data into their own database items that were children of the portfolio draft.  However, in the end concluded that such a design introduced too much complexity where we have nested hierarchies.  Including these pre-computed values at the top level of the portfolio draft (and in a PortfolioDraftSummary object) will require some efforts to keep them in sync with the stored data, but will ease the burden of retrieving the data needed for the My Portfolios page.

Apart from this listing of several Portfolio Drafts, almost all of our other access patterns are directly through a specific portfolio draft where our Partition Key gives us direct access.  By contrast the access pattern required by the My Portfolio's design however requires something different.  Definitely open to alternatives.

kylelaker-ccpo:
For the linting failures, you can possibly solve them with `npm run lint:fix` but I also use https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint and have applied the:

```
    "editor.codeActionsOnSave": {
        "source.fixAll.eslint": true
    }
```

config to my VS Code `settings.json` file which automatically formats files on save which gets it right about 99.9% of the time.

juliusfitzhugh-ccpo:
@kylelaker-ccpo Thanks! I used `npm run lint:fix` to fix the linting errors. I'm used to them being fixed upon saving the file, so I'll update my VS code settings as you suggested.

atat_provisioning_wizard_api.yaml
```diff
@@ -551,8 +551,16 @@ components:
         - $ref: '#/components/schemas/BaseObject'
         - type: object
           properties:
+            name:
```
zachclark-ccpo, line 554: What does this name represent?

packages/api/portfolioDrafts/createPortfolioDraft.ts
```diff
@@ -23,7 +23,11 @@ export async function handler(event: APIGatewayProxyEvent): Promise<APIGatewayPr
     created_at: now,
     updated_at: now,
     status: ProvisioningStatus.NOT_STARTED,
+    name: "",
```
zachclark-ccpo, line 26: Why are we setting an empty name here?

packages/api/portfolioDrafts/portfolio/createPortfolioStep.ts
```diff
@@ -70,13 +70,17 @@ export async function createPortfolioStepCommand(
       Key: {
         id: portfolioDraftId,
       },
-      UpdateExpression: "set #portfolioVariable = :portfolio, updated_at = :now",
+      UpdateExpression: `set #portfolioVariable = :portfolio, updated_at = :now,
+        #portfolioName = :portfolioName, num_portfolio_managers = :numOfManagers`,
       ExpressionAttributeNames: {
         "#portfolioVariable": PORTFOLIO_STEP,
+        "#portfolioName": "name",
```
zachclark-ccpo, line 77: I thought the portfolioName was going to be set in the PORTFOLIO_STEP

packages/api/portfolioDrafts/funding/createFundingStep.ts
```diff
@@ -35,7 +36,9 @@ export interface ClinValidationError {
  * @returns a collection of clin validation errors
  */
 export function validateFundingStepClins(fs: FundingStep): Array<ClinValidationError> {
-  return fs.clins.map(validateClin).reduce((accumulator, validationErrors) => accumulator.concat(validationErrors), []);
+  const clins: Array<Clin> = [];
+  fs.task_orders.forEach((taskOrder) => taskOrder.clins.forEach((clin) => clins.push(clin)));
```
kylelaker-ccpo, line 40: Correct me if I am wrong, but I think that this would be equivalent to:

```typescript
fs.task_orders.flatMap((taskOrder) => taskOrder.clins);
```

packages/api/portfolioDrafts/funding/createFundingStep.ts
```diff
@@ -35,7 +36,9 @@ export interface ClinValidationError {
...
+  return clins.map(validateClin).reduce((accumulator, validationErrors) => accumulator.concat(validationErrors), []);
```
kylelaker-ccpo, line 41: So then this should all be able to be:

```typescript
return fs
         .task_orders.flatMap(taskOrder) => taskOrder.clins)
         .map(validateClin)
         .reduce(accumulator, validationErrors) => accumulator.concat(validationErrors), []);
```

+/- some indentation and line-breaks there to appease the linter. In general if you're noticing a need to nest a `forEach` to create a list of elements of inner lists, that can become a `flatMap`.

packages/api/portfolioDrafts/funding/createFundingStep.ts
```diff
@@ -73,7 +76,7 @@ export async function handler(event: APIGatewayProxyEvent): Promise<APIGatewayPr
         Key: {
           id: portfolioDraftId,
         },
-        UpdateExpression: "set #stepKey = :step, #updateAtKey = :now",
+        UpdateExpression: "set #stepKey = :step, #updateAtKey = :now, num_task_orders = :numOfTaskOrders",
```
kylelaker-ccpo, line 79: This works for now during the Create but we've got to be careful about this (and similar) things during the the Updates or if we add additional task orders. I don't really remember if that's implemented yet. But we're going to need to look into [atomic counters](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/WorkingWithItems.html#WorkingWithItems.AtomicCounters) once we get closer to that.

packages/api/portfolioDrafts/createPortfolioDraft.ts
zachclark-ccpo: You could just set the name attribute when doing the createPortfolioStep, by adding a new attribute in the UpdateExpression. I disagree with setting an empty name here.

kylelaker-ccpo (changes requested):
One super small change requested to turn a forEach of a forEach to a flatMap but otherwise this is looking really good! Remember to mark the PR as ready for review so that we can move towards merging it as well.

I do want to make sure that we are testing that the `numOfX` attributes are being set correctly. Am I missing that in the tests?

atat_provisioning_wizard_api.yaml
juliusfitzhugh-ccpo: This name represents the portfolio draft defined in the `portfolio_step`. Do you think it should be named as `portfolioDraftName` or `portfolioName` to make it clear?

packages/api/portfolioDrafts/portfolio/createPortfolioStep.ts
juliusfitzhugh-ccpo: It was, but to align with the UI portfolio draft summary (see earlier comments) some of the data is being denormalized for easy access. Thanks for pointing the DELETE portfolioStep.

packages/api/portfolioDrafts/funding/createFundingStep.ts
juliusfitzhugh-ccpo: I have not used a `flatMap` previously, but that makes sense and it does make it a lot cleaner. 

I did not add that test for `numOfX`. I'll included it. Thanks for the feed back.

packages/api/portfolioDrafts/funding/createFundingStep.ts
juliusfitzhugh-ccpo: Thanks. Noted and much cleaner.

jasonburkert-ccpo: Neat.
> Equivalent to a .map() and a .flat() of depth 1.

packages/api/portfolioDrafts/createPortfolioDraft.ts
jeffsegal-ccpo: Thanks Jason. In general, we aren't dealing with a normalized data set and a typical strategy for DynamoDB is to make data queryable ahead of time rather than performing JOINs and doing expensive queries, so this is in line with that notion. I also want to make sure we are biased toward reads rather than writes. It just came down to wanting to avoid performing JSON parsing over a large number of Portfolios during a GET /portfolioDrafts. That said, it's not the end of the world to go that route. However, we're likely to need quite a few other denormalized fields anyway in order to be able to perform filtering and sorting on various fields from the My Portfolios screen.

packages/api/portfolioDrafts/funding/createFundingStep.ts
juliusfitzhugh-ccpo: Changes regarding the `flatMap` are done. Tests for `numOfX` and `portfolio_name` have been added.

atat_provisioning_wizard_api.yaml
jasonburkert-ccpo: @juliusfitzhugh-ccpo No I think name is in keeping with what we've done in other schema objects.  If we're going to deviate from that, let's discuss changing them all to be more specific/for clarity.

packages/api/models/PortfolioDraftSummary.ts
```diff
@@ -1,7 +1,17 @@
...
-  num_portfolio_managers: number;
+  portfolio_name?: string;
+  portfolio_step?: PortfolioStep;
+  num_portfolio_managers?: number;
+  funding_step?: FundingStep;
+  num_task_orders?: number;
+  application_step?: ApplicationStep;
+  num_applications?: number;
+  num_environments?: number;
 }
```
kylelaker-ccpo, line 17: These are too many fields for the PortfolioDraftSummary, per the spec. Any reason why they're here? Other interfaces later extend this interface to add some of these fields. This should have the exact items shown in the spec for this type (minus ones already specified in `BaseDocument`.

packages/api/utils/validation.ts
```diff
@@ -68,6 +69,19 @@ export function isPortfolioStep(object: unknown): object is PortfolioStep {
  * @returns true if the object has all the attributes of a {@link FundingStep}
  */
 export function isFundingStep(object: unknown): object is FundingStep {
+  if (!isValidObject(object)) {
+    return false;
+  }
+  return ["task_orders"].every((item) => item in object);
```
kylelaker-ccpo, line 75: This is a little unnecessary since it's just one item. I guess it makes it easier to enforce in the future. But this could be:

```
return isValidObject(object) && "task_orders" in object;
```

packages/api/models/PortfolioDraftSummary.ts
kylelaker-ccpo: Specifcally, `x_step` gets added on the `PortfolioDraft` interface (which extends `PortfolioDraftSummary`). We _must not_ return those fields (or imply they are undefined/null) for a `PortfolioDraftSummary` object.

And if `portfolio_name` is part of the spec, it probably shouldn't be optional. This could be a breaking API change if we ever except a `PortfolioDraftSummary` as input though.

kylelaker-ccpo (dismissed):
Lots of good fixes but some questions on these changes to `PortfolioDraftSummary`

packages/api/models/PortfolioDraftSummary.ts
jasonburkert-ccpo: @kylelaker-ccpo Agreed.  Many of these don't belong in the PD Summary object.  It should be just enough to support what's displayed for each card of the _My Portfolios_ page.

packages/api/portfolioDrafts/application/commonMockData.ts
```diff
@@ -176,3 +179,43 @@ export const mockApplicationStepsBadData: Array<ApplicationStep> = [
...
+
+/**
+ * A base Portfolio Draft Summary
+ * @returns a base portfolio summary with good data
+ */
+export function mockBasePortfolioSummary(): PortfolioDraftSummary {
+  return {
+    id: uuidv4(),
+    status: ProvisioningStatus.NOT_STARTED,
+    updated_at: new Date().toISOString(),
```
jasonburkert-ccpo, line 191: Gotta const now = new Date() to get the same value for these two props.

packages/api/portfolioDrafts/funding/createFundingStep.test.ts
```diff
@@ -142,6 +142,27 @@ describe("Successful operation tests", function () {
...
+      num_task_orders: 3,
+      status: ProvisioningStatus.NOT_STARTED,
+      id: uuidv4(),
+    };
+    ddbMock.on(UpdateCommand).resolves({
+      Attributes: mockResponse,
+    });
+    const result = await handler(validRequest);
+    const responseBody = JSON.parse(result.body);
+    expect(responseBody.task_orders.length === mockResponse.num_task_orders).toBeFalsy();
```
jasonburkert-ccpo, line 165: Since you're comparing integers, suggest using `not.toBe()` as in other comment

jasonburkert-ccpo (changes requested):
Mostly suggestions, but need to change PortfolioDraftSummary per Kyle's comments.

packages/api/models/PortfolioDraftSummary.ts
juliusfitzhugh-ccpo: Agreed, I felt the same and should have referenced the API spec. Updated accordingly.

packages/api/utils/validation.ts
juliusfitzhugh-ccpo: I was doing it to stay consistent and incase changes would be made for additional keys. But I like the one liner.

packages/api/models/PortfolioDraftSummary.ts
```diff
@@ -1,7 +1,17 @@
...
   status: ProvisioningStatus;
-  num_portfolio_managers: number;
+  portfolio_name?: string;
+  portfolio_step?: PortfolioStep;
+  num_portfolio_managers?: number;
+  funding_step?: FundingStep;
+  num_task_orders?: number;
+  application_step?: ApplicationStep;
+  num_applications?: number;
+  num_environments?: number;
```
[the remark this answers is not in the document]
juliusfitzhugh-ccpo: Updated.

packages/api/portfolioDrafts/application/commonMockData.ts
```diff
@@ -176,3 +179,43 @@ export const mockApplicationStepsBadData: Array<ApplicationStep> = [
...
+    created_at: new Date().toISOString(),
+  };
+}
```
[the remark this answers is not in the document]
juliusfitzhugh-ccpo: Agreed.

packages/api/portfolioDrafts/application/createApplicationStep.test.ts
```diff
@@ -109,6 +115,34 @@ describe("Successful operation tests", function () {
...
+    ddbMock.on(UpdateCommand).resolves({
+      Attributes: mockApplicationStepsBadData,
+    });
+    const result = await handler(validRequest);
+    const responseBody = JSON.parse(result.body);
+    expect(responseBody.applications.length === mockBadPortfolioSummary.num_applications).toBeFalsy();
+    expect(
+      responseBody.applications.flatMap((app: Application) => app.environments).length ===
+        mockBadPortfolioSummary.num_environments
+    ).toBeFalsy();
```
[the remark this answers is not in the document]
juliusfitzhugh-ccpo: Yeah, the `toBe()` and `.not.toBe()` slipped my mind.

jasonburkert-ccpo (approved):
Looks good.  My comments and Kyle's have been addressed.

jasonburkert-ccpo:
@juliusfitzhugh-ccpo Showed me branch functionality deployed in AWS environment.

PR #193, opened 2021-09-13
Normalize environment names and IDs

kylelaker-ccpo:
There are some places (like stack names) where the title-case variant of
the environment name looks and is best; however, there are some
scenarios where we need a lowercase identifier as well.

This gives us a consistent way to handle environment names and moves all
of that logic to the app level so that it is not happening in individual
stacks and constructs. Previously, we would have had problems if an
environment ID with special characters was entered but that is no longer
an issue.

Currently, we are only passing the environment ID anywhere and it is
only to one stack. The Environment Name is purely used for
CloudFormation stack names. Both are still declared and initialized so
that we can pass them in Prop structs in the future as needed. Similar
code will need to be added to the UI repo since there isn't a
particularly great way to share code between the two repositories.

Tests are added for this code; however, we have not and are not
collecting coverage data for the CDK code yet.

### DIFF
--- a/infrastructure/bin/atat-web-api.ts
+++ b/infrastructure/bin/atat-web-api.ts
@@ -5,6 +5,7 @@ import "source-map-support/register";
 import { AtatAuthStack } from "../lib/atat-auth-stack";
 import { AtatWebApiStack } from "../lib/atat-web-api-stack";
 import { getTags } from "../lib/load-tags";
+import { isString, lowerCaseEnvironmentId, normalizeEnvironmentName } from "../lib/util";
 
 const app = new cdk.App();
 if (process.env.CDK_NAG_ENABLED === "1") {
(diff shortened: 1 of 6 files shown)


kylelaker-ccpo:
If/when this is approved, I will open a PR on the UI repo to perform the same normalization.

kylelaker-ccpo:
The really important thing here is that it will help guard against things like using "DEV" on the backend, "Dev" on the frontend, and not being able to look up environments. Or ending up with a "Dev" stack and a "DEV" stack who then trip over each other when one of them tries to create resources owned by the other. With this normalization, they'll be the same stack, and so a CDK diff would show the changes to the existing stack.

It's imperfect. "Dev" != "Development" and "Prod" != "Production" but I think I am okay with that for now and we can special case special environments at a later time if needed too. I have some worries about making this overly clever too soon and limiting ourselves too much.

And as far as the tests for special characters, I don't _actually_ expect anyone to try to name an environment "Dev!" but they should be free to type whatever into the context. We just need to forbid/strip invalid characters in things like CFN stack names, SSM parameter names, and Cognito domains. And I wanted to make sure consistent rules were applied to prevent two stacks with different names getting the same lowercase ID.

zachclark-ccpo:
Looking at the "Run Test Suite" step of the "Build and Synthesize" check on GitHub, it doesn't look like `util.test.ts ` fired off (would show up as `util.ts`). What does show up is 
```
utils                       |   95.65 |    93.75 |    91.3 |   96.59 |                   
  dynamodb.ts                |     100 |      100 |     100 |     100 |                   
  errors.ts                  |     100 |      100 |     100 |     100 |                   
  response.ts                |     100 |      100 |     100 |     100 |                   
  validation.ts              |   89.47 |    92.59 |   85.71 |   91.43 | 85-88       
```
and I am 99% sure that `utils` in this case is representing `packages/api/utils` folder.

jasonburkert-ccpo (approved):
Nice normalization.

PR #198, opened 2021-09-13
Add checking for invalid Clin Numbers and Funding Amounts

jasonburkert-ccpo:
Adds checking for invalid Clin Numbers and Funding Amounts (Total Clin Value and Obligated Funds), including unit tests

Adds `isClinNumber()` and `isFundingAmount()` to validation module. including unit tests

Replaced two specific Funding Step validation messages (`TOTAL_GT_ZERO`, `OBLIGATED_GT_ZERO`) with the more general `INVALID_FUNDING_AMOUNT`

Ticket: AT-6409

### DIFF
--- a/packages/api/utils/validation.ts
+++ b/packages/api/utils/validation.ts
@@ -137,3 +137,16 @@ export function isClin(object: unknown): object is Clin {
     (item) => item in object
   );
 }
+
+/**
+ * Check whether a given string is a valid CLIN number.
+ * @param str - The string to check
+ * @returns true if the string is a valid CLIN number; false otherwise
+ */
+export function isClinNumber(str: string): boolean {
+  if (str.length !== 4) {
+    return false;
+  }
+  const num: number = parseInt(str);
+  return num >= 1 && num <= 9999;
+}
--- a/packages/api/utils/validation.ts
+++ b/packages/api/utils/validation.ts
@@ -150,3 +150,16 @@ export function isClinNumber(str: string): boolean {
   const num: number = parseInt(str);
   return num >= 1 && num <= 9999;
 }
+
+/**
+ * Check whether a given string is a valid Funding Amount.
+ * @param str - The string to check
+ * @returns true if the string is a valid Funding Amount; false otherwise
+ */
+export function isFundingAmount(str: string): boolean {
+  if (str.length === 0) {
+    return false;
+  }
+  const num: number = parseFloat(str);
+  return num > 0;
+}
--- a/packages/api/models/FundingStep.ts
+++ b/packages/api/models/FundingStep.ts
@@ -10,6 +10,7 @@ export enum ValidationMessage {
   TOTAL_GT_ZERO = "total clin value must be greater than zero",
   OBLIGATED_GT_ZERO = "obligated funds must be greater than zero",
   TOTAL_GT_OBLIGATED = "total clin value must be greater than obligated funds",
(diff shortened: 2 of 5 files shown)


jasonburkert-ccpo:
Reviewers:

1. What should be done funding amount values that include three or more digits to the right of the decimal?  See `questionableAmounts` in unit tests.  I've added a TODO in the tests here and added a comment to _ATAT Business Rules_ in confluence.  Unlikely to be a significant problem, but we should have a definitive decision.

packages/api/utils/validation.ts
```diff
@@ -137,3 +137,29 @@ export function isClin(object: unknown): object is Clin {
...
+ * @param str - The string to check
+ * @returns true if the string is a valid CLIN number; false otherwise
+ */
+export function isClinNumber(str: string): boolean {
+  if (str.length !== 4) {
+    return false;
+  }
+  const num: number = parseInt(str);
+  return num >= 1 && num <= 9999;
+}
```
kylelaker-ccpo, line 152: We should investigate changing the API spec to just require that this field be a number. That way if something stringy is provided it'll get dropped at the API GW level (since it's a structural/typing issue). We could even specifying a `minimum` and `maximum` attribute there. That way we don't even have to worry about it on the API side.

packages/api/utils/validation.ts
```diff
@@ -137,3 +137,29 @@ export function isClin(object: unknown): object is Clin {
...
+ * @param str - The string to check
+ * @returns true if the string is a valid Funding Amount; false otherwise
+ */
+export function isFundingAmount(str: string): boolean {
+  if (str.length === 0) {
+    return false;
+  }
+  const num: number = parseFloat(str);
+  return num > 0;
+}
```
kylelaker-ccpo, line 165: Feels like another thing we can fix with the API spec

kylelaker-ccpo (approved):
I shared a few thoughts on how we can handle some of the stuff around numbers. It feels like we should fix it with typing rather than parsing and validating but I understand that may be a larger and more comprehensive/complex change.

There also seemed to be a conversation on how the funding value needed to be trimmed/rounded to 2 decimal places. Did we determine the outcome of that?

kylelaker-ccpo:
> Reviewers:
> 
> 1. What should be done funding amount values that include three or more digits to the right of the decimal?  See `questionableAmounts` in unit tests.  I've added a TODO in the tests here and added a comment to _ATAT Business Rules_ in confluence.  Unlikely to be a significant problem, but we should have a definitive decision.

Missed this when I started my review. I wonder if the API should just totally reject them rather than guess what to do with arbitrary values? There is a schema [`multipleOf` ](https://json-schema.org/understanding-json-schema/reference/numeric.html#multiples) that we can use (so `0.01` for example). But of course multiples get wonky when dealing with floating point. I almost wonder if we should do something like:

```
"fundingAmount": {
    "dollars": number,
    "cents": number,
}
```

or even just expecting the amount to always be in cents and just accept integer values. Both of those are generally the safer way to handle floating point numbers within a range and especially when dealing with money.

Personally, I prefer the route of accepting dollar amounts always as number of cents.

packages/api/utils/validation.ts
jasonburkert-ccpo: I think the spec is stringy because typical CLIN values are written and understood with leading zeros like `0001`, `0002`, `0003`, etc.

But certainly if the API Gateway can ensure we get 1 - 9999 by changing the API spec, that's potentially better.

PR #203, opened 2021-09-14
Implement 'CreateApplicationStep' function

jasonburkert-ccpo:
Change model `ApplicationStep` to contain array _applications_

Add model `Application`

Relocate and organize mock data into a common location `application/commonMockData.ts `
 - includes valid objects with good data (above demarcation)
 - includes invalid objects with missing fields and bad data (below demarcation)

Change references in unit tests for GET function to use common mock data
 - remove tests for `ErrorCode.OTHER` due to recent error response improvements in e51344485bbab60be35f3e665f0299e6c4ca1314 and 9497c9b422933bdece3b424e4a9dbbc9e5fc45c6

Apply TDD and lessons from unit testing funding step to function impl and unit tests

Decompose input validation functions for increased testability

Add unit tests for individual input validation

Define enum for validation messages in `ApplicationStep.ts`

Add isApplicationStep(), isApplication(), isEnvironment() to validation module, plus unit tests for each

Create appropriate `ValidationErrorResponse` with `error_map` containing input validation errors

Ticket: AT-6467

### DIFF
--- a/packages/api/portfolioDrafts/application/createApplicationStep.ts
+++ b/packages/api/portfolioDrafts/application/createApplicationStep.ts
@@ -1,0 +1,104 @@
+import { UpdateCommand } from "@aws-sdk/lib-dynamodb";
+import { APIGatewayProxyEvent, APIGatewayProxyResult } from "aws-lambda";
+import { ApplicationStep } from "../../models/ApplicationStep";
+import { APPLICATION_STEP } from "../../models/PortfolioDraft";
+import { dynamodbDocumentClient as client } from "../../utils/dynamodb";
+import {
+  DATABASE_ERROR,
+  NO_SUCH_PORTFOLIO_DRAFT,
+  PATH_PARAMETER_REQUIRED_BUT_MISSING,
+  REQUEST_BODY_EMPTY,
+  REQUEST_BODY_INVALID,
+} from "../../utils/errors";
+import { ApiSuccessResponse, ErrorResponse, ErrorStatusCode, SuccessStatusCode } from "../../utils/response";
+import { isApplicationStep, isValidJson } from "../../utils/validation";
+import { ErrorCodes, ValidationError } from "../../models/Error";
+
+/**
+ * Submits the Application Step of the Portfolio Draft Wizard
+ *
+ * @param event - The POST request from API Gateway
+ */
+export async function handler(event: APIGatewayProxyEvent): Promise<APIGatewayProxyResult> {
+  if (!event.body) {
+    return REQUEST_BODY_EMPTY;
+  }
+
+  const portfolioDraftId = event.pathParameters?.portfolioDraftId;
+
+  if (!portfolioDraftId) {
+    return PATH_PARAMETER_REQUIRED_BUT_MISSING;
+  }
+
+  if (!isValidJson(event.body)) {
+    return REQUEST_BODY_INVALID;
+  }
+  const requestBody = JSON.parse(event.body);
+
+  // remove this, and function?
+  if (!isApplicationStep(requestBody)) {
+    return REQUEST_BODY_INVALID;
+  }
+  const now = new Date().toISOString();
+  const applicationStep: ApplicationStep = requestBody;
+
+  // NOTE This is absolutely not want we want to do long term but I am just trying to meet the ACs as simply as possible
+  const error = validate(applicationStep);
+  if (error) {
+    return new ErrorResponse({ code: error.code, message: error.message }, ErrorStatusCode.BAD_REQUEST);
+  }
+
+  const command = new UpdateCommand({
+    TableName: process.env.ATAT_TABLE_NAME ?? "",
+    Key: {
+      id: portfolioDraftId,
+    },
+    UpdateExpression: "set #portfolioVariable = :application, updated_at = :now",
+    ExpressionAttributeNames: {
+      "#portfolioVariable": APPLICATION_STEP,
+    },
+    ExpressionAttributeValues: {
+      ":application": applicationStep,
+      ":now": now,
+    },
+    ConditionExpression: "attribute_exists(created_at)",
+  });
+
+  try {
+    await client.send(command);
+  } catch (error) {
+    if (error.name === "ConditionalCheckFailedException") {
+      return NO_SUCH_PORTFOLIO_DRAFT;
+    }
+    console.error("Database error: " + error);
+    return DATABASE_ERROR;
+  }
+  return new ApiSuccessResponse<ApplicationStep>(applicationStep, SuccessStatusCode.CREATED);
+}
+
+function validate(applicationStep: ApplicationStep): ValidationError | null {
+  // NOTE: this is a quick and dirty validation attempt that will likely be replaced by a full-featured solution in AT-6549
+  // It will not provide an error map because of the complexity involved.
+  if (applicationStep.name.length < 4 || applicationStep.name.length > 100) {
+    return {
+      code: ErrorCodes.INVALID_INPUT,
+      message: "application name invalid",
+    } as ValidationError;
+  }
+  if (!applicationStep.environments || applicationStep.environments?.length === 0) {
+    return {
+      code: ErrorCodes.INVALID_INPUT,
+      message: "environments must be included",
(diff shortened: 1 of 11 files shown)


jasonburkert-ccpo:
I like the approach of having the common mock data for applications in `application/commonMockData.ts`
This is referenced by both the GET and POST functions.

Note the demarcation mid-file.
- Above the line is an implementation of Example Data in the API spec and is all valid objects with good data.
- Below the line is invalid objects (*MissingFields) for testing functions in validation module such as isApplicationStep, and objects with bad data to be used to input validation tests.

Not sure I like the file name.

packages/api/portfolioDrafts/application/createApplicationStep.ts
```diff
@@ -0,0 +1,104 @@
...
+    UpdateExpression: "set #portfolioVariable = :application, updated_at = :now",
+    ExpressionAttributeNames: {
+      "#portfolioVariable": APPLICATION_STEP,
+    },
+    ExpressionAttributeValues: {
+      ":application": applicationStep,
+      ":now": now,
+    },
+    ConditionExpression: "attribute_exists(created_at)",
+  });
```
juliusfitzhugh-ccpo: Do we need the `ReturnValues: "ALL_NEW"`? The default value is `NONE` according to the [DynamoDB docs](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_UpdateItem.html). 

I have tried changing this value for the portfolioStep on my cdk deploys, but still got the same response. Not sure if I was doing something wrong though.

jasonburkert-ccpo: Good question.  In this case, I believe this is satisfying [this part of the 201 response](https://github.com/dod-ccpo/atat-web-api/blob/c95b728959676097b7aafe76f55a7f210a864c38/atat_provisioning_wizard_api.yaml#L338) in the API spec that says the content (body) should be a JSON representation of an `ApplicationsStep`.

>           content:
>            application/json:
>              schema:
>                $ref: '#/components/schemas/ApplicationStep'
>              examples:
>                  ApplicationStepEx:
>                    $ref: '#/components/examples/ApplicationStepEx'

We seem to be doing that consistently across the board for the POST functions.  Apart from being in the spec, at present I think it's also immensely helpful to the testers.
image

zachclark-ccpo: ALL_NEW - Returns all of the attributes of the item, as they appear after the UpdateItem operation. We are mainly using this to validate that the application_step was actually added to the DB, so it should be fine to leave it in.

zachclark-ccpo:
In the yaml, you will need to change the fn arn:
```
post:
      tags:
        - portfolioDraft
      description: Submits the Application Step of the Portfolio Draft Wizard
      operationId: createApplicationStep
      parameters:
        - name: portfolioDraftId
          in: path
          description: ID of the Portfolio Draft
          required: true
          schema:
            type: string
      responses:
        '201':
          description: Successful operation
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/ApplicationStep'
              examples:
                  ApplicationStepEx:
                    $ref: '#/components/examples/ApplicationStepEx'
        '400':
          description: Invalid input
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/ValidationError'
        '404':
          description: Portfolio Draft with the given ID does not exist
      requestBody:
        description: Input data from the Application Step of the Portfolio Draft Wizard.
        required: true
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/ApplicationStep'
      x-amazon-apigateway-integration:
          httpMethod: "POST"
          uri: 
            Fn::Sub: "arn:${AWS::Partition}:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${NotImplementedFunction.Arn}/invocations"
          passthroughBehavior: "when_no_match"
          type: "aws_proxy"
```

`NotImplementedFunction.Arn` changing this to CreateApplicationStepFunction (or however it is declared in the `atat-web-api-stack.ts`

packages/api/portfolioDrafts/application/createApplicationStep.ts
```diff
@@ -0,0 +1,146 @@
...
+ */
+export async function handler(event: APIGatewayProxyEvent): Promise<APIGatewayProxyResult> {
+  const portfolioDraftId = event.pathParameters?.portfolioDraftId;
+  if (!portfolioDraftId) {
+    return PATH_PARAMETER_REQUIRED_BUT_MISSING;
+  }
+  if (!isValidUuidV4(portfolioDraftId)) {
+    return NO_SUCH_PORTFOLIO_DRAFT;
+  }
+  if (!event.body) {
```
zachclark-ccpo, line 37: I think you should use `isBodyPresent` from `validation.ts` for this instead of `!event.body`

packages/api/portfolioDrafts/application/createApplicationStep.ts
```diff
@@ -0,0 +1,146 @@
...
+  return new ApiSuccessResponse<ApplicationStep>(applicationStep, SuccessStatusCode.CREATED);
+}
+
+/**
+ * Validates the given ApplicationStep
+ * Performs input validation on all Applications and Environments contained therein
+ * @param applicationStep the object to validate
+ * @returns a collection of application validation errors
+ */
+export function validateApplicationStep(applicationStep: ApplicationStep): Array<ApplicationValidationError> {
```
zachclark-ccpo, line 87: I think these functions (`validateApplicationStep`, `validateApplication`, `validateEnvironment`) should be living in `validation.ts` . `createValidationErrorResponse` might be fine to keep in this fn, if its renamed.

zachclark-ccpo (changes requested):

packages/api/portfolioDrafts/application/createApplicationStep.ts
jasonburkert-ccpo: I believe there is some confusion due to overloaded use of the words _valid_ and _validate_ amongst ourselves.

I think as a team we need to adopt shared terminology for:
1. testing and evaluating the shape of an object and whether it meets an interface
2. performing tests on specific data within an object

I'll start a slack thread for this.

As to the specifics above, all of these are for the latter and involving input validation for data inside the objects.
I see `validation.ts` as performing the former and I don't believe these functions belong there.  Perhaps I need to avoid using the word _validate_ for these functions.  Or maybe I need to expand my interpretation of what belongs in `validation.ts`

In either case I think we should discuss these issues as a team.

packages/api/portfolioDrafts/application/createApplicationStep.ts
jasonburkert-ccpo: Will check that out.

packages/api/portfolioDrafts/application/createApplicationStep.ts
zachclark-ccpo: I agree

jasonburkert-ccpo:
> In the yaml, you will need to change the fn arn:

> `NotImplementedFunction.Arn` changing this to CreateApplicationStepFunction (or however it is declared in the `atat-web-api-stack.ts`

Function ARN updated in be8ab0b6d95fd017036908ece60b1951cfb42021
Thanks for your help.

packages/api/portfolioDrafts/application/createApplicationStep.ts
jasonburkert-ccpo: Fixed in 75ba11cb67aa4f9d5590d80382cfc19160c97ab4

packages/api/portfolioDrafts/application/createApplicationStep.ts
jasonburkert-ccpo: Preliminary result of that discussion has been to be purposeful in the distinction and organization of the _object validation_ and `data validation` solutions.

jasonburkert-ccpo: Renamed functions in 3db4345f993390b575b222786ecdcf8303b2d149

zachclark-ccpo (approved):

PR #205, opened 2021-09-15
Share the names of the created IdPs in SSM

kylelaker-ccpo:
This is required because the UI repo's stack needs to be able to query
the list of IdPs so that the Cognito User Pool Client can configure them
as the supported IdPs.

This will unblock AT-6532.

### DIFF
--- a/infrastructure/lib/atat-auth-stack.ts
+++ b/infrastructure/lib/atat-auth-stack.ts
@@ -35,10 +35,15 @@ export class AtatAuthStack extends cdk.Stack {
         },
       ],
     });
-    const ssmParam = new ssm.StringParameter(this, "UserPoolIdParameter", {
+    const poolIdParam = new ssm.StringParameter(this, "UserPoolIdParameter", {
       description: "Cognito User Pool ID",
       stringValue: cognitoAuthentication.userPool.userPoolId,
       parameterName: `/atat/${props.environmentId}/cognito/userpool/id`,
+    });
+    const idpNamesParam = new ssm.StringListParameter(this, "CognitoIdPNamesParameter", {
+      description: "Names of configured identity providers",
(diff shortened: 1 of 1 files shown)


infrastructure/lib/atat-auth-stack.ts
```diff
@@ -35,10 +35,15 @@ export class AtatAuthStack extends cdk.Stack {
         },
       ],
     });
-    const ssmParam = new ssm.StringParameter(this, "UserPoolIdParameter", {
+    const poolIdParam = new ssm.StringParameter(this, "UserPoolIdParameter", {
       description: "Cognito User Pool ID",
       stringValue: cognitoAuthentication.userPool.userPoolId,
       parameterName: `/atat/${props.environmentId}/cognito/userpool/id`,
     });
+    const idpNamesParam = new ssm.StringListParameter(this, "CognitoIdPParameter", {
```
jasonburkert-ccpo, line 43:
```suggestion
    const idpNamesParam = new ssm.StringListParameter(this, "CognitoIdPNamesParameter", {
```

jasonburkert-ccpo (changes requested):
After comparing these PRs side by side, I think this one may need to have the parameter name changed as suggested.
- API: https://github.com/dod-ccpo/atat-web-api/pull/205
- UI: https://github.com/dod-ccpo/atat-web-ui/pull/123

jasonburkert-ccpo (approved):

infrastructure/lib/atat-auth-stack.ts
kylelaker-ccpo: Good catch!

kylelaker-ccpo:
I'm going to go ahead and merge this. Even though the consumer on the UI side isn't quite ready, this sets up an additional parameter. If we end up needing additional data, we can always fix it in another PR later.

PR #209, opened 2021-09-17
Add CORS workaround

kylelaker-ccpo:
In the long-term, we need a more complete solution here. And this may
not necessarily be enough. But we're going to want to limit the actual
origins that can interact with ATAT (it should not be \*); however, today
it could be any number of sandbox, local, or "real" environments with
different names so we don't have a true static list of allowed
environments nor a convenient way to dynamically build a list. The
Access-Control-Allow-Origin header also _only_ accepts "none", "*", or a
single origin so we'd need some logic to handle dynamically setting the
header based on the `Origin` header.

We probably want to look into a pattern where this is built based on a
reaction to the event. Or we even want to see if we can push this down
into the API Spec document.

This is added as an inline function with a name to clearly represent
what its purpose is.

### DIFF
--- a/packages/api/utils/response.test.ts
+++ b/packages/api/utils/response.test.ts
@@ -44,13 +44,21 @@ describe("Validate parsing results in the same object", () => {
 });
 
 describe("Validate header handling", () => {
-  it("should not modify headers", async () => {
+  it("should not delete headers", async () => {
     const successObject = { test: "object" };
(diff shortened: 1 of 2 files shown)


jasonburkert-ccpo (approved):
Thank you for the `temporaryIncompleteCorsHeaderWorkaround`
👍

PR #225, opened 2021-09-22
Fix missing and update dependencies

kylelaker-ccpo:
We have a good chunk of dependencies that need updates so those have all
been updated. Most are dev dependencies but others include: AWS SDK, AWS
CDK, cdk-nag, and linters. We also were not installing `typescript`
specifically as a dev dependency in each project which resulted in
warnings during install.


kylelaker-ccpo:
Test failure seems to be due to https://github.com/aws/aws-sdk-js-v3/issues/2806

jeffsegal-ccpo (approved):

PR #235, opened 2021-09-24
Add table construct with PITR enabled

juliusfitzhugh-ccpo:
Create PITRTable construct that defaults to having point-in-time
recovery (PITR) enabled for NIST SP 800-53 controls.

- [x] scans with cdk-nag do not report an error for enabling PITR on DynamoDB tables

Ticket: AT-6555

### DIFF
--- a/infrastructure/lib/atat-web-api-stack.ts
+++ b/infrastructure/lib/atat-web-api-stack.ts
@@ -1,11 +1,10 @@
 import * as apigw from "@aws-cdk/aws-apigateway";
 import * as dynamodb from "@aws-cdk/aws-dynamodb";
-import * as s3 from "@aws-cdk/aws-s3";
 import * as s3asset from "@aws-cdk/aws-s3-assets";
 import * as cdk from "@aws-cdk/core";
 import { ApiDynamoDBFunction } from "./constructs/api-dynamodb-function";
 import { ApiS3Function } from "./constructs/api-s3-function";
-import { SecureBucket } from "./constructs/compliant-resources";
+import { SecureBucket, PITRTable } from "./constructs/compliant-resources";
 import { TaskOrderLifecycle } from "./constructs/task-order-lifecycle";
 import { HttpMethod } from "./http";
 import { packageRoot } from "./util";
@@ -21,12 +20,14 @@ export class AtatWebApiStack extends cdk.Stack {
     this.templateOptions.description = "Resources to support the ATAT application API";
 
     // Create a shared DynamoDB table that will be used by all the functions in the project.
-    const table = new dynamodb.Table(this, "AtatTable", {
-      partitionKey: { name: "id", type: dynamodb.AttributeType.STRING },
-      billingMode: dynamodb.BillingMode.PROVISIONED,
-      readCapacity: 1,
-      writeCapacity: 1,
-      removalPolicy: props?.removalPolicy,
+    const { table } = new PITRTable(this, "AtatTable", {
+      tableProps: {
+        partitionKey: { name: "id", type: dynamodb.AttributeType.STRING },
+        billingMode: dynamodb.BillingMode.PROVISIONED,
+        readCapacity: 1,
+        writeCapacity: 1,
+        removalPolicy: props?.removalPolicy,
+      },
     });
     const tableOutput = new cdk.CfnOutput(this, "TableName", {
       value: table.tableName,
--- a/infrastructure/lib/atat-web-api-stack.ts
+++ b/infrastructure/lib/atat-web-api-stack.ts
@@ -4,7 +4,7 @@ import * as s3asset from "@aws-cdk/aws-s3-assets";
 import * as cdk from "@aws-cdk/core";
 import { ApiDynamoDBFunction } from "./constructs/api-dynamodb-function";
 import { ApiS3Function } from "./constructs/api-s3-function";
-import { SecureBucket, PITRTable } from "./constructs/compliant-resources";
+import { SecureBucket, SecureTable } from "./constructs/compliant-resources";
 import { TaskOrderLifecycle } from "./constructs/task-order-lifecycle";
 import { HttpMethod } from "./http";
 import { packageRoot } from "./util";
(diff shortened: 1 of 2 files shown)


infrastructure/lib/constructs/compliant-resources.ts
```diff
@@ -31,3 +32,33 @@ export class SecureBucket extends cdk.Construct {
...
+  /**
+   * The properties to configure the DynamoDB table
+   */
+  tableProps: dynamodb.TableProps;
+}
+/**
+ * Creates a DynamoDB table with point-in-time recover (PITR)
+ * enabled by default.
+ */
+export class PITRTable extends cdk.Construct {
```
kylelaker-ccpo, line 46: Love this construct! We want to consider whether `PITRTable` is the best name. Eventually we will want to consider things like KMS encryption at rest and a few other things. Is there maybe a more generic name that we could give this? I don't want to bikeshed too much but I don't want to have to refactor a name change later (which could result in resource replacement)

infrastructure/lib/atat-web-api-stack.ts
```diff
@@ -21,12 +20,14 @@ export class AtatWebApiStack extends cdk.Stack {
...
-      billingMode: dynamodb.BillingMode.PROVISIONED,
-      readCapacity: 1,
-      writeCapacity: 1,
-      removalPolicy: props?.removalPolicy,
+    const { table } = new PITRTable(this, "AtatTable", {
+      tableProps: {
+        partitionKey: { name: "id", type: dynamodb.AttributeType.STRING },
+        billingMode: dynamodb.BillingMode.PROVISIONED,
+        readCapacity: 1,
+        writeCapacity: 1,
```
kylelaker-ccpo, line 28: While we're here, should we consider flipping to On Demand? We had previously done PROVISIONED to have predictable, static cost. But I think with our usage patterns, On Demand would be cheaper.

infrastructure/lib/constructs/compliant-resources.ts
juliusfitzhugh-ccpo: I agree. I originally had `SecureTable` to go along with the buckets and compliant resources, but changed it since the only thing that was really done was PITR (it wasn't really "secure" at the moment). `SecureTable` or something else?

infrastructure/lib/atat-web-api-stack.ts
juliusfitzhugh-ccpo: That makes sense. I thought this might have been previously decide, so I didn't change anything.

infrastructure/lib/constructs/compliant-resources.ts
```diff
@@ -31,3 +32,33 @@ export class SecureBucket extends cdk.Construct {
     this.bucket = bucket;
   }
 }
+
+export interface PITRTableProps {
+  /**
+   * The properties to configure the DynamoDB table
+   */
+  tableProps: dynamodb.TableProps;
```
kylelaker-ccpo, line 40:
```suggestion
  tableProps?: dynamodb.TableProps;
```

Let's make this optional

infrastructure/lib/constructs/compliant-resources.ts
kylelaker-ccpo: yeah... it's not a secure table _yet_. but it will be when we add KMS CMK support... maybe `SecureTable` is good enough for now.

juliusfitzhugh-ccpo: Okay, I'll go with that.

infrastructure/lib/constructs/compliant-resources.ts
juliusfitzhugh-ccpo: Okay.

infrastructure/lib/constructs/compliant-resources.ts
```diff
@@ -33,31 +33,31 @@ export class SecureBucket extends cdk.Construct {
...
+  constructor(scope: cdk.Construct, id: string, props: SecureTableProps) {
     super(scope, id);
 
     const table = new dynamodb.Table(this, id, {
       // passed in table configuration
       ...props.tableProps,
       // secure defaults that cannot be overridden
       pointInTimeRecovery: true,
-    });
+    } as dynamodb.TableProps);
```
kylelaker-ccpo, line 60: It feels weird that we're being forced to cast/coerce here. Is there a linter or type system error that prompted this change?

infrastructure/lib/constructs/compliant-resources.ts
juliusfitzhugh-ccpo: The `tableProps` is now optional. When I made this change the table created in the class with `dynamobdb.Table` TypeScript was giving an error for the props. I did type casting to remove the error, but not sure if this was okay.

infrastructure/lib/constructs/compliant-resources.ts
juliusfitzhugh-ccpo: yes

juliusfitzhugh-ccpo: see above comment

infrastructure/lib/constructs/compliant-resources.ts
kylelaker-ccpo: ohh i see what the problem is. i'm sorry. i gave bad advice. i forgot that `partitionKey` is a required attribute and so therefore `TableProps` is required. do you mind reverting the `?` and the cast? sorry about that

kylelaker-ccpo (approved):

infrastructure/lib/atat-web-api-stack.ts
```diff
@@ -21,12 +20,12 @@ export class AtatWebApiStack extends cdk.Stack {
...
-    const table = new dynamodb.Table(this, "AtatTable", {
-      partitionKey: { name: "id", type: dynamodb.AttributeType.STRING },
-      billingMode: dynamodb.BillingMode.PROVISIONED,
-      readCapacity: 1,
-      writeCapacity: 1,
-      removalPolicy: props?.removalPolicy,
+    const { table } = new SecureTable(this, "AtatTable", {
+      tableProps: {
+        partitionKey: { name: "id", type: dynamodb.AttributeType.STRING },
+        billingMode: dynamodb.BillingMode.PAY_PER_REQUEST,
```
kylelaker-ccpo, line 26: Just a note to others, per the CloudFormation docs, changing the billing mode requires "No Interruption". We're about to test that in Dev :)

PR #237, opened 2021-09-24
Update buckets removal policy

juliusfitzhugh-ccpo:
This is only temporary for dev purposes to ensure that deployed
buckets can be destroyed after the completion of tickets.

Ticket: AT-6557

### DIFF
--- a/infrastructure/lib/constructs/compliant-resources.ts
+++ b/infrastructure/lib/constructs/compliant-resources.ts
@@ -24,7 +24,8 @@ export class SecureBucket extends cdk.Construct {
       // secure defaults that cannot be overridden
       publicReadAccess: false,
       blockPublicAccess: s3.BlockPublicAccess.BLOCK_ALL,
-      autoDeleteObjects: false,
+      // NOTE: removed for dev purposes only, once envs are differentiated this will be fixed
+      // autoDeleteObjects: false,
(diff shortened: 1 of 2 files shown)


infrastructure/lib/constructs/compliant-resources.ts
```diff
@@ -24,7 +24,8 @@ export class SecureBucket extends cdk.Construct {
       // secure defaults that cannot be overridden
       publicReadAccess: false,
       blockPublicAccess: s3.BlockPublicAccess.BLOCK_ALL,
-      autoDeleteObjects: false,
+      // NOTE: removed for dev purposes only, once envs are differentiated this will be fixed
```
kylelaker-ccpo, line 27: Can you change this to phrasing that includes `TODO` so that it's more searchable?
```suggestion
      // TODO: Prevent auto-delete when we properly handle differences between dev and
      // higher environments
```

or something like that

kylelaker-ccpo (approved):
Thank you for this! 1 small ask if you've got time for it but otherwise we can move forward with merging.

infrastructure/lib/constructs/compliant-resources.ts
juliusfitzhugh-ccpo: Yeah, this is the second time today. I need to just go with my gut (I had `TODO` originally).

kylelaker-ccpo (approved):

jasonburkert-ccpo (approved):

PR #239, opened 2021-09-27
Feature/at 6658 integration test sample

tomarnold-ccpo:


kylelaker-ccpo:
I think this is a good start but I don't like seeing us convert the expectations set out in the OpenAPI spec into code ourselves. All of the `expect().toHaveProperty()` all have to be manually changed if we change the API. They also rely on us understanding the API spec correctly as we write the tests. And that's just... not great. We're basically duplicating the contents of the spec into the tests.

Is there a library, tool, or framework that we can pass the spec file to and have it automatically test our API? Perhaps we'd need to add some metadata, examples, or other data, but it'd be far better than adding a mountain of maintenance burden. And while this PR isn't doing that today, if we don't find a way to make the API spec the single source of truth for automatically generating as much as possible, it has the possibility of becoming that burden.

kylelaker-ccpo:
https://github.com/openapi-library/OpenAPIValidators/tree/master/packages/jest-openapi maybe?

PR #246, opened 2021-09-27
Require Authentication for all API operations

jasonburkert-ccpo:
Adds an API Gateway token-based lambda authorizer function `authorizer.ts` and unit tests

TODO
- [x] create the lambda function (API GW dev guide p.297)
- [x] create unit tests
- [ ] parse token to get principal identifier in function
- [x] select [a package](https://openid.net/developers/jwt/) for token parsing - added [jwt-decode](https://www.npmjs.com/package/jwt-decode).  Also added [jose](https://www.npmjs.com/package/jose) but may remove.
- [ ] configure function as API GW authorizer in CDK stack (API GW dev guide p.302)
- [ ] configure initial API operation `getPortfolioDrafts` to require authorizer (API GW dev guide p.304)
- [ ] configure remaining API operations
- [ ] make configurable.  `On` by default.

Ticket: AT-6533

### DIFF
--- a/packages/api/authorizer.test.ts
+++ b/packages/api/authorizer.test.ts
@@ -1,0 +1,85 @@
+import { APIGatewayAuthorizerResultContext, APIGatewayTokenAuthorizerEvent } from "aws-lambda";
+import { generatePolicyDocument, handler, PolicyEffect } from "./authorizer";
+
+describe("handler()", function () {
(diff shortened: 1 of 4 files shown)


jasonburkert-ccpo:
This was headed in the wrong direction.  Want to use an API Gateway authorizer that leverages a Cognito user pool, and not a Lambda authorizer.

PR #255, opened 2021-09-29
Add COGNITO_USER_POOLS authorizer to API spec

jasonburkert-ccpo:
Added a COGNITO_USER_POOLS type authorizer named `cognitoAuthorizer` to API spec

All API operations now require authorization through `cognitoAuthorizer`

User pool logicalId is now `AtatUserPool`

Ticket: AT-6533


atat_provisioning_wizard_api.yaml
```diff
@@ -834,6 +834,18 @@ components:
...
+      description: >-
+        Authorizes API access by token. Token obtained from custom IdP and provided in subsequent API requests.
+      type: apiKey
+      name: Authorization
+      in: header
+      x-amazon-apigateway-authtype: cognito_user_pools
+      x-amazon-apigateway-authorizer:
+        type: cognito_user_pools
+        providerARNs:
+          - Fn::GetAtt: ApplicationsApiUserPool.Arn
```
jasonburkert-ccpo, line 848: Need to replace `ApplicationsApiUserPool` here to get the ARN of the pool.

kylelaker:
Good stuff! Should just need a logical ID override!

atat_provisioning_wizard_api.yaml
jasonburkert-ccpo: Done in 10f27bd5ea564f332396af736d434b72eefcb0a7

kylelaker-ccpo:
Eventually (*not in this ticket*) we should take an action to see if we can make the authorization toggleable.

Something like:

```typescript
const forceAuth = new cdk.CfnCondition(this, "ForceAuthorization", {
  expression: cdk.Fn.conditionEquals(props?.requireAuthorization ?? true, true),
});
forceAuth.overrideLogicalId("IsAuthorizationRequired");
```

and then across the stack:

```yaml
security:
  Fn::If:
    - IsAuthorizationRequired
    - [ { cognitoAuthorizer: [] } ]
    - Ref: AWS::NoValue
```

We're going to have to play around with it to be sure and it feels like maybe it'd be messy to troubleshoot in this PR.

kylelaker-ccpo:
I put the comments above into PR #259. I'd like to see that get merged (into this branch) so that we can handle the configurability option. And this way no one has to make manual changes to the code to disable authorization during their test workflow.

jeffsegal-ccpo (approved):
Love it

jasonburkert-ccpo:
Good colab.  Thanks everyone!

PR #256, opened 2021-09-29
Map additional attributes

kylelaker-ccpo:
These are some standard attributes available from Azure AD.


kylelaker-ccpo:
This supports AT-6532 and is a fairly trivial addition. We may need to change it later. But this gives us something pretty to show in the UI while we're still ironing details.

jeffsegal-ccpo (approved):

PR #257, opened 2021-09-29
Create MFA and Password Policies for UserPool

kylelaker-ccpo:
Because the user pool theoretically could have a static user created,
this enforces strict MFA and password requirements on the pool to ensure
that if a user is for some reason created that the password policy in
place is a strict as possible.


juliusfitzhugh-ccpo (approved):
And here I thought my password length of 25 was long. Looks super secure to me.

PR #259, opened 2021-09-30
Conditionally require authorization

kylelaker-ccpo:
This adds another environment variable to override this setting. By
default, authorization is required. It can be overriden with:

```
export ATAT_REQUIRE_AUTH=false
```


jeffsegal-ccpo (approved):
Awesome.

PR #261, opened 2021-09-30
Add two scripts for developer convenience

jasonburkert-ccpo:
`reinstall.sh`
Enables install/update of nearly all software tools (nvm not included) and package dependencies in under a minute.  Runs bootstrap, build, test.  This is great for establishing a baseline especially when moving between branches.

`versions.sh`
Displays versions of relevant software tools.


jasonburkert-ccpo:
Sample output of `reinstall.sh`
[output.txt](https://github.com/dod-ccpo/atat-web-api/files/7258094/output.txt)

jasonburkert-ccpo:
I suppose if your machine is not single-purpose (used just for this project) the removal of node versions is a bit aggressive and may not be desired.  Thoughts?

reinstall.sh
```diff
@@ -0,0 +1,27 @@
...
+echo " - install the latest TypeScript"
+echo " - install all package dependencies"
+echo " - bootstrap, build, and test"
+echo " - display software tool versions"
+read -rp "Press enter to continue or Ctrl+C to cancel"
+. $NVM_DIR/nvm.sh
+. $HOME/.profile
+. $HOME/.bashrc
+echo; echo "[Remove node versions]" && rm -rf $NVM_DIR/versions/node/ && mkdir $NVM_DIR/versions/node/
+echo; echo "[Install AWS CLI]" && sudo apt install awscli
```
juliusfitzhugh-ccpo, line 23: Th script might stop here because of `apt` command for mac users. I'm assuming not too many people are using macOS, so it'll work for the majority. I use Homebrew to install and manage my packages like awscli.

reinstall.sh
```diff
@@ -0,0 +1,27 @@
...
+echo; echo "[Install node]" && cd $ATAT_WEB_API_DIR && nvm install
+echo; echo "[Install TypeScript and AWS CDK Toolkit]" && npm install -g typescript aws-cdk
+echo; echo "[Install package dependencies, bootstrap, build, and test]" && cd $ATAT_WEB_API_DIR && npm ci && npm run bootstrap && npm run build && npm run test
```
juliusfitzhugh-ccpo, line 26: I feel like this line is a little gold nugget, especially before pushing to do a PR.

juliusfitzhugh-ccpo:
This is a good baseline to get up and running and can be used for onboarding and function testing purposes. 

I do think the removal of node versions is a bit aggressive.

reinstall.sh
jasonburkert-ccpo: Definitely good feedback.  I've only run it on Ubuntu.

kylelaker-ccpo:
I am not sure that we want to put these in the root of the repo. Maybe a `scripts/` folder?

reinstall.sh
jasonburkert-ccpo: > I am not sure that we want to put these in the root of the repo. Maybe a `scripts/` folder?

That makes sense.

jasonburkert-ccpo:
`scripts/` makes sense

jasonburkert-ccpo:
Re: existing node versions, just leave them alone and install the prescribed version?

scripts/reinstall.sh
```diff
@@ -0,0 +1,25 @@
...
+echo " - install all package dependencies"
+echo " - bootstrap, build, and test"
+echo " - display software tool versions"
+read -rp "Press enter to continue or Ctrl+C to cancel"
+. $NVM_DIR/nvm.sh
+. $HOME/.profile
+. $HOME/.bashrc
+echo; echo "[Install AWS CLI]" && sudo apt install awscli
+echo; echo "[Install node]" && cd $ATAT_WEB_API_DIR && nvm install
+echo; echo "[Install TypeScript and AWS CDK Toolkit]" && npm install -g typescript aws-cdk
```
kylelaker-ccpo, line 23: Perhaps we should start encouraging accessing `tsc` and `cdk` via `npx` rather than installing the tools globally?

scripts/reinstall.sh
```diff
@@ -0,0 +1,25 @@
...
+echo " - install the latest AWS CDK Toolkit"
+echo " - install the latest TypeScript"
+echo " - install all package dependencies"
+echo " - bootstrap, build, and test"
+echo " - display software tool versions"
+read -rp "Press enter to continue or Ctrl+C to cancel"
+. $NVM_DIR/nvm.sh
+. $HOME/.profile
+. $HOME/.bashrc
+echo; echo "[Install AWS CLI]" && sudo apt install awscli
```
kylelaker-ccpo, line 21: As @juliusfitzhugh-ccpo pointed out, this will break on macOS; though, not just there but any Linux that either doesn't use apt or that doesn't package the AWS CLI. We also want to encourage folks to use AWS CLI v2 and some distros may still be packaging v1.

scripts/exportrestapi.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+if [ $# -eq 0 ]
+  then
+    echo "No arguments supplied."
+    echo "Expected Parameters:"
+    echo "  1. The Physical ID of the AWS::ApiGateway::RestApi, for example 'd82arw4bnk'"
+    exit 1
+fi
+today=$(date +"%Y-%m-%d")
+aws apigateway get-export --export-type swagger --parameters extensions='postman' --profile atat-sandbox-dev --stage-name prod --rest-api-id $1 ~/AtatWebApi-restapi-$1-postman-export-${today}.json
```
kylelaker-ccpo: Why Swagger and not OpenAPI v3?

kylelaker-ccpo: Let's not hardcode a profile

scripts/exportrestapi.sh
```diff
@@ -0,0 +1,19 @@
...
+    echo
+    echo "RestApis visible to AWS profile ${AWS_PROFILE}:"
+    aws apigateway get-rest-apis --profile ${AWS_PROFILE} | jq -c '.items[] | (.createdDate |= todate) | {rest_api_id: .id, stack_name: .tags."aws:cloudformation:stack-name", created_date: .createdDate}'
+    echo
+    echo "To switch AWS profiles, set an environment variable, for example 'export AWS_PROFILE=atat-sandbox-dev'"
+    exit 1
+fi
+today=$(date +"%Y-%m-%d")
+output=~/${AWS_PROFILE}-restapi-$1-${today}.json
+aws apigateway get-export --export-type swagger --parameters extensions='postman' --profile ${AWS_PROFILE} --stage-name prod --rest-api-id $1 ${output}
```
kylelaker-ccpo, line 17: Any reason for swagger and not OpenAPI?

scripts/exportrestapi.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+AWS_PROFILE="${AWS_PROFILE:-atat-sandbox-dev}"
+if [ $# -eq 0 ]
+  then
+    echo "No arguments supplied."
+    echo "Expected Parameters:"
+    echo "  1. The ID of a AWS::ApiGateway::RestApi, for example 'd82arw4bnk'"
+    echo
+    echo "RestApis visible to AWS profile ${AWS_PROFILE}:"
+    aws apigateway get-rest-apis --profile ${AWS_PROFILE} | jq -c '.items[] | (.createdDate |= todate) | {rest_api_id: .id, stack_name: .tags."aws:cloudformation:stack-name", created_date: .createdDate}'
```
kylelaker-ccpo, line 10: You should be able to pass the `jq` stuff to the `--query` flag of the AWS CLI?

scripts/exportrestapi.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+AWS_PROFILE="${AWS_PROFILE:-atat-sandbox-dev}"
```
kylelaker-ccpo, line 2: I don't know whether we necessarily want a default? A user that has not configured their `AWS_PROFILE` env var may also not use the same naming conventions we have documented privately in Confluence.

kylelaker-ccpo (commented):
There are useful for specific use cases but I wonder if we want to be introducing platform-specific code into the codebase. Is keeping these scripts maintained a requirement in the event we make a change that causes them to break? How will we ensure they stay up-to-date? Do we have an obligation to scan these with SonarCloud?

Would it be better for us to have a place in our internal Confluence where we document some of these snippets and provide explanation/context around them? I am just not sure we want to introduce scripts that tie us to Linux (and specifically Debian/Ubuntu-ish Linux) if that's not where all our devs are today and I am concerned about the implications of adding scripts into the git repo as far as ownership and maintainability.

We lose good version control in Confluence and the scripts are no longer public, but maybe we'd also have more space for them to feel less magical and to let users actually understand what's happening under the hood.

If we have scripts that become necessary for the build or release process, it'd make more sense to have those in the repo (but I'd still be hesitant). I know that some of these concerns are slippery-slope-type things but with SonarCloud setup and compliance questions coming up pretty heavily recently, I just want to make sure we understand the impact of adding non-app code to the repo and if it's safer/easier to store it somewhere else.

reinstall.sh
jasonburkert-ccpo: I've replaced `reinstall.sh` with `install.sh`.  It does slightly less and stays away from those OS-specific commands and tasks like installing `awscli`.

reinstall.sh
jasonburkert-ccpo: Refined previous script resulting in `install.sh` and kept this line intact.

scripts/exportrestapi.sh
jasonburkert-ccpo: I've made the improved version use environment var `AWS_PROFILE` with a sensible default.

As for the Swagger vs OpenAPI: I'm not sure.  Been using these settings for several sprints and it works with Postman.  Will look into it.

scripts/exportrestapi.sh
jasonburkert-ccpo: So, you advise removing the default?

scripts/exportrestapi.sh
jasonburkert-ccpo: I'm providing the RestApi id options for selection.  Do you mean that I could feed those options into AWS CLI so they can pick?

scripts/reinstall.sh
jasonburkert-ccpo: Good points.  I've removed the `awscli` installation from the replacement `install.sh`

scripts/exportrestapi.sh
jasonburkert-ccpo: `oas30` would be better here

> [Required] The type of export. Acceptable values are 'oas30' for OpenAPI 3.0.x and 'swagger' for Swagger/OpenAPI 2.0.

jasonburkert-ccpo: Changed in 4dfa3a8a5635fa364e99e34a44e577ecf59cf1f3

scripts/exportrestapi.sh
jasonburkert-ccpo: Changed in 4dfa3a8a5635fa364e99e34a44e577ecf59cf1f3

scripts/exportrestapi.sh
jasonburkert-ccpo: Oh, you mean do away with the `jq` call altogether.

jasonburkert-ccpo: Changed in db668a3aba43aef84786a95cb13d304c6f38a527

PR #263, opened 2021-09-30
Add initial sonar config

kylelaker-ccpo:


.github/workflows/test.yml
```diff
@@ -26,6 +26,34 @@ jobs:
...
+      - name: Checkout the repository
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Trigger SonarCloud scan
+        uses: SonarSource/sonarcloud-github-action@master
+        with:
+          projectBaseDir: ${{ matrix.dir }}
+          args: |
+            -Dsonar.typescript.tsconfigPath=/github/workspace/tsconfig.json
```
kylelaker-ccpo, line 53: This will override the value(s) you're specifying directly in the files

kylelaker-ccpo: I did this to try to let us use a shared tsconfig (which seems to no longer be feasible)

kylelaker-ccpo:
@jeffsegal-ccpo I have rebased this branch and cleaned up the history. I also updated the top comment. Unfortunately, because of the way SonarCloud and GitHub Actions work, we won't be able to really test this properly until we merge it and open a new PR to trigger the workflow. It's unclear whether the SonarCloud report will come back to the PR correctly.

This also renames/reorganizes one of the tests so if we're good with this, I'll need to update the branch protection rules.

I am going to mark this PR as ready for review.

PR #269, opened 2021-10-01
Trim PortfolioDraftSummary responses correctly

kylelaker-ccpo:
This has an unfortunate requirement to create an object that maps all
the fields of an interface to null; however, it gives us a few really
cool benefits. First is that it is an exhaustive list of all the fields
of an interface (including its parent interfaces) which means that we
can trust it when using the newly-created `pick` function. It also
enforces that the list has to be updated whenever the interface gains or
removes members.

This is all required because interfaces do not exist in JavaScript and
so the information is not available at runtime. Having a data structure
whose fields are enforced at compile time with basically that type data
exposes it at run time.

This method is currently only implemented for the GetPortfolioDrafts
route; however, it is generic and can be added and adapted to other
routes over time (which we should do because it also allows us to start
storing metadata in DynamoDB that we don't want to expose via the API).

This also adds a ProjectionExpression to the DynamoDB Scan which will
limit the attributes retrieved from the database. This utilizes our new
attribute mapping. Perhaps the subsequent pick is not really needed but
it ensures that if there is extra data in the DynamoDB response that we
trim it out at the app layer.


packages/api/models/TypeFields.ts
```diff
@@ -0,0 +1,16 @@
...
+export type ExhaustiveAttributeMap<T> = { [key in keyof T]: null };
+
+function pick<T>(obj: T, ...keys: Array<keyof T>): T {
+  const entries = keys.map((key) => [key, obj[key]]);
+  return Object.fromEntries(entries);
+}
+
+export function exhaustivePick<T>(obj: T, map: ExhaustiveAttributeMap<T>): T {
+  return pick(obj, ...(Object.keys(map) as Array<keyof T>));
+}
```
kylelaker-ccpo, line 16: This is really where the new magic is

packages/api/portfolioDrafts/getPortfolioDrafts.ts
```diff
@@ -37,6 +38,14 @@ export async function handler(event: APIGatewayProxyEvent): Promise<APIGatewayPr
   const params: ScanCommandInput = {
     TableName: TABLE_NAME,
     Limit: limit,
+    ProjectionExpression: Object.keys(portfolioDraftFields)
+      .join(", ")
+      .replace("name", "#name")
+      .replace("status", "#status"),
```
kylelaker-ccpo, line 44: This is one example of how to use part of it, to query the DB for expected fields

packages/api/portfolioDrafts/getPortfolioDrafts.ts
```diff
@@ -46,7 +55,9 @@ export async function handler(event: APIGatewayProxyEvent): Promise<APIGatewayPr
 
   try {
     const data = await client.send(command);
-    return new ApiSuccessResponse<PortfolioDraftSummary[]>(data.Items as PortfolioDraftSummary[], SuccessStatusCode.OK);
+    const items = (data.Items ?? []) as PortfolioDraftSummary[];
+    const summaries = items.map((draft) => exhaustivePick(draft, portfolioDraftFields));
```
kylelaker-ccpo, line 59: And this is how we can use the `pick` and now that we're going to have _all_ the required fields from the interface

packages/api/models/PortfolioDraftSummary.ts
```diff
@@ -12,3 +10,15 @@ export interface PortfolioDraftSummary extends BaseDocument {
...
+export const portfolioDraftFields: ExhaustiveAttributeMap<PortfolioDraftSummary> = {
+  id: null,
+  created_at: null,
+  updated_at: null,
+  status: null,
+  name: null,
+  num_portfolio_managers: null,
+  num_task_orders: null,
+  num_applications: null,
+  num_environments: null,
```
kylelaker-ccpo, line 23: If you play around with this locally and try to remove or add items to this list (without also changing the interface), you'll see that you get TypeScript errors

kylelaker-ccpo:
We are also going to want to start doing this for _input_ validation as well.

packages/api/models/PortfolioDraftSummary.ts
zachclark-ccpo: Is it possible to do
```
  num_portfolio_managers?: int,
  num_task_orders?: int,
  num_applications?: int,
  num_environments?: int,

File: packages/api/portfolioDrafts/getPortfolioDrafts.ts (hunk 3 above)
Comment (jeffsegal-ccpo) [reply packages/api/portfolioDrafts/getPortfolioDrafts.ts:59]: Looks good and glad we can apply this elsewhere as needed!

File: packages/api/models/PortfolioDraftSummary.test.ts [hunk 5]
```diff
@@ -0,0 +1,36 @@
...
+        description: "test",
+        dod_components: [],
+        portfolio_managers: [],
+      },
+    };
+    const trimmed = exhaustivePick(portfolioDraft, portfolioDraftFields);
+    expect(trimmed).not.toHaveProperty("funding_step");
+    expect(trimmed).not.toHaveProperty("portfolio_step");
+    expect(trimmed).not.toHaveProperty("application_step");
+  });
```
jeffsegal-ccpo, line 35: Probably want to have an inverse test to make sure that it does still include at least some of the attributes that we expect to have

jeffsegal-ccpo (changes requested):

packages/api/models/PortfolioDraftSummary.ts
kylelaker-ccpo: @zachclark-ccpo `number` is a type, so not really. `null` is a value and it's a value that we currently don't apply any meaning to (and is friendlier than `null` or `undefined`).

packages/api/models/PortfolioDraftSummary.test.ts
kylelaker-ccpo: Done! Also renamed the test file to make it more clear that we're not _really_ testing `PortfolioDraftSummary` itself. It's really just more of an example (and I am being lazy by not creating a specific type to use for testing).

kylelaker-ccpo:
@jeffsegal-ccpo This is ready for review again. @jasonburkert-ccpo This will have a small conflict with your PR since I chose `TaskOrder` as the example object for input validation.

packages/api/models/PortfolioDraftSummary.test.ts
jeffsegal-ccpo: No worries, sometimes I create `TestEntity` classes and sometimes I get lazy, depending on how complex that is...

jeffsegal-ccpo (approved):

zachclark-ccpo (approved):

PR #272, opened 2021-10-01
Add consistent CORS support

kylelaker-ccpo:
This ensures that all responses have the CORS headers included as well as adding an `OPTIONS` route for request pre-flighting. For now, we allow the `*` origin so that we can have different frontends point to the app; however, we will eventually need to lock this down to a trusted origin.

See https://docs.aws.amazon.com/apigateway/latest/developerguide/enable-cors-for-resource-using-swagger-importer-tool.html for more information on this implementation.

### DIFF
--- a/atat_provisioning_wizard_api.yaml
+++ b/atat_provisioning_wizard_api.yaml
@@ -87,6 +87,29 @@ paths:
           - 
             - cognitoAuthorizer: []
           - Ref: AWS::NoValue
+    options:
+      tags:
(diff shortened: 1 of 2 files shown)


juliusfitzhugh-ccpo (approved):
According to the docs you sent, looks good.

jeffsegal-ccpo (approved):

jasonburkert-ccpo (approved):

PR #275, opened 2021-10-04
Experimental change to get sonarcloud to pick up TS files again

jeffsegal-ccpo:


tsconfig.json
```diff
@@ -28,6 +28,12 @@
     "typeRoots": [
       "./node_modules/@types"
     ],
+    "baseUrl": ".",
+    "paths": {
+      "@/*": [
+        "src/*"
```
kylelaker-ccpo, line 34: This may need to be tweaked since I don't think we ever use a `src/` directory

jeffsegal-ccpo:
Looks like we got our answer: https://community.sonarsource.com/t/sonarcloud-not-picking-up-typescript-files/50832

PR #285, opened 2021-10-05
Replace a less-than-optimal regex

kylelaker-ccpo:
This was picked up by SonarCloud as a potential issue. The regex matches
one of the known problematic patterns. Replacing the regex with a
split-then-trim actually makes the intent of the code more clear anyway.

### DIFF
--- a/packages/cognito/preTokenGeneration/groups.ts
+++ b/packages/cognito/preTokenGeneration/groups.ts
@@ -10,7 +10,9 @@ export function processSamlGroupData(groups: string[]): string[] {
     allGroups.push(
       ...groupString
         .trim()
-        .split(/\s*,\s*/)
+        .split(",")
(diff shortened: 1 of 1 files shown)


jeffsegal-ccpo (approved):
Would love to have UTs here but I'm OK with punting a bit longer.

jasonburkert-ccpo (approved):

kylelaker-ccpo:
@jeffsegal-ccpo we actually should have tests for this file! Is there something https://github.com/dod-ccpo/atat-web-api/blob/feature/fix-redos-bug/packages/cognito/preTokenGeneration/groups.test.ts#L62 doesn't cover that we could get added? Ideally https://github.com/dod-ccpo/atat-web-api/blob/feature/fix-redos-bug/packages/cognito/preTokenGeneration/groups.test.ts#L92 is catching this case

PR #287, opened 2021-10-06
Add additional schema objects and required fields

juliusfitzhugh-ccpo:
Update to API spec

Add schema objects and reference in appropriate areas
- `TaskOrder`
- `Clin`
- `Application`
- `Environment`
- `Operator`

Update 'required' properties on schema objects and set
'additionalProperties' to false for validation.

Update requestBody of POST /taskOrderFiles to be required.

### DIFF
--- a/atat_provisioning_wizard_api.yaml
+++ b/atat_provisioning_wizard_api.yaml
@@ -870,14 +870,6 @@ components:
                 - "in_progress"
                 - "complete"
                 - "failed"
-          required:
-            - status
-            - name
-            - description
-            - num_portfolio_managers
-            - num_applications
-            - num_environments
-            - num_task_orders
     Error:
       type: object
       additionalProperties: false
@@ -981,7 +973,7 @@ components:
           items:
             $ref: '#/components/schemas/TaskOrder'
     TaskOrder:
-      description: Represents a Task Order Object in the Funding Step
+      description: Represents a Task Order in the Funding Step
       type: object
       additionalProperties: false
       properties:
@@ -1000,7 +992,7 @@ components:
         - task_order_file
         - clins
     Clin:
-      description: Represents a Clin Object in a Task Order
+      description: Represents a Clin in a Task Order
       type: object
       additionalProperties: false
       properties:
@@ -1009,9 +1001,9 @@ components:
         idiq_clin:
           type: string
         total_clin_value:
-          type: "integer"
+          type: number
         obligated_funds:
-          type: "integer"
+          type: number
         pop_start_date:
           type: string
           format: date


atat_provisioning_wizard_api.yaml
```diff
@@ -924,6 +938,8 @@ components:
         - name
         - dod_components
         - portfolio_managers
+        - csp
+        - description
```
jasonburkert-ccpo, line 942: Description on Portfolio Step should be optional

atat_provisioning_wizard_api.yaml
```diff
@@ -961,73 +977,114 @@ components:
...
+        clins:
+          type: array
+          item:
+            $ref: '#/components/schemas/ClinObject'
+      required:
+        - task_order_number
+        - task_order_file
+        - clins
+    ClinObject:
+      description: Represents a Clin Object in the Funding Step
```
jasonburkert-ccpo, line 1001: Clin is within the Funding Step, but I think this would be better described as part of a Task Order

atat_provisioning_wizard_api.yaml
```diff
@@ -961,73 +977,114 @@ components:
...
+          type: string
+          enum:
+            - administrator
+            - read_only
+      required:
+        - first_name
+        - last_name
+        - email
+        - access
     PortfolioDraft:
```
jasonburkert-ccpo, line 1088: Do we want to add 'additionalProperties: false' to the remaining schema objects?  PortfolioDraft, PortfolioDraftSummary, etc.

atat_provisioning_wizard_api.yaml
```diff
@@ -961,73 +977,114 @@ components:
...
-                    pop_start_date:
-                      type: string
-                      format: date
-                      example: '2021-07-01'
-                    pop_end_date:
-                      type: string
-                      format: date
-                      example: '2022-07-01'
+            $ref: '#/components/schemas/TaskOrderObject'
+    TaskOrderObject:
```
jasonburkert-ccpo, line 981:
```suggestion
    TaskOrder:
```

atat_provisioning_wizard_api.yaml
```diff
@@ -961,73 +977,114 @@ components:
...
+            - $ref: '#/components/schemas/FileMetadataSummary'
+        clins:
+          type: array
+          item:
+            $ref: '#/components/schemas/ClinObject'
+      required:
+        - task_order_number
+        - task_order_file
+        - clins
+    ClinObject:
```
jasonburkert-ccpo, line 1000:
```suggestion
    Clin:
```

atat_provisioning_wizard_api.yaml
```diff
@@ -961,73 +977,114 @@ components:
...
+        task_order_number:
+          type: string
+        task_order_file:
+          description: Metadata associated with file which was previously uploaded by POST /taskOrderFiles
+          allOf:
+            - $ref: '#/components/schemas/FileMetadataSummary'
+        clins:
+          type: array
+          item:
+            $ref: '#/components/schemas/ClinObject'
```
jasonburkert-ccpo, line 995:
```suggestion
            $ref: '#/components/schemas/Clin'
```

atat_provisioning_wizard_api.yaml
```diff
@@ -961,73 +977,114 @@ components:
...
+      additionalProperties: false
+      properties:
+        name:
+          type: string
+        description:
+          type: string
+        environments:
+          type: array
+          items:
+            $ref: '#/components/schemas/EnvironmentObject'
```
jasonburkert-ccpo, line 1046:
```suggestion
            $ref: '#/components/schemas/Environment'
```

atat_provisioning_wizard_api.yaml
```diff
@@ -840,6 +841,10 @@ components:
         updated_at:
           type: string
           format: "date-time"
+      required:
+        - id
+        - created_at
+        - updated_at
```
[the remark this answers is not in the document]
juliusfitzhugh-ccpo: @zachclark-ccpo I was thinking the same at first, but then realized that the `BaseObject` is used for the `FileMetadata` schema as well which is used in the `POST /taskOrderFiles` which does take a requestBody. Or does the `FileMetadata` not require a requestBody? I added the `required: true` for the `requestBody` that is currently there, thinking it was required.

atat_provisioning_wizard_api.yaml
juliusfitzhugh-ccpo: Added `additionalProperties: false` to all schema Objects. If there is some reason not include it on certain objects let me know.

atat_provisioning_wizard_api.yaml
```diff
@@ -867,8 +870,17 @@ components:
                 - "in_progress"
                 - "complete"
                 - "failed"
+          required:
```
jeffsegal-ccpo, line 873: Probably don't need required fields for this or any other object which is never part of a `requestBody`

atat_provisioning_wizard_api.yaml
```diff
@@ -957,81 +974,124 @@ components:
...
+      description: Represents a Clin Object in a Task Order
+      type: object
+      additionalProperties: false
+      properties:
+        clin_number:
+          type: string
+        idiq_clin:
+          type: string
+        total_clin_value:
+          type: "integer"
```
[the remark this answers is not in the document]
jeffsegal-ccpo: We can tack a few things like that onto my PR. E.g. I was also going to add `readOnly` to fields like `id`.

atat_provisioning_wizard_api.yaml
juliusfitzhugh-ccpo: Removed from `PortfolioDraftSummary`.

PR #290, opened 2021-10-07
Fetch whole history for blame data during SonarCloud job

kylelaker-ccpo:


kylelaker-ccpo:
Going to merge this. There's a chance another follow-up will be required. (cc @jeffsegal-ccpo)

PR #291, opened 2021-10-07
AT-6697: Internal Spec updates

jeffsegal-ccpo:
Updated spec to accommodate the addition of operators at the Application and Portfolio level. Also customized the available access levels and fixed one tiny bug in the `clins` property of `FundingStep`

Ticket: AT-6697

### DIFF
--- a/atat_provisioning_wizard_api.yaml
+++ b/atat_provisioning_wizard_api.yaml
@@ -1076,8 +1076,7 @@ components:
           type: string
           format: "email"
       required:
-        - first_name
-        - last_name
+        - display_name
         - email
     PortfolioOperator:
       allOf:
@@ -1103,9 +1102,9 @@ components:
       description: Represents a set of Operators who should be granted access to a Portfolio at a specific access level. Uses an enum construct for forward compatibility should other access levels be specified.
       type: string
       enum:
-        - root_administrator
+        - portfolio_administrator
     AppEnvAccess:
-      description: Represents a set of Operators who should be granted access to an Application or Environment at a specific access level. Is a single type because the same access levels are available for both Applications & Environments.
+      description: Represents a set of Operators who should be granted access to an Application or Environment at a specific access level. The same access levels are available for both Applications & Environments.
       type: string
       enum:
         - administrator


atat_provisioning_wizard_api.yaml
```diff
@@ -1061,24 +1070,47 @@ components:
...
-        access:
-          description: Operator Access Level to Environment
-          type: string
-          enum:
-            - administrator
-            - read_only
       required:
         - first_name
         - last_name
         - email
```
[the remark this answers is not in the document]
jasonburkert-ccpo: yes, required should be updated also

atat_provisioning_wizard_api.yaml
```diff
@@ -1061,24 +1070,47 @@ components:
...
+      properties:
+        access:
+          $ref: '#/components/schemas/AppEnvAccess'
+    PortfolioAccess:
+      description: Represents a set of Operators who should be granted access to a Portfolio at a specific access level. Uses an enum construct for forward compatibility should other access levels be specified.
+      type: string
+      enum:
+        - root_administrator
+    AppEnvAccess:
+      description: Represents a set of Operators who should be granted access to an Application or Environment at a specific access level. Is a single type because the same access levels are available for both Applications & Environments.
```
jasonburkert-ccpo, line 1108:
```suggestion
      description: Represents a set of Operators who should be granted access to an Application or Environment at a specific access level. The same access levels are available for both Applications & Environments.
```

atat_provisioning_wizard_api.yaml
```diff
@@ -1061,24 +1070,47 @@ components:
...
+      type: string
+      enum:
+        - administrator
+        - contributor
+        - read_only
```
jasonburkert-ccpo, line 1113: Do we need a `no_access` level?  Eric was suggesting at the top of today's call that we might have a need to store that.

atat_provisioning_wizard_api.yaml
```diff
@@ -1061,24 +1070,47 @@ components:
...
+      required:
         - access
+      properties:
+        access:
+          $ref: '#/components/schemas/AppEnvAccess'
+    PortfolioAccess:
+      description: Represents a set of Operators who should be granted access to a Portfolio at a specific access level. Uses an enum construct for forward compatibility should other access levels be specified.
+      type: string
+      enum:
+        - root_administrator
```
jasonburkert-ccpo, line 1106: portfolio_administrator?

jasonburkert-ccpo (changes requested):
See Julius comments about required fields for Operator

PR #292, opened 2021-10-08
AT-6697: Small tweak for example data

jeffsegal-ccpo:
Updated the operator examples to use display_name in lieu of first_name and last_name

Ticket: AT-6697


juliusfitzhugh-ccpo (approved):
Oh, yeah the examples, good catch

PR #300, opened 2021-10-08
Run SonarCloud under pull_request context

kylelaker-ccpo:
Technically, SonarCloud does not support external PRs:
 - https://portal.productboard.com/sonarsource/1-sonarcloud/c/50-sonarcloud-analyzes-external-pull-request
 - https://jira.sonarsource.com/browse/MMF-1371

We are intentionally skipping the SonarCloud step for Dependabot (which are external) PRs.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -70,6 +70,8 @@ jobs:
     steps:
       - name: Checkout the repository
         uses: actions/checkout@v2
+        with:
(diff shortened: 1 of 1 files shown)


kylelaker-ccpo:
I see that coverage and duplication is missing here but with the same config, it's been running on `develop` with no problem. I am curious if we see the same thing for this on merge

PR #305, opened 2021-10-08
Implement changes to internal spec for Step 4: Add Team Members

juliusfitzhugh-ccpo:
Implement changes to internal spec for Step 4: Add Team Members
- `GET getApplicationStep` route returns the added team members as operators in the application step
- ` POST createApplicationStep` route adds and update changes to operators (team members) in the application step
- The request body is valid JSON or rejected with an error.
- Both routes accept a `{portfolioDraftId}` that corresponds to a portfolio draft.

Ticket: AT-6698

### DIFF
--- a/packages/api/models/ApplicationStep.ts
+++ b/packages/api/models/ApplicationStep.ts
@@ -10,7 +10,7 @@ export enum ValidationMessage {
   INVALID_APPLICATION_NAME = "application name must be between 4 and 100 characters",
   INVALID_ENVIRONMENT_NAME = "environment name must be between 4 and 100 characters",
   INVALID_OPERATOR_NAME = "operator display_name must be between 1 and 100 characters",
-  INVALID_OPERATOR_EMAIL = "operator email must include '@' and be a '.mil' top-level domain (e.g. first.last@domain.mil)",
+  INVALID_OPERATOR_EMAIL = "operator email must include '@' and end with '.mil'",
 }
 interface BaseValidationError {
   invalidParameterName: string;


packages/api/utils/validation.ts
```diff
@@ -207,3 +223,14 @@ export function isFundingAmount(str: string): boolean {
...
 }
+
+/**
+ * Check whether an operator email is a valid email.
+ *
+ * @param email - The email to check
+ * @returns boolean - true if the email matches the regex pattern, and false if not
+ */
+export function isMilEmail(email: string): boolean {
+  const emailRegex = /^[\w.-]+@[\w-]+.mil$/i;
```
kylelaker-ccpo, line 234: So... are we 100% that all `.mil` email addresses will match this regex? Office 365 is migrating towards allowing `+` extensions and Gmail already supports them. Not sure about Exchange (on-prem). So for example, we would reject something like `email-address+atat@domain.mil`. We actually use email addresses with `+` extensions _today_ internally for our ATAT group mailboxes.

This also has a pretty significant flaw after the `@` by not allowing subdomains. So unless you're directly under `.mil`, your address cannot be used. For example, `user.name@a.b.mil` would be rejected. But we can't just add `.` to the character class because then we'd accept things with multiple `.`s back-to-back.

I believe that with edge cases (such as quoted strings and comments), the language for email addresses may not even be regular; of course, we can accept tradeoffs, but I think forbidding `+` and subdomains may not be acceptable ones.

What if we did something a little more generic like:

```typescript
// All emails have to at least contain "@"
if (!email.contains("@")) {
  return false;
}

// Users should not input email addresses with comments
// so we can safely reject those. They aren't part of the actual
// address and so get dropped when the email is being sent anyway.
if (email.contains("(")) {
  return false;
}

// We make the assumption that we will not be receiving
// email addresses with quoted local parts so that we can
// ignore the complexities that come with parsing those.
if (email.contains("\"")) {
  return false;
}

// Split into the two halves
const [localPart, domain] = email.split("@");

// Reject values that are too long
if (localPart.length > 63 || domain.length > 253) {
  return false;
}

// Domain restrictions (per DNS spec and business logic)
// No need to handle IP address logic; those are rejected
// by business logic
// We're required to end in .mil per business logic and no
// domain can have any segment > 63 characters.
domainRegex = /^([a-z0-9-]{1,63}\.)+mil$/i;
if (!domainRegex.test(domain)) {
  return false;
}

// RFC 5321 defines the local part as a Dot-string (or Quoted-string, though we've
// rejected all those possible strings), which is an Atom followed by a dot and then
// an Atom any number of times.  An Atom is defined as one or more characters
// in the atext class, which per the errata is defined to be the same as in RFC 5322;
// the definition can be found at:
//   https://datatracker.ietf.org/doc/html/rfc5322#section-3.2.3.
// This lazily encodes that as a regular expression with some logic to forbid
// back-to-back `.` or starting/ending with `.`. Trying to do this in regex would likely
// require backtracking or other issues which might lead to a ReDoS vulnerability.
localPartCharacters = /^[a-z0-9!#$%&'*+\/=?^_`{|}~.-]+$/i
return (
  !localPart.startsWith(".") &&
  !localPart.endsWith(".") &&
  !localPart.contains("..") &&
  localPartCharacters.test(localPart)
);
```

Maybe this is a little too much correctness but with the comments stripped it's not really all that much longer than some other validation performed in the same file. I've been burned before about making assumption on what an email address looks like in an environment. Of course, we probably want to make sure that the front-end isn't performing more restrictive validation.

Regardless of whether the proposed start of a solution here is viable, we do want to make sure that we accept subdomains and `+`s.

packages/api/portfolioDrafts/application/createApplicationStep.test.ts
```diff
@@ -154,45 +207,94 @@ describe("Individual Application input validation tests", function () {
   const tooShortName = "abc";
   const tooLongName =
     "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Ut eleifend lectus ut luctus ultricies nisi.";
-  it("should return error map entries when given Application has a name that is too short", () => {
-    const errors = performDataValidationOnApplication(mockApplicationStepsBadData[0].applications[0]);
+  const tooLongDisplayName =
+    "waaaaaaaaaaaaaaaaaaaaaaaaayyyyyy tooooooooooooooooooooooooooooooooooooo loooooooooonnnnnnnnnnngggggggggg";
+  it("should return error map entries when an operator has a name that is too short", () => {
```
kylelaker-ccpo, line 212:
```suggestion
  it("should return error map entries when an operator has a name that is too short", async () => {
```

Can we be consistent in this on all our tests? Others here do it, so let's do it for new tests too. If these tests need to be an exception for one reason or another, let's document why.

packages/api/portfolioDrafts/application/createApplicationStep.ts
```diff
@@ -44,8 +50,27 @@ export async function handler(event: APIGatewayProxyEvent): Promise<APIGatewayPr
...
+  }
+
+  const appEnvs: Array<Environment> = applications.flatMap((app) => app.environments);
+  const invalidEnvs = appEnvs.filter((env) => !isEnvironment(env));
+  if (invalidEnvs.length) {
+    return REQUEST_BODY_INVALID;
+  }
+
+  const appOperators: Array<AppEnvOperator> = applications.flatMap((app) => app.operators);
+  const envOperators: Array<AppEnvOperator> = appEnvs.flatMap((env) => env.operators);
```
kylelaker-ccpo, line 66: I am curious why the types of these cannot be inferred. The fact it's an array feels obvious so the real issue seems to be arround `.environments` and `.operators`. Any insight to what error message (if any) you get when removing the type info here?

packages/api/utils/validation.test.ts
```diff
@@ -186,6 +189,54 @@ describe("isEnvironment()", () => {
...
+});
+describe("isMilEmail()", () => {
+  const goodMilEmails = [
+    "good-email@testing.mil",
+    "good.email_1234567890@testing.mil",
+    "good_email_1234567890@testing_1234567-890.mil",
+  ];
+  const badMilEmails = [
+    "bad.email_1234567890@testing.au",
+    "bad-email!@#$%^&*()+=@testing.co",
```
kylelaker-ccpo, line 229: If you were to take out the parentheses and change the TLD to `.mil` this actually could be valid

packages/api/utils/validation.test.ts
```diff
@@ -186,6 +189,54 @@ describe("isEnvironment()", () => {
...
+  it.each(missingOperatorFields)("should reject Operator with missing field", (item) => {
+    expect(isOperator(item)).toEqual(false);
+  });
+});
+describe("isMilEmail()", () => {
+  const goodMilEmails = [
+    "good-email@testing.mil",
+    "good.email_1234567890@testing.mil",
+    "good_email_1234567890@testing_1234567-890.mil",
+  ];
```
kylelaker-ccpo, line 226:
```suggestion
  const goodMilEmails = [
    "good-email@testing.mil",
    "good.email_1234567890@testing.mil",
    "good_email_1234567890@testing_1234567-890.mil",
    "good_email+test@testing.mil",
    "good_email@atat.testing.mil",
  ];
```

packages/api/utils/validation.test.ts
```diff
@@ -186,6 +189,54 @@ describe("isEnvironment()", () => {
...
+  const badMilEmails = [
+    "bad.email_1234567890@testing.au",
+    "bad-email!@#$%^&*()+=@testing.co",
+    "bad_email_1234567890@testing!@#$%^&*(+=).io",
+  ];
```
kylelaker-ccpo, line 231:
```suggestion
  const badMilEmails = [
    "bad.email_1234567890@testing.au",
    "bad-email!@#$%^&*()+=@testing.co",
    "bad_email_1234567890@testing!@#$%^&*(+=).io",
    // Domains cannot have "null" segments
    "bad_email@test..mil",
    // Domains cannot start with a .
    "bad_email@.test.mil",
    // Email addresses may not begin with a dot
    ".bad_email@test.mil",
    // The local part may not end with a dot
    "bad_email.@test.mil",
    // A local part cannot have 2 or more . in a row
    "bad..email@test.mil",
    // An email address must contain a "@"
    "bad_email",
    // An email address must have characters before and after the .
    "@", "@test.mil", "bad_email@",
  ];
```

packages/api/portfolioDrafts/application/createApplicationStep.ts
```diff
@@ -44,8 +50,27 @@ export async function handler(event: APIGatewayProxyEvent): Promise<APIGatewayPr
...
+  const allOperators: Array<Operator> = [...operators, ...appOperators, ...envOperators];
+  if (allOperators.filter((oper) => !isOperator(oper)).length) {
+    return REQUEST_BODY_INVALID;
+  }
```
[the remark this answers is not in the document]
juliusfitzhugh-ccpo: Nice translation, that is correct. That is much cleaner/readable. Can't say i've seen/used `.every()` much, but will definitely use it more now.

packages/api/portfolioDrafts/application/createApplicationStep.ts
```diff
@@ -85,21 +110,26 @@ export async function handler(event: APIGatewayProxyEvent): Promise<APIGatewayPr
...
-  return applicationStep.applications
+export function performDataValidation(applicationStep: ApplicationStep): Array<ApplicationStepValidationErrors> {
+  const operatorErrors: ApplicationStepValidationErrors[] = applicationStep.operators
+    .map(performDataValidationOnOperator)
+    .reduce((accumulator, validationErrors) => accumulator.concat(validationErrors), []);
+  const applicationErrors: ApplicationStepValidationErrors[] = applicationStep.applications
     .map(performDataValidationOnApplication)
     .reduce((accumulator, validationErrors) => accumulator.concat(validationErrors), []);
+
+  return [...operatorErrors, ...applicationErrors];
```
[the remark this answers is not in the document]
juliusfitzhugh-ccpo: Used `.concat()` for consistency. Was trying to stay with the `.concat()` but was getting some TypeScript error about concat array type or something. I think I fixed it but never changed it back. I try to lean towards consistency in most cases.

packages/api/utils/validation.ts
juliusfitzhugh-ccpo: Nope, naive regex for sure. I didn't see this message until after I made another regex... 
```typescript
const emailRegex = /^[^.]([\w+-]+)(.)?([\w+-]+)?[^.]@([\w-]+.)+mil$/i
```
I initially started my approach similar to how you did, but decide to go with regex (less code). Although, like you pointed out my test cases missed a lot of edge cases. I've updated to include your approach.

Changed the strings to use `includes` since `contains` is not on strings. I left in most comments for others.

packages/api/portfolioDrafts/application/createApplicationStep.test.ts
juliusfitzhugh-ccpo: Added `async` to tests.

packages/api/utils/validation.ts
jasonburkert-ccpo: Since the API spec is specifying format for email properties, and API Gateway will enforce the most common rules, can we make this minimal, such as ending in .mil?  Unless we need it otherwise.  But I think we're already prevented from receiving just any string by API Gateway.
```yaml
        email:
          type: string
          format: "email"
```

packages/api/models/AccessLevel.ts
```diff
@@ -1,4 +1,6 @@
 export enum AccessLevel {
   ADMINISTRATOR = "administrator",
+  CONTRIBUTOR = "contributor",
   READ_ONLY = "read_only",
+  PORTFOLIO_ADMINISTRATOR = "portfolio_administrator",
```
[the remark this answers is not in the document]
juliusfitzhugh-ccpo: Plan to refactor with the above.

jasonburkert-ccpo (changes requested):
Nice work.  I appreciate the test improvements.

Biggest feedback, as discussed, is that we'll allow API Gateway to ensure email values are valid via `format: "email"`, and we can additionally check that then end in accepted TLD `.mil`.

The email regex and tests can be removed, but, I suggest holding onto them in a branch.  Maybe we'll end up using them later.

packages/api/models/ApplicationStep.ts
```diff
@@ -1,9 +1,38 @@
...
+}
+interface OperatorEmailValidationError extends BaseValidationError {
+  operatorEmail: string;
+}
+
+export type ApplicationStepValidationErrors =
+  | ApplicationValidationError
+  | EnvironmentValidationError
+  | OperatorDisplayNameValidationError
+  | OperatorEmailValidationError;
```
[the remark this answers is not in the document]
juliusfitzhugh-ccpo: Agreed.

packages/api/portfolioDrafts/application/createApplicationStep.ts
```diff
@@ -108,28 +134,64 @@ export function performDataValidationOnApplication(application: Application): Ar
...
+      invalidParameterName: "display_name",
+      invalidParameterValue: operator.display_name,
+      validationMessage: ValidationMessage.INVALID_OPERATOR_NAME,
+    });
+  }
+  // Using API Gateway "email" format to enforce the most common email rules
+  // and only ensuring the .mil top-level domain here.
+  // TODO: possibly add patterns into the internal API spec to help
+  const emailRegex = /.mil$/i;
+  if (!emailRegex.test(operator.email)) {
```
jasonburkert-ccpo, line 187: Was going to ask why not `match` here, but after reading MDN, `test` seems more appropriate.

kylelaker-ccpo:
SonarCloud smells are related to TODOs. No worries there

PR #316, opened 2021-10-13
Put CORS on API Gateway Responses

kylelaker-ccpo:
This ensures that we don't get errors about being unable to make a CORS
request when API Gateway itself sent back a 4XX/5XX response.


atat_provisioning_wizard_api.yaml
```diff
@@ -1264,6 +1264,10 @@ x-amazon-apigateway-gateway-responses:
   # response handler will be used for responses we are not explicitly customizing.
   ACCESS_DENIED:
     statusCode: 403
+    responseParameters:
+      gatewayresponse.header.Access-Control-Allow-Headers: '''*'''
```
jeffsegal-ccpo, line 1268: We'll re-use whatever parameterization strategy we devise to set these properly right?

PR #317, opened 2021-10-13
Fix crash on CreatePortfolioStep

kylelaker-ccpo:
This adds additional test cases, refactors the tests to use `it.each` (and `describe.each`) and ensures that we actually use the `isPortfolioStep` function. To so, we refactor the function for validating post requests to accept additional validators as arguments.


juliusfitzhugh-ccpo (approved):
Tested locally and tests pass, tested in deployed stack with different variations and the original error and request was successful.

No `description` required.

PR #302, opened 2021-10-08
Implement middy middleware handling

zachclark-ccpo:
### Overview
Edit createPortfolioStep to implement middy middleware for validation.
Ticket: [AT-6549](https://ccpo.atlassian.net/browse/AT-6549)

#### Usage
To add more middleware, use
```
handler.use(middlewareName)
```
This can be custom made (if so: put the custom middleware in the `/utils`) or installed with `npm`
[Middleware List](https://github.com/middyjs/middy#available-middlewares)

Check out the middy [docs](https://github.com/middyjs/middy) for instructions to implement custom middleware, and general guidance.

Ticket: AT-6549

PR #320, opened 2021-10-13
Protect against XSS attacks

jasonburkert-ccpo:
- Adds regex patterns to API spec to be more selective about accepted input using Confluence page _ATAT Business Rules (Data Fields)_ as guidance
- Adds package `xss`
- Adds custom [middy](https://github.com/middyjs/middy) middleware `xssSanitizer`
- Adds simple unit test for `xssSanitizer`
- `xssSanitizer` is used in `createPortfolioStep`
- Add new XSS unit tests for `createPortfolioStep`

TODO:

- [x] correct failing unit tests
- [ ] test in a deployed environment

Builds on #302

Ticket: AT-6615

### DIFF
--- a/packages/api/portfolioDrafts/portfolio/createPortfolioStep.test.ts
+++ b/packages/api/portfolioDrafts/portfolio/createPortfolioStep.test.ts
@@ -102,24 +102,21 @@ describe("Cross-Site Scripting (XSS) tests", () => {
   // example attacks from ASD STIG: V-222602 text
   const attacks = ["<script>alert('attack!')</script>", "<img src=x onerror='alert(document.cookie);'"];
   it.each(attacks)("should sanitize input for XSS attacks", async (attackCode) => {
-    const xssAttackPortfolioStep = {
+    const requestBody: PortfolioStep = {
       ...mockValidPortfolioSteps[0],
       description: attackCode,
       name: attackCode,
     };
-    // TODO: Place this in impl, put impl here
-    const sanitizedAttackCode = xss(attackCode);
-    // expect inputs to contain an XSS attack
-    expect(xssAttackPortfolioStep.description).toEqual(attackCode);
-    expect(xssAttackPortfolioStep.name).toEqual(attackCode);
-    // confirm attack is present in the input
-    const input = JSON.stringify(xssAttackPortfolioStep);
-    expect(input).toContain(attackCode);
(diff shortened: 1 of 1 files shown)


atat_provisioning_wizard_api.yaml
```diff
@@ -977,7 +983,9 @@ components:
       additionalProperties: false
       properties:
         task_order_number:
+          description: TO numbers are 13 characters. TO modifications are 17 characters.
           type: string
+          pattern: '^[\d]{13,17}$'
```
juliusfitzhugh-ccpo, line 988: Not sure if the full business rule is in scope for this ticket, but this pattern would probably need to modified in the future to meet the full business rule for `Task Order Number`.

juliusfitzhugh-ccpo (commented):
Looks good so far. Small comment about the `Task Order Number`.

atat_provisioning_wizard_api.yaml
tomarnold-ccpo: @juliusfitzhugh-ccpo these were changed to be more specific just a few days ago. We still haven't had the chance to update the ui with these new requirements.  Imo this is sufficient for now, and we can create another ticket to capture the additional business rules

packages/api/portfolioDrafts/xssSanitizer.ts
```diff
@@ -0,0 +1,20 @@
...
+
+// keep plain text only
+const xssOptions = {
+  whiteList: {}, // empty means remove all tags
+  stripIgnoreTag: true, // remove all tags not in whitelist
+  stripIgnoreTagBody: ["script"], // remove script tag content
+};
+const xssSanitizer = (): middy.MiddlewareObj<APIGatewayProxyEvent, APIGatewayProxyResult> => {
+  const before: middy.MiddlewareFn<APIGatewayProxyEvent, APIGatewayProxyResult> = async (request): Promise<void> => {
+    request.event.body = JSON.parse(xss(JSON.stringify(request.event.body ?? {}), xssOptions));
```
kylelaker-ccpo, line 13: Odd that this is coming in as an object already, but this does seem to be necessary. Without the JSON stringify/parse `xss` just gives the string `[object Object]`.

jasonburkert-ccpo: Thanks for the help.

PR #321, opened 2021-10-14
Add fixes for SonarCloud Code Smells

kylelaker-ccpo:
This cleans up the vast majority of non-TODO-related code smells in the
code base.


infrastructure/jest.config.js
```diff
@@ -4,5 +4,6 @@ module.exports = {
   testMatch: ['**/*.test.ts'],
   transform: {
     '^.+\\.tsx?$': 'ts-jest'
-  }
+  },
+  collectCoverage: true,
```
kylelaker-ccpo, line 8: Note: This starts collecting coverage during an `npm run test` but this _does not_ add SonarCloud coverge checking for the `infrastructure/` directory yet (and that's intentional).

packages/api/portfolioDrafts/funding/getFundingStep.ts
```diff
@@ -1,6 +1,6 @@
 import { APIGatewayProxyEvent, APIGatewayProxyResult } from "aws-lambda";
 import { ApiSuccessResponse, SuccessStatusCode } from "../../utils/response";
-import { dynamodbDocumentClient as client } from "../../utils/dynamodb";
+import { dynamodbDocumentClient as client } from "../../utils/aws-sdk/dynamodb";
```
zachclark-ccpo, line 3: I am confused what this change actually does?

packages/api/utils/response.ts
```diff
@@ -40,7 +40,7 @@ abstract class Response implements APIGatewayProxyResult {
   body: string;
   headers?: Headers;
   multiValueHeaders?: MultiValueHeaders;
-  isBase64Encoded?: boolean | undefined;
```
zachclark-ccpo, line 43: nice

packages/api/utils/response.ts
```diff
@@ -40,7 +40,7 @@ abstract class Response implements APIGatewayProxyResult {
...
+  isBase64Encoded?: boolean;
```
zachclark-ccpo, line 43: good fix here

packages/api/portfolioDrafts/createPortfolioDraft.ts
```diff
@@ -38,7 +38,7 @@ export async function handler(event: APIGatewayProxyEvent): Promise<APIGatewayPr
         Item: item,
       })
     );
-    return new ApiSuccessResponse<PortfolioDraftSummary>(item as PortfolioDraftSummary, SuccessStatusCode.CREATED);
+    return new ApiSuccessResponse<PortfolioDraftSummary>(item, SuccessStatusCode.CREATED);
```
zachclark-ccpo, line 41: Does this not produce an error with Postman?
We are asserting most of our other calls with 
```item as <type>``` in the ApiSuccessResponse

kylelaker-ccpo: SonarCloud says that this is useless and I agree. We should either specify it in the `<>` _or_ with `as` _or_ neither. But not both. Not sure how Postman would have an issue. I'll provide a URL to test against internally in a moment.

packages/api/portfolioDrafts/funding/getFundingStep.ts
jasonburkert-ccpo: Suppose I could read the docs.  🤷

PR #354, opened 2021-10-20
Implement a generic lambda that sends emails

juliusfitzhugh-ccpo:
Create a generic lambda function that can be used by more than one part of the application to send email (e.g., invite portfolio managers or members, alerts, etc.). An SMTP service provider will be used. Ensure the function has secure access to the credentials necessary to use the SMTP service.


packages/api/emails/subscribeSendEmails.ts
```diff
@@ -0,0 +1,59 @@
...
+ * Receives SQS Events for sending emails
+ *
+ * @param event - The SQS Event, triggered when a message is sent to the queue
+ */
+export async function handler(event: SQSEvent): Promise<void> {
+  // get creds for smtp service
+  const smtp = retrieveSecrets(SMTP_SECRET);
+
+  // config for smtp service using nodemailer
+  const defaults = { from: "julius.fitzhugh-ctr@ccpo.mil" };
```
kylelaker-ccpo, line 17: Let's see if we can gather requirements for what this _should_ be,

packages/api/emails/subscribeSendEmails.ts
```diff
@@ -0,0 +1,59 @@
...
+const SMTP_SECRET = process.env.SMTP;
+
+/**
+ * Receives SQS Events for sending emails
+ *
+ * @param event - The SQS Event, triggered when a message is sent to the queue
+ */
+export async function handler(event: SQSEvent): Promise<void> {
+  // get creds for smtp service
+  const smtp = retrieveSecrets(SMTP_SECRET);
```
kylelaker-ccpo, line 14: Love this!

packages/api/emails/subscribeSendEmails.ts
```diff
@@ -0,0 +1,59 @@
...
+    auth: { user: smtp.hostname, pass: smtp.password },
+  };
+  const transporter = createTransport(smtpOptions, defaults);
+
+  // batch message polled from queue (max 10)
+  const { Records } = event;
+  const messagesEmailsAndTypes = Records.map((record) => {
+    // body contains info needed to send emails (emails, emailType, etc.)
+    return { messageId: record.messageId, body: JSON.parse(record.body) };
+  });
```
kylelaker-ccpo, line 32:
```suggestion
  const messagesEmailsAndTypes = Records.map((record) =>
    // body contains info needed to send emails (emails, emailType, etc.)
    { messageId: record.messageId, body: JSON.parse(record.body) };
  );
```

packages/api/emails/subscribeSendEmails.ts
```diff
@@ -0,0 +1,59 @@
...
+  // get creds for smtp service
+  const smtp = retrieveSecrets(SMTP_SECRET);
+
+  // config for smtp service using nodemailer
+  const defaults = { from: "julius.fitzhugh-ctr@ccpo.mil" };
+  const smtpOptions = {
+    host: smtp.hostname,
+    port: Number(smtp.port),
+    secure: false,
+    // requireTSL: true,
```
kylelaker-ccpo, line 22: Before we merge, we have to use either TLS or STARTTLS

packages/api/emails/subscribeSendEmails.ts
```diff
@@ -0,0 +1,59 @@
...
+      });
+      console.log("Message sent: ", message.messageId);
+      console.log("Response: ", response);
+    }
+  } catch (error) {
+    // TODO: send to DLQ
+  }
+}
+
+async function retrieveSecrets(id: string) {
```
kylelaker-ccpo, line 51: This will need a return type (which may require defining a new type). And of course it has to include `| undefined` as an option

infrastructure/lib/atat-web-api-stack.ts
```diff
@@ -133,9 +158,35 @@ export class AtatWebApiStack extends cdk.Stack {
...
+      }).fn,
+      new ApiSQSFunction(this, "SendEmails", {
+        queue: this.emailQueue,
+        lambdaVpc: props.vpc,
+        method: HttpMethod.GET,
+        handlerPath: this.determineApiHandlerPath("subscribeSendEmails", "emails/"),
+        functionPropsOverride: {
+          deadLetterQueue: this.emailDeadLetterQueue,
+          // ? set reserved concurrency to zero?
+          reservedConcurrentExecutions: 0,
```
kylelaker-ccpo, line 181: We probably just don't want to specify this field at all, right? What was the reason for this more specifically? Just trying to understand the context.

kylelaker-ccpo (commented):
Off to a really good start! Wanted to provide a few early thoughts

packages/api/emails/subscribeSendEmails.ts
juliusfitzhugh-ccpo: Yeah, I was thinking about what the return type should be, and that a new type might be needed.

infrastructure/lib/atat-web-api-stack.ts
juliusfitzhugh-ccpo: Yeah, I didn't have it originally, but came across something in the docs about setting it to zero (and I can't find it now...). If I remember correctly it was related to errors in the lambda and not having it be retried. I think I was thinking along the lines of if we wanted errors to go straight to the DLQ, the lambda fails with no retries.

It can be removed though.

packages/api/emails/submitEmails.test.ts
```diff
@@ -0,0 +1,13 @@
+import { handler } from "./submitEmails";
+import { APIGatewayProxyEvent, APIGatewayProxyResult } from "aws-lambda";
+
+describe("Successfully post emails to queue", function () {
+  const validRequest: APIGatewayProxyEvent = {
+    body: JSON.stringify({ emails: ["test@email.mil"], emailType: "invitation" }),
+  } as any;
+  it("should successfully send message to queue and return 200", async () => {
```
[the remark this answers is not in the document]
jasonburkert-ccpo: Oh, yeah, your impl specifies .ACCEPTED

PR #375, opened 2021-10-28
Refactor and organize mock data for Portfolio Drafts steps

jasonburkert-ccpo:
This organizes mock data referenced in unit tests into files named `common*MockData.ts`
There is a file for each of the first three Portfolio Draft Wizard steps, and another file one directory up.

- Renames `application/commonMockData.ts` to `application/commonApplicationMockData.ts`
- Adds `portfolio/commonPortfolioMockData.ts`
- Adds `funding/commonFundingMockData.ts`
- Adds `portfolioDrafts/commonPortfolioDraftMockData.ts`
- Refactors existing unit tests to use the organized mock data
- Renames elements of `TypeFields` to refer to object _properties_ rather than _attributes_ or _fields_
- Adds `isPortfolioDraft()` and `isPortfolioDraftSummary()` to validation module leveraging `TypeFields.containsExactlyProperties()`
- Adds some unit tests (for new functions above, and Clins with good and bad data)
- Corrects some flawed unit tests where `.each()` was missing from tests against arrays of objects with missing fields
- Removes some ineffective tests where the purpose was not clear

Ticket: AT-6739


packages/api/portfolioDrafts/application/createApplicationStep.test.ts
```diff
@@ -45,7 +45,7 @@ const validRequest: APIGatewayProxyEvent = {
   pathParameters: { portfolioDraftId: uuidv4() },
 } as any;
 
-describe("Handle service level error", function () {
+describe("Handle service level error", () => {
```
kylelaker-ccpo, line 48: Glad to see the consistency here. Thank you!

packages/api/portfolioDrafts/commonPortfolioDraftMockData.ts
```diff
@@ -0,0 +1,196 @@
...
+  num_environments: mockApplicationStep.applications.flatMap((app) => app.environments).length,
+  name: mockPortfolioStep.name,
+};
+
+export const validRequest: APIGatewayProxyEvent = {
+  pathParameters: { portfolioDraftId: uuidv4() },
+} as any;
+
+/** ABOVE THIS LINE ARE VALID OBJECTS WITH GOOD DATA **/
+/** BELOW THIS LINE ARE INVALID OBJECTS WITH MISSING FIELDS AND BAD DATA **/
```
kylelaker-ccpo, line 42:
```suggestion
/* ABOVE THIS LINE ARE VALID OBJECTS WITH GOOD DATA */
/* BELOW THIS LINE ARE INVALID OBJECTS WITH MISSING FIELDS AND BAD DATA */
```

just to prevent any chance of these being interpreted as JSDoc/TSDoc. Love the comment explaining. We may want to (eventually) consider whether there's a way we could put these in some sort of named data structure to clarify whether they're valid/invalid rather than relying on comments. but i see that this format is used pretty consistently so that's good; and this is definitely an improvement from where we were.

packages/api/utils/validation.test.ts
```diff
@@ -206,8 +167,8 @@ describe("isApplicationStep()", () => {
   it("should accept an ApplicationStep object", async () => {
     expect(isApplicationStep(mockApplicationStep)).toEqual(true);
   });
-  it("should reject an ApplicationStep missing any field", async () => {
-    expect(isApplicationStep(mockApplicationStepsMissingFields)).toEqual(false);
+  it.each(mockApplicationStepsMissingFields)("should reject an ApplicationStep missing any field", async (item) => {
```
kylelaker-ccpo, line 170: Love to see more usage of `each`

kylelaker-ccpo (changes requested):
Just a really minor change around comment formatting that I think may need to be done in a couple of places.

packages/api/portfolioDrafts/application/createApplicationStep.test.ts
jasonburkert-ccpo: Did a global replace for this.

packages/api/utils/validation.test.ts
jasonburkert-ccpo: This test was falsely passing because the array wasn't an `ApplicationStep`.  Fixed this instance and a few others that were similar.

packages/api/portfolioDrafts/commonPortfolioDraftMockData.ts
jasonburkert-ccpo: Fixed in c58f0e988f9ce67bf1d68d860bbc01c9c1aeb6f9

packages/api/utils/validation.test.ts
jasonburkert-ccpo: This, among other changes in the PR, increased our test count to 280.

kylelaker-ccpo (approved):

PR #376, opened 2021-10-28
created pre auth trigger & associated lambda for the purpose of creat…

matthewdvertola-ccpo:
…ing logs on auth events


kylelaker-ccpo:
This is looking really good! I think the only thing we might be "missing" is an explanation somewhere of _why_ a Post Authentication Event was chosen -- that can probably be in a ticket comment. And for context, that reason is that it fires _after_ Pre Token Generation, so the groups will already be correct, and it always fires for successful auth. It's also documented by AWS as being the preferred event for capturing metrics and logs.

I will approve this once it's marked as ready for review.

kylelaker-ccpo (approved):

kylelaker-ccpo:
Good stuff on this! I know you had to do quite a bit of testing to find the right event on this. Really glad we were able to break this out into a Post Authentication event.

PR #380, opened 2021-10-29
Update Funding Step API operations to use middy validation

zachclark-ccpo:
This replaces API Gateway validation with middy-based validation begun in #302.
API operations for Funding Step of the Portfolio Draft Wizard were changed to perform validation of business rules using middy middleware.

- `createFundingStep`
- `getFundingStep`

### TODO

- [x] Update YAML spec to include empty array validation, minimum value clin validation
- [x] Update fundingStep to include business validation
- [x] Clean up requestValidation (right now its still using a lot of the old FundingStep validation, some of it we don't need)
- [x] Write tests for createFundingStep, getFundingStep
- [x] Update getFundingStep
- [x] Update YAML spec to include date-time validation? This was removed earlier when we were still doing API gateway validation.

See also #381

Ticket: AT-6724

PR #381, opened 2021-10-30
Update Application Step API operations to use middy validation

juliusfitzhugh-ccpo:
This replaces API Gateway validation with middy-based validation begun in #302.
API operations for Application Step of the Portfolio Draft Wizard were changed to perform validation of business rules using middy middleware.

- `createApplicationStep`
- `getApplicationStep`

### TODO

- [x] Replace API Gateway Validation in the API spec with middy validation
- [x] Implement middy on `getApplicationStep` and `createApplicationStep` operations
- [x] Add unit tests to enforce business rules in addition to shape
- [x] Ensure all business rules are met.  
- [x] Test the endpoint with Postman

See also #380

Ticket: AT-6723


packages/api/portfolioDrafts/application/createApplicationStep.ts
```diff
@@ -101,111 +63,22 @@ export async function handler(event: APIGatewayProxyEvent): Promise<APIGatewayPr
...
-}
-
-/**
- * Performs data validation on the given Operator object
- * @param operator the object containing data to validate
- * @returns a collection of operator validation errors
- */
-export function performDataValidationOnOperator(operator: Operator): Array<ApplicationStepValidationErrors> {
-  const errors = Array<ApplicationStepValidationErrors>();
-  if (operator.display_name.length < 1 || operator.display_name.length > 100) {
```
zachclark-ccpo, line 175: You might need to add this validation to the yaml spec (add exclusiveMinimum: 1, exclusiveMaximum:100).

juliusfitzhugh-ccpo: I added the regex pattern to the spec, like we discussed.

packages/api/portfolioDrafts/application/createApplicationStep.ts
```diff
@@ -101,111 +74,18 @@ export async function handler(event: APIGatewayProxyEvent): Promise<APIGatewayPr
...
-  });
-  return new ValidationErrorResponse("Invalid input", invalidProperties);
-}
+const wrappedApplicationStepSchema = {
+  type: "object",
+  required: ["body"],
+  properties: {
+    body: applicationStepSchema,
+  },
+};
```
zachclark-ccpo, line 85: To avoid repetition, I wrote a function (utils/schemaWrapper.ts) to wrap the schema and save us 6 lines: 
```
export function wrapSchema(schema: Record<string, unknown>): Record<string, unknown> {
  const wrappedSchema = {
    type: "object",
    required: ["body"],
    properties: {
      body: schema,
    },
  };
  return wrappedSchema;
}
```
You would then call it with:
```
.use(validator({ inputSchema: wrapSchema(schema.ApplicationStep)}))
```

juliusfitzhugh-ccpo: Awesome! I was considering a func for this since it would be the same everywhere middy is being used.

packages/api/portfolioDrafts/application/getApplicationStep.ts
```diff
@@ -51,3 +62,5 @@ export async function handler(event: APIGatewayProxyEvent): Promise<APIGatewayPr
     return DATABASE_ERROR;
   }
 }
+
+export const handler = middy(baseHandler).use(cors({ headers: "*", methods: "*" }));
```
kylelaker-ccpo, line 66: @juliusfitzhugh-ccpo and @zachclark-ccpo Can you coordinate on a place to put this CORS configuration? When the time comes to limit the headers/methods/origins, I'd like to be able to change it in one place rather than having to modify all our functions. So ideally, something like:

```ts
// TODO(AT-XXXX): Limit the origin(s) permitted by the CORS policy
export const CORS_CONFIGURATION: Type = {
  origin: "*",
  headers: "*",
  methods: "*",
}
```

packages/api/utils/requestValidation.ts
```diff
@@ -29,3 +32,69 @@ export function shapeValidationForPostRequest<T>(
...
+    hasPortfolioAdminRole,
+    hasAdminForEachApplication,
+    hasAdminForEachEnvironment,
+    applicationsWithNoAdmin,
+    environmentsWithNoAdmin,
+    acceptableAdministratorRoles,
+  };
+}
+
+export function adminRoleFound(operators: Operator[]): boolean {
```
kylelaker-ccpo, line 93: Can we add documentation here? I am confused about why we return `true` if _any_ admin operator is found

packages/api/portfolioDrafts/application/getApplicationStep.ts
```diff
@@ -37,7 +48,7 @@ export async function handler(event: APIGatewayProxyEvent): Promise<APIGatewayPr
       })
     );
     if (!result.Item) {
-      return NO_SUCH_PORTFOLIO_DRAFT;
+      return NO_SUCH_PORTFOLIO_DRAFT_FOUND;
```
zachclark-ccpo, line 51: You should use `NO_SUCH_PORTFOLIO_DRAFT` from `api/utils/errors.ts` instead of redeclaring it here.

packages/api/utils/requestValidation.ts
juliusfitzhugh-ccpo: Added something like the below. @kylelaker-ccpo does that help clear up the confusion? I also added documentation for the `findAdministrators` to help state what the business rules are for admin roles. 

```typescript
/**
 * Checks if at least one operator admin is present in a group of operators
 * according to the business rules.
 *
 * Only 'portfolio_administrator' operators are found at the top level, and
 * 'administrator' operators are found at the app and env levels.
 *
 * @param operators - operators at one of the three application step levels
 * @returns - true if one operator admin found, and false otherwise
 */
```

packages/api/portfolioDrafts/application/getApplicationStep.ts
```diff
@@ -18,14 +27,16 @@ export const NO_SUCH_PORTFOLIO_DRAFT = new OtherErrorResponse(
...
-  if (!isPathParameterPresent(portfolioDraftId)) {
-    return PATH_PARAMETER_REQUIRED_BUT_MISSING;
-  }
-  if (!isValidUuidV4(portfolioDraftId)) {
-    return NO_SUCH_PORTFOLIO_DRAFT;
+export async function baseHandler(
+  event: ApiGatewayEventParsed<ApplicationStep>,
+  context?: Context
+): Promise<APIGatewayProxyResult> {
+  const setupResult = shapeValidationForPostRequest<ApplicationStep>(event);
```
zachclark-ccpo, line 34: This is a GET not a POST, we may want to rename it to something like: `requestShapeValidation`

packages/api/portfolioDrafts/application/getApplicationStep.ts
juliusfitzhugh-ccpo: I added `_FOUND` because there were two `NO_SUCH_PORTFOLIO_DRAFT` being imported into the test file causing a naming collision. I added `_FOUND` to the one found on lines 19-22 of `getApplicationStep.ts`. One has a status code of `404` while the other is `400`. Should we still remove one and accept a 404 status code for both cases?

packages/api/portfolioDrafts/application/getApplicationStep.ts
juliusfitzhugh-ccpo: I think `requestShapeValidation` is good. I know you are changing it on your branch, so I won't do it here.

packages/api/portfolioDrafts/application/getApplicationStep.ts
juliusfitzhugh-ccpo: Adding on this branch. Coordinated with @zachclark-ccpo and we decided to add it to a new file `/api/utils/corsConfig.ts`.  Created an `Options` interface that the `cors` function expects.  I also added `AT-6712` for the TODO, is that the correct one or should it be added to a different ticket?

packages/api/portfolioDrafts/application/getApplicationStep.ts
zachclark-ccpo: We should keep the `404` one because `404` is the statusCode for `NOT_FOUND`

packages/api/portfolioDrafts/application/getApplicationStep.ts
```diff
@@ -1,14 +1,23 @@
...
+import cors from "@middy/http-cors";
+import { ApiGatewayEventParsed } from "../../utils/eventHandlingTool";
+import { shapeValidationForPostRequest } from "../../utils/requestValidation";
 
 // Note that API spec calls for 400 and not 404
-export const NO_SUCH_PORTFOLIO_DRAFT = new OtherErrorResponse(
+export const NO_SUCH_PORTFOLIO_DRAFT_FOUND = new OtherErrorResponse(
   "The given Portfolio Draft does not exist",
   ErrorStatusCode.BAD_REQUEST
 );
```
zachclark-ccpo: We should use `NO_SUCH_PORTFOLIO_DRAFT` from errors.ts instead here.

zachclark-ccpo (changes requested):
Looks good, just request 1 quick fix.

packages/api/portfolioDrafts/application/getApplicationStep.ts
juliusfitzhugh-ccpo: I think I just pushed that change. I removed `NO_SUCH_PORTFOLIO_DRAFT_FOUND`.

zachclark-ccpo (approved):

packages/api/utils/requestValidation.ts
kylelaker-ccpo: Looks good! I think we're missing an explanation that this is because an environment/portfolio _must_ contain an administrator (which is the context I lacked)

packages/api/utils/requestValidation.ts
```diff
@@ -29,3 +32,91 @@ export function shapeValidationForPostRequest<T>(
...
+  const applicationsWithNoAdmin: number[] = [];
+  const environmentsWithNoAdmin: EnvironmentWithNoAdminProps[] = [];
+  applications.forEach((app, appIndex) => {
+    // apps without admin roles
+    !adminRoleFound(app.operators) && applicationsWithNoAdmin.push(appIndex);
+
+    // env without admin roles
+    app.environments.forEach((env, envIndex) => {
+      !adminRoleFound(env.operators) && environmentsWithNoAdmin.push({ appIndex, envIndex });
+    });
```
kylelaker-ccpo, line 75: I am concerned that this is a little too clever, relying on short-circuiting like this. How about:

```
applicationsWithNoAdmin = applications.filter((app) => !adminRoleFound(app.operators));
environmentsWithNoAdmin = applications.flatMap((app) => app.environments).filter((env) => !adminRoleRound(env.operators));
```
?

kylelaker-ccpo (changes requested):

packages/api/utils/requestValidation.ts
kylelaker-ccpo: If you don't want to go that route, please switch to using an `if (COND) { }` instead

packages/api/utils/requestValidation.ts
```diff
@@ -29,3 +32,91 @@ export function shapeValidationForPostRequest<T>(
...
+ * @returns - true if one operator admin found, and false otherwise
+ */
+export function adminRoleFound(operators: Operator[]): boolean {
+  return operators.some(isAdministrator);
+}
+
+// TODO: move into Operators.ts and update with changes from Operators.ts
+export function isAdministrator(operator: Operator): boolean {
+  return operator.access === AccessLevel.PORTFOLIO_ADMINISTRATOR || operator.access === AccessLevel.ADMINISTRATOR;
+}
```
kylelaker-ccpo, line 122: Can we just do this now?

packages/api/utils/requestValidation.ts
kylelaker-ccpo: You actually might need something a little more involved like:

```ts
applicationsWithNoAdmin = applications.map((app, appIndex) => { app, appIndex }).filter(({ app }) => !adminRoleFound(app.operators)).map(({app, appIndex}) => appIndex);
```

packages/api/portfolioDrafts/application/getApplicationStep.ts
jasonburkert-ccpo: That's great.  Looks like the right ticket.  I linked to the above in that ticket.

packages/api/portfolioDrafts/application/getApplicationStep.ts
jasonburkert-ccpo: The above is sound thinking, however, this is due to the API spec.

These operations all use http status 404 to indicate when their respective step does not exist/wasn't found.
`GET /portfolioDrafts/{portfolioDraftId}/portfolio`
`GET /portfolioDrafts/{portfolioDraftId}/funding`
`GET /portfolioDrafts/{portfolioDraftId}/application`

For those same operations http status 400 is used to indicate when the requested portfolio draft does not exist/wasn't found.

For example, from `.../application`

>         '404':
>          description: Application Step not found for this Portfolio Draft
>        '400':
>          description: The given Portfolio Draft does not exist

We need the similar messages paired with different status codes for these cases.

packages/api/utils/requestValidation.ts
jasonburkert-ccpo: Agree.  #387 recently merged, so go ahead an do this now.

packages/api/utils/requestValidation.ts
juliusfitzhugh-ccpo: Understood. Thanks for asking so I could add some context to make things clearer.

packages/api/portfolioDrafts/application/getApplicationStep.ts
juliusfitzhugh-ccpo: Thanks for providing that explanation. I added the 400 error to `errors.ts` and updated 
the names with `_404` and `_400` to differentiate.

packages/api/utils/requestValidation.ts
juliusfitzhugh-ccpo: Updated with your more involved suggestion and removed the short-circuiting. 
It was not my intent to be clever with this.

packages/api/utils/requestValidation.ts
juliusfitzhugh-ccpo: Thanks for getting that merged @jasonburkert-ccpo. My intent was to do that 
once those changes were merged with the expectation that the merge would 
happen soon and I could do it on this branch.

packages/api/portfolioDrafts/application/getApplicationStep.ts
```diff
@@ -51,3 +51,5 @@ export async function handler(event: APIGatewayProxyEvent): Promise<APIGatewayPr
     return DATABASE_ERROR;
   }
 }
+
+export const handler = middy(baseHandler).use(cors(CORS_CONFIGURATION));
```
kylelaker-ccpo, line 55: Why don't we use the JSONErrorHandlerMiddleware for the GET?

packages/api/utils/corsConfig.ts
```diff
@@ -0,0 +1,19 @@
+interface Options {
```
kylelaker-ccpo, line 1: I like this but I have two questions:
- is this or a similar type already available from the Middy `cors` package?
- if not, can we give this a better name like `CorsHeaders` or something?

packages/api/portfolioDrafts/application/getApplicationStep.ts
juliusfitzhugh-ccpo: @kylelaker-ccpo I was under the impression that we didn't really need it from a previous discussion, so I did not included it. I will add it. I believe @zachclark-ccpo is working on a unified response with AT-6734 that might not include the `JSONErrorHandlerMiddleware`.

packages/api/utils/corsConfig.ts
juliusfitzhugh-ccpo: This is basically the type from the middy `cors` package. I tried to import and use it, but I was not able to, so I copied it.

packages/api/portfolioDrafts/application/getApplicationStep.ts
kylelaker-ccpo: Oh I'm not challenging the decision -- I'm just curious about the justification

kylelaker-ccpo (approved):

packages/api/portfolioDrafts/application/getApplicationStep.ts
juliusfitzhugh-ccpo: I'll leave it in to stay consistent with the others routes.

atat_provisioning_wizard_api.yaml
```diff
@@ -1054,8 +1066,10 @@ components:
           pattern: '^[a-zA-Z\d _-]{4,100}$'
         description:
           type: string
+          pattern: ^[\w\d !@#$%^&*_|:;,'.-]{0,300}$
```
jasonburkert-ccpo, line 1069: Business Rules say:
> Alphanumeric (A-z, 0-9) + space + special characters (interpreted as !@#$%^&*_|:;,'.-)
> Max length: 300 characters
👍

❔ The description is optional.  Is it still possible to omit a description?

atat_provisioning_wizard_api.yaml
```diff
@@ -1074,47 +1087,48 @@ components:
...
-      required:
         - access
-      properties:
-        access:
-          $ref: '#/components/schemas/PortfolioAccess'
     AppEnvOperator:
-      allOf:
-        - $ref: '#/components/schemas/Operator'
-      type: object
+      description: Represents an Application or Environment Operator
```
jasonburkert-ccpo, line 1116:
```suggestion
      description: Represents an operator of an Application or Environment
```

atat_provisioning_wizard_api.yaml
```diff
@@ -1074,47 +1087,48 @@ components:
...
       additionalProperties: false
-      required:
-        - access
+      type: object
       properties:
+        display_name:
+          type: string
+          pattern: '^[a-zA-Z\d ,.-]{1,100}$'
```
jasonburkert-ccpo, line 1122: Business Rules page says:

> Alphanumeric (A-z, 0-9) + space + comma (,) + period (.) + hyphen (-)
> Max length: 100 characters

👍

atat_provisioning_wizard_api.yaml
```diff
@@ -1074,47 +1087,48 @@ components:
       properties:
         name:
           type: string
-          pattern: '^[a-zA-Z\d _-]{4,100}$'
+          pattern: '^[a-zA-Z\d ,.-]{1,100}$'
```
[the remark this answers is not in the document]
juliusfitzhugh-ccpo: Good catch thanks!

juliusfitzhugh-ccpo:
@jasonburkert-ccpo changes addressed in PR #405 .

PR #382, opened 2021-11-01
Add OpenSearch

kylelaker-ccpo:
Research spike

Ticket: AT-6804


kylelaker-ccpo:
Current status: each DB write triggers the Lambda and the data is successfully put into OpenSearch. Still need to work on a good way to get the data _out_.

PR #386, opened 2021-11-01
Update AWS SDK to 3.39.0

juliusfitzhugh-ccpo:


packages/email/package.json
```diff
@@ -28,7 +28,6 @@
     "xss": "^1.0.10"
   },
   "devDependencies": {
-    "@aws-sdk/types": "3.37.0",
```
kylelaker-ccpo, line 31: @juliusfitzhugh-ccpo can you just briefly confirm that this is intentional?

juliusfitzhugh-ccpo: @kylelaker-ccpo Yes, this was intentional. I noticed that the dependences that required `@aws-sdk/types` had it in the `package-lock.json` as `required`. The version would also have to chance to `3.38.0`. It did not seem like it was needed as a `devDependence` as I originally thought, so I removed.

kylelaker-ccpo (approved):

PR #402, opened 2021-11-08
Use an Aspect to set Removal Policy

kylelaker-ccpo:
Currently, disjoint mechanisms are used to set the removal policy for
constructs across the application. This uses an Aspect to consistently
apply a removal policy whenever we're dealing with a sandbox
environment. For resources in more permanent environments, the default
retention policies are used. In general, the defaults are generally to
retain resources that hold data such as S3 buckets, DynamoDB tables, and
OpenSearch Domains; however, we need to delete those in our Sandbox
environments. This ensures the proper policy is applied to all resources
in those sandboxes while leaving the default for dev/staging/prod, etc.

While this does now special case "dev", "staging", and "prod", which I
am not a fan of, it resolves the issue of "dev" currently having removal
policies set to DESTROY incorrectly. The same TODO of needing to
determine a long-term fix for this still applies; new tech debt is not
added, it's just moved.

This was originally started in #382.

### DIFF
--- a/infrastructure/bin/atat-web-api.ts
+++ b/infrastructure/bin/atat-web-api.ts
@@ -3,19 +3,22 @@ import * as ec2 from "@aws-cdk/aws-ec2";
 import * as cdk from "@aws-cdk/core";
 import { NIST80053R4Checks } from "cdk-nag";
 import "source-map-support/register";
+import { RemovalPolicySetter } from "../lib/aspects/removal-policy";
 import { AtatIamStack } from "../lib/atat-iam-stack";
 import { AtatNetStack } from "../lib/atat-net-stack";
 import { AtatWebApiStack } from "../lib/atat-web-api-stack";
 import { getTags } from "../lib/load-tags";
-import { isString, lowerCaseEnvironmentId, normalizeEnvironmentName } from "../lib/util";
+import {
+  isString,
+  lowerCaseEnvironmentId,
+  normalizeEnvironmentName,
+  isPossibleTemporaryEnvironment,
+} from "../lib/util";
 
 const app = new cdk.App();
 if (process.env.CDK_NAG_ENABLED === "1") {
   cdk.Aspects.of(app).add(new NIST80053R4Checks({ verbose: true }));
 }
-// TODO: Dynamically set this based on whether this is a sandbox environment,
-// a dev environment, or something else. For now, destroy everything on delete.
-const removalPolicy = cdk.RemovalPolicy.DESTROY;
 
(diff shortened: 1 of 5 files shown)


infrastructure/bin/atat-web-api.ts
```diff
@@ -32,13 +30,21 @@ if (app.node.tryGetContext("TicketId")) {
 const environmentName = normalizeEnvironmentName(environmentParam);
 const environmentId = lowerCaseEnvironmentId(environmentParam);
 
+// Delete all resources in developer-specific sandbox environments by default, while
+// using the default for the specially-named long-lived "sandbox" resources, used
+// for persistent IAM configuration, etc.
+// TODO?: Make dev/staging/prod not "special" names -- introduce a different flag to
+// signify whether something is temporary
+if (!["sandbox", "dev", "staging", "prod"].includes(environmentId)) {
```
jeffsegal-ccpo, line 38: What about adding a new flag to indicate that this is a disposable env? That way we don't open ourselves up to prod getting destroyed because someone changed it from `prod` to `production`.

jeffsegal-ccpo (changes requested):

infrastructure/bin/atat-web-api.ts
kylelaker-ccpo: I guess we could also write that as one regex if we wanted to:

```ts
/^(sandbox|dev|stage?|prod)/.test(environmentId)
```

jasonburkert-ccpo: Checking for _starts with_ rather than _exact match_ does seem less problematic until that TODO is done.

infrastructure/lib/util.ts
```diff
@@ -103,3 +103,18 @@ export function apiSpecOperationMethod(operationId: string): HttpMethod {
...
+ * absolutely temporary.
+ *
+ * @param environmentId the environment id to check
+ * @returns true if the environment ID looks like it is for a temporary environment
+ */
+// TODO: Replace with a more complete means of indicating an environment is not a
+// long-lived environment.
+export function isPossibleTemporaryEnvironment(environmentId: string): boolean {
+  return !/^(sandbox|dev|stag|prod)/.test(environmentId);
+}
```
kylelaker-ccpo, line 120: @jeffsegal-ccpo and @jasonburkert-ccpo this replaces the code block we were previously talking about.

jasonburkert-ccpo (approved):

jeffsegal-ccpo (approved):

PR #418, opened 2021-11-10
Integrate error handling middleware into recently refactored functions

zachclark-ccpo:
## Overview
The following functions will be updated to include errorHandlingMiddleware:

- createApplicationStep
- createPortfolioStep
- getApplicationStep
- getFundingStep
- createFundingStep (already had it, small improvements).

### Brief history of `errorHandlingMiddleware`
This PR builds on `errorHandlingMiddleware` that was implemented in **AT-6724**, now also handling shape validation errors and other various improvements. `errorHandlingMiddleware` was _supposed_ to be implemented in this ticket (**AT-6734**), however I spent some time creating it during **AT-6724** because I didn't want to have to refactor every single unit test twice. 
#### Request shape validation error handling
Refactoring the lambdas to remove:
```
if (setupResult instanceof SetupError) {
    return setupResult.errorResponse;
  } 
```
and adding logic to `errorHandlingMiddleware` to handle it.

Due to recent database discussion, it is probably a waste of time to refactor 
```
if (databaseResult instanceof DynamoDBException) {
    return databaseResult.errorResponse;
```
which only exists in `createPortfolioStep.ts`, will be removing the `createPortfolioStepCommand` and just placing it in a `try/catch` block to keep it consistent with other lambdas.
#### Refactored IF to switch statement
```
const error = serializeError(request.error)!;
    const errorMessage = error?.message as string;
    switch (errorMessage) {
      case "Event object failed validation":
        return new ValidationErrorResponse("Request failed validation", error.details as Record<string, unknown>);
      case "Business rules validation failed":
        return new ValidationErrorResponse(
          "Request failed validation (business rules)",
          error.error_map as Record<string, unknown>
        );
      case "Shape validation failed, invalid UUIDv4":
        return NO_SUCH_PORTFOLIO_DRAFT_404;
    }
```
Using a switch statement instead of multiple ifs will give a marginal performance increase to our functions.

## TODO 

- [x] Clean up commented out code
- [x] Refactor failing tests (these are due to us now wrapping the middy error response with our `ValidationErrorResponse`)
- [x] Merge changes from develop (last task until this ticket is ready for official peer review).
#### Ticket
[AT-6734](https://ccpo.atlassian.net/browse/AT-6734)


packages/api/portfolioDrafts/application/createApplicationStep.ts
```diff
@@ -33,6 +35,7 @@ export async function baseHandler(
   const portfolioDraftId = setupResult.path.portfolioDraftId;
   const applicationStep = event.body;
 
+  validateBusinessRulesForApplicationStep(applicationStep);
```
zachclark-ccpo, line 38: I need to turn this off, standby for the commit.

packages/api/portfolioDrafts/application/createApplicationStep.ts
```diff
@@ -33,7 +35,8 @@ export async function baseHandler(
   const portfolioDraftId = setupResult.path.portfolioDraftId;
   const applicationStep = event.body;
 
-  // TODO(AT-?): add new endpoint for Step 4 that includes this below business rule
+  // validateBusinessRulesForApplicationStep(applicationStep);
+  // TODO(AT-6835): add new endpoint for Step 4 that includes this below business rule
```
juliusfitzhugh-ccpo, line 39: Thanks for updating the TODO along with the func.

packages/api/portfolioDrafts/portfolio/createPortfolioStep.ts
```diff
@@ -38,12 +27,8 @@ export async function baseHandler(
...
   context?: Context
 ): Promise<APIGatewayProxyResult> {
-  // Perform shape validation
-  const setupResult = validateRequestShape<PortfolioStep>(event);
-  if (setupResult instanceof SetupError) {
-    return setupResult.errorResponse;
-  }
-  const portfolioDraftId = setupResult.path.portfolioDraftId;
+  validateRequestShape<PortfolioStep>(event);
+  const portfolioDraftId = event.pathParameters?.portfolioDraftId as string;
```
kylelaker-ccpo, line 31: Would this be better as
```suggestion
  const portfolioDraftId = event.pathParameters!.portfolioDraftId;
```

Or something with a `!`? Instead of the cast to a string?

kylelaker-ccpo: Looks like there are a few places where we alternate. Please be consistent with our behavior elsewhere in the app (whichever that is)

packages/api/utils/errorHandlingMiddleware.ts
```diff
@@ -2,33 +2,31 @@ import middy from "@middy/core";
...
+          "Request failed validation (business rules)",
+          error.error_map as Record<string, unknown>
+        );
+      case "Shape validation failed, invalid UUIDv4":
+        return NO_SUCH_PORTFOLIO_DRAFT_404;
+      case "Shape validation failed, invalid request body":
+        return REQUEST_BODY_INVALID;
+      case "Content type defined as JSON but an invalid JSON was provided":
+        return REQUEST_BODY_INVALID;
     }
```
kylelaker-ccpo, line 27: Good stuff

kylelaker-ccpo (commented):
A few small thoughts on this that I'd like to talk about before actively approving

packages/api/portfolioDrafts/portfolio/createPortfolioStep.ts
zachclark-ccpo: > Looks like there are a few places where we alternate. Please be consistent with our behavior elsewhere in the app (whichever that is)

Don't worry, I went through all the functions that have been refactored with middy, kept them consistent.

> I sort of prefer the `!` because it forces us to add an `eslint-disable-next-line` which also encourages an explanation for _why_ we're disabling that check on that line

I disagree, the warning you would get with `const portfolioDraftId = event.pathParameters!.portfolioDraftId;` is
```
Forbidden non-null assertion.eslint@typescript-eslint/no-non-null-assertion
```
Which exists because: `event.pathParameters.portfolioDraftId;` gives a TS error 
```
Object is possibly 'null'.ts(2531)
```
It cannot be null at all because API Gateway checks if it is a required field, and `validateRequestShape<PortfolioStep>(event);` checks to see if the UUID exists and is a valid UUIDv4, which throws an error.

It was changed to `const portfolioDraftId = event.pathParameters?.portfolioDraftId as string;` because validateRequestShape no longer returns at SetupResult object with a path.

[Optional chaining TS docs](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining)

kylelaker-ccpo: Right. But I still want a comment explaining _why_ it can't be null somewhere. And if we have to do this a lot, perhaps it's an indication we need a have more abstractions around it. But I don't want us to get into the habit of ignoring nullability. The `!` forces a comment anyway. Maybe there's a better way we could expose these values into the functions using Middy so that it can definitely be type safe and so we can safely log an error if for some reason it wasn't present?

kylelaker-ccpo: But I guess the point I am trying to make is that I don't want to mask type errors through casting. I want to accept them and document why it's safe in this scenario.

zachclark-ccpo: We could put that comment in validateRequestShape? I don't want to put it in the handler because it's so _clean_ right now.

zachclark-ccpo: Pushed a commit with comment.

kylelaker-ccpo (approved):
This is looking really good!

PR #433, opened 2021-11-15
Combine two API spec files; Expand burndown operations

jasonburkert-ccpo:
Collapses API defined in `atat_management_and_reporting_api.yaml` into existing API GW resource for API defined in `atat_provisioning_wizard_api.yaml`.

Single burndown operation expanded to:
- `GET /portfolios/{portfolioId}/burndown`
- `GET /portfolios/{portfolioId}/taskorders/{taskOrderNumber}/burndown`
- `GET /portfolios/{portfolioId}/taskorders/{taskOrderNumber}/clins/{clin}/burndown`

Ticket: AT-6769


infrastructure/lib/atat-web-api-stack.ts
```diff
@@ -269,7 +246,45 @@ export class AtatWebApiStack extends cdk.Stack {
...
+      apiDefinition: apigw.ApiDefinition.fromInline(apiSpecAsTemplateInclude2),
+      deployOptions: {
+        tracingEnabled: true,
+      },
+    }).restApi;
     this.restApi = apiGateway;
+    this.restApi2 = apiGateway2;
     this.addEmailRoutes(props);
     this.addTaskOrderRoutes(props);
     this.ssmParams.push(
```
kylelaker-ccpo, line 290: This should work just fine but to be clear, if this is really the route we want to go (two `SpecRestApi` resources) then we need to talk about how it impacts:
 - the UI (and its config endpoint)
 - CORS
 - the Systems Manager parameters we create

I'd _really_ prefer if we found a way to create a single file for this unless we truly expect to develop it as 2+ APIs over time. 

In any case, we should go ahead and plan for a [zero one infinity rule](https://en.wikipedia.org/wiki/Zero_one_infinity_rule) sort of thing here and support N API GW resources (and therefore use arrays and for loops).

kylelaker-ccpo (changes requested):

infrastructure/lib/atat-web-api-stack.ts
jasonburkert-ccpo: Looks like we're collapsing the two API specs into a single API spec, so only a single `SpecRestApi` will be required.  Thank you for the thoughtful input.

atat_provisioning_wizard_api.yaml
```diff
@@ -797,7 +797,189 @@ paths:
...
+          required: true
+          schema:
+            type: string
+        - name: taskOrderNumber
+          in: path
+          description: Task Order Number (not assigned by ATAT)
+          required: true
+          schema:
+            type: string
+        - name: start_date
```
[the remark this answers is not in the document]
jeffsegal-ccpo: Thanks, fixed it and renamed it to just `clin`.

jeffsegal-ccpo (approved):

juliusfitzhugh-ccpo (approved):
Reviewed and nothing stood out to me. The second spec for management and reporting operations is being combined into the original `atat_provisioning_wizard_api.yaml` spec, so the need for additional CDK infrastructure is no longer needed.

PR #435, opened 2021-11-16
Add IP logging w/ middy

matthewdvertola-ccpo:
Logging source IP addresses of incoming traffic enables auditing.

I had to npm install http-errors and looks like it changed around package-lock. Not sure if thats okay, but assuming that it probably is not.

Ticket: AT-6634


kylelaker-ccpo:
Yeah, we probably don't want to have that installed in the root package. Can you maybe try reverting those changes and then doing a `lerna bootstrap --ci` to see if it installs the package correctly for you?

packages/api/portfolioDrafts/application/createApplicationStep.ts
```diff
@@ -95,4 +96,5 @@ export const handler = middy(baseHandler)
   .use(jsonBodyParser())
   .use(validator({ inputSchema: wrappedApplicationStepSchema }))
   .use(JSONErrorHandlerMiddleware())
-  .use(cors(CORS_CONFIGURATION));
+  .use(cors(CORS_CONFIGURATION))
+  .use(IpCheckerMiddleware());
```
kylelaker-ccpo, line 100: We probably want this to always be the _frist_ middleware that gets executed so that the logging happens almost immediately at the start of the request. Thoughts @juliusfitzhugh-ccpo and @zachclark-ccpo?

zachclark-ccpo: I agree, it should be the first middleware called.

juliusfitzhugh-ccpo: Yeah, if we want to log before anything happens it would be best to be the first middleware.

packages/email/processingEmails/submitEmails.ts
```diff
@@ -62,4 +63,5 @@ export const handler = middy(baseHandler)
   .use(jsonBodyParser())
   .use(validator({ inputSchema: schemaWrapper }))
   .use(JSONErrorHandlerMiddleware())
-  .use(cors({ headers: "*", methods: "*" }));
+  .use(cors({ headers: "*", methods: "*" }))
+  .use(IpCheckerMiddleware());
```
juliusfitzhugh-ccpo, line 67: I think this one was missed by accident since you made the `IpCheckerMiddleware` first on the other ones.

packages/api/portfolioDrafts/createPortfolioDraft.ts
```diff
@@ -44,3 +46,8 @@ export async function handler(event: APIGatewayProxyEvent): Promise<APIGatewayPr
     return DATABASE_ERROR;
   }
 }
+
+// IP logging middy
+const middyHandler = middy(handler);
+middyHandler.use(IpCheckerMiddleware());
+export { middyHandler };
```
juliusfitzhugh-ccpo, line 53: I would suggestion staying consistent with the other files and change `handler` to  `baseHandler` and use `handler` for the `baseHandler` wrapped with middy (example below). From my understanding `handler` is the default method invoked by lambdas ([AWS Lambda Functions - handler](https://docs.aws.amazon.com/lambda/latest/dg/nodejs-handler.html)). There were other ones that were similar to this, so you would want to update them too.  
```ts
export const handler = middy(baseHandler)
    .use()
```

packages/api/utils/ipLogging.test.ts
```diff
@@ -0,0 +1,9 @@
+import { IpCheckerMiddleware } from "./ipLogging";
+
+describe("Test console logging", function () {
+  it("should console log the event", async () => {
+    const consoleLogSpy = jest.spyOn(console, "log");
+    IpCheckerMiddleware();
+    expect(consoleLogSpy).toHaveBeenCalled();
```
juliusfitzhugh-ccpo, line 7: I did come across this test failing as well with `consoleLogSpy` not being called. I think this is because of how the middleware is being used with middy. A quick sample is below. You can also look at `xssSanitizer.test.ts` as an alternative to testing the middleware.

```ts
    const request: any = {
      ...validRequest,
      body: mockPortfolioStep,
      requestContext: { identity: { sourceIp: "10.2.2.2" } },
    };
    const baseHandler = (event: any) => {
      return event;
    };
    const handler = middy(baseHandler).use(IpCheckerMiddleware());
    await handler(request, {} as Context, () => null);
```

juliusfitzhugh-ccpo (approved):
Modified some unit tests to help with coverage and reverted back to using `handler` in tests to make sure the middleware is being tested. Otherwise looks good and logging the sourceIP is working nicely!

Tested the below endpoints in a deploy stack using Postman and the source IP address was being logged for each endpoint. Not noticeable errors.
- `createApplicationStep`
- `createFundingStep`
- `createPortfolioStep`
- `createPortfolioStep`
- `getApplicationStep`
- `getFundingStep`
- `getPortfolioStep`
- `getPortfolioDraft`
- `getPortfolioDrafts`
- `submitPortfolioDraft`

packages/api/utils/ipLogging.ts
```diff
@@ -0,0 +1,14 @@
+import middy from "@middy/core";
+import { APIGatewayProxyEvent, APIGatewayProxyResult } from "aws-lambda";
+
+// Create the middleware and export it
+export const IpCheckerMiddleware = (): middy.MiddlewareObj<APIGatewayProxyEvent, APIGatewayProxyResult> => {
+  // Set up a before check, this will run before the handler
+  const before: middy.MiddlewareFn<APIGatewayProxyEvent, APIGatewayProxyResult> = async (request): Promise<void> => {
+    // log the sourceIp
+    console.log(request.event.requestContext.identity.sourceIp);
```
jeffsegal-ccpo, line 9: Can you add context to the line here so that it doesn't just come across as `1.2.3.4` but also tells us what is being logged?

jasonburkert-ccpo, line 9: In addition to added context per Jeff, I suggest using the `info`-level to log.

This would necessitate a corresponding change to any unit test with a spy method for _log_.
`const consoleLogSpy = jest.spyOn(console, "log");`

```suggestion
    console.info("Source IP: ", request.event.requestContext.identity.sourceIp);
```

packages/api/portfolioDrafts/submit/submitPortfolioDraft.test.ts
```diff
@@ -18,13 +18,16 @@ beforeEach(() => {
 
 describe("Validation tests", () => {
   it("should require path param", async () => {
-    const emptyRequest: APIGatewayProxyEvent = {} as any;
-    expect(await handler(emptyRequest)).toEqual(PATH_PARAMETER_REQUIRED_BUT_MISSING);
+    const emptyRequest: APIGatewayProxyEvent = {
+      // ...validRequest,
```
jasonburkert-ccpo, line 22:
```suggestion
```

packages/api/utils/ipLogging.test.ts
```diff
@@ -0,0 +1,22 @@
...
+  it("should console log the event", async () => {
+    const request: any = {
+      validRequest,
+      body: mockPortfolioStep,
+      requestContext: { identity: { sourceIp: "10.2.2.2" } },
+    };
+    const baseHandler = (event: any) => {
+      return event;
+    };
+    const consoleLogSpy = jest.spyOn(console, "log");
```
jasonburkert-ccpo, line 17: If log level is changed to `info`, spy must be updated, here and in similar places in other unit tests
```suggestion
    const consoleLogSpy = jest.spyOn(console, "info");
```

jasonburkert-ccpo (changes requested):
In addition to the suggestion about logging with `info`-level,
can we turn the test value "10.2.2.2" into a constant, put it someplace like `commonPortfolioDraftMockData.ts`, and reference that?

packages/api/utils/ipLogging.ts
jasonburkert-ccpo: Suggest something like
`console.log("Source IP:", request.event.requestContext.identity.sourceIp);`

jasonburkert-ccpo:
Per team discussion, `console.log()` is good since it will log at the desired `INFO` level.  Ignore any previous requests to change to _console.info_, or change spy method to _info_.

jasonburkert-ccpo (approved):

jasonburkert-ccpo:
Approving with the assumption that context is added per @jeffsegal-ccpo

packages/api/utils/ipLogging.ts
kylelaker-ccpo:
```suggestion
    console.log("Client IP address: " + request.event.requestContext.identity.sourceIp);
```

packages/api/utils/ipLogging.ts
```diff
@@ -0,0 +1,14 @@
+import middy from "@middy/core";
+import { APIGatewayProxyEvent, APIGatewayProxyResult } from "aws-lambda";
+
+// Create the middleware and export it
+export const IpCheckerMiddleware = (): middy.MiddlewareObj<APIGatewayProxyEvent, APIGatewayProxyResult> => {
+  // Set up a before check, this will run before the handler
+  const before: middy.MiddlewareFn<APIGatewayProxyEvent, APIGatewayProxyResult> = async (request): Promise<void> => {
+    // log the sourceIp
+    console.log("Client IP address: " + request.event.requestContext.identity.sourceIp);
```
jasonburkert-ccpo, line 9: 👍

PR #439, opened 2021-11-16
Add GitHub Actions OIDC integration

kylelaker-ccpo:
This performs a base integration that just does an
`sts:GetCallerIdentity`.

Ticket: AT-6847


kylelaker-ccpo:
For a sample successful run with the role, see https://github.com/dod-ccpo/atat-web-api/runs/4231461938?check_suite_focus=true (or look at the runs on the _first_ commit on this branch).

jeffsegal-ccpo (approved):

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,28 @@
...
+jobs:
+  deployInfrastructure:
+    name: Test STS
+    runs-on: ubuntu-latest
+    permissions:
+      # Required for the OIDC endpoint
+      id-token: write
+    steps:
+      # - name: Checkout code
+      #   uses: actions/checkout@v2
```
juliusfitzhugh-ccpo, line 18: I know we are testing this first with just `aws sts get-caller-identity`, and I assume we will use `actions/checkout@v2` when we want to actually deploy, but I wanted to make sure this was intentional.

juliusfitzhugh-ccpo (approved):
One comment for clarity, but other than that everything looked good to me.

.github/workflows/deploy.yml
kylelaker-ccpo: Yep! There's no need to check out the code (as-is) because we're not using it. I sort of wanted that present as a way to almost explicitly show that we're not doing that as part of the test today.

PR #434, opened 2021-11-15
Add an RDS database

kylelaker-ccpo:
Adds an Amazon RDS resource with a PostgreSQL database instance type

Ticket: AT-6834

PR #448, opened 2021-11-18
Configure TypeORM; Define Portfolio data model

jasonburkert-ccpo:
- adds packages `typeorm`, `reflect-metadata`, `pg`
- entities located in `packages/api/orm/entity`
- adds entities
  - `Portfolio`
  - `Application`
  - `Environment`
  - `TaskOrder`
  - `Clin`
- adds abstract entities
  - `BaseEntity`
  - `ProvisionableEntity`
- adds relations
  - `Portfolio-Application`
  - `Application-Environment`
  - `Portfolio-TaskOrder`
  - `TaskOrder-Clin`
- each of `Portfolio`, `Application`, `Environment` have same groups of operators defined in latest API inherited from `ProvisionableEntity`

Ticket: AT-6836


packages/api/ormconfig.json
```diff
@@ -0,0 +1,16 @@
+{
```
kylelaker-ccpo, line 1: I think we'll probably need to avoid committing this file and providing the configuration inline during our connection setup code since we're going to dynamically pull the host, port, username, database, and password.

packages/api/orm/index.ts
```diff
@@ -0,0 +1,16 @@
...
+  .then(async (connection) => {
+    const pd = new Portfolio();
+    pd.status = ProvisioningStatus.IN_PROGRESS;
+    await connection.manager.save(pd);
+    console.log("Saved a new portfolio with id: " + pd.id);
+
+    const pfs = await connection.manager.find(Portfolio);
+    console.log("Loaded portfolios: ", pfs);
+  })
+  .catch((error) => console.log(error));
```
kylelaker-ccpo, line 16: As I've been working in my #434 PR I've been using `await` rather than `.then()`/`.catch()` to match some of our existing style across the app. Do you have a preference going forward?

kylelaker-ccpo (commented):
I know you're early in the process so the things I'm commenting on are likely not final but I do want to mention them to make sure we're sort of going in the same direction

packages/api/ormconfig.json
jasonburkert-ccpo: Sure, using this to get the models together.  Will remove if not needed.

packages/api/orm/index.ts
jasonburkert-ccpo: Await is fine.  The TypeORM guides use this `then()` / `catch()` syntax.

packages/api/ormconfig.json
jasonburkert-ccpo: Removed in 5852a8bfc32a595be3f925a56a516c859e56e37a

jasonburkert-ccpo:
Sample output for inserts and retrievals
[typeorm_sample_out.txt](https://github.com/dod-ccpo/atat-web-api/files/7591174/typeorm_sample_out.txt)

packages/api/orm/entity/Clin.ts
```diff
@@ -0,0 +1,51 @@
...
+import { Column, Entity, ManyToOne } from "typeorm";
+import { TaskOrder } from "./TaskOrder";
+
+@Entity()
+export class Clin extends BaseEntity {
+  @Column({
+    type: String,
+    nullable: false,
+    length: 4,
+    comment: "contract line item number from task order, 0001 through 9999",
```
kylelaker-ccpo, line 11: Will updating these comments later require a database schema update? Do we want to commit to that potential maintenance burden for nearly every column if so?

packages/api/orm/entity/Clin.ts
```diff
@@ -0,0 +1,51 @@
...
+  })
+  clinNumber: string;
```
kylelaker-ccpo, line 13: Postgres actually has functions to leftpad a number with 0s. Is it better to store these as unsigned integers or as strings?

packages/api/orm/entity/Environment.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  // operators - simple-array or AppEnvOperator[] ?
+  @Column({ type: "simple-array", default: "" })
+  administrators: string[];
+
+  @Column({ type: "simple-array", default: "" })
+  contributors: string[];
+
+  @Column({ type: "simple-array", default: "" })
+  readOnlyOperators: string[];
```
kylelaker-ccpo, line 21: Can these move to the `ProvisionableEntity` class? It seems like `Portfolio`, `Environment`, and `Application` all have them. Am I missing another specializing class?

packages/api/orm/entity/Clin.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  @Column({
+    type: "decimal",
+    nullable: false,
+    precision: 17,
+    scale: 2,
+    default: 0.0,
+    comment: "amount already disbursed by USG for project or contract to be spent during CLIN POP",
+  })
+  obligatedFunds: number;
```
kylelaker-ccpo, line 41: Postgres does have a `money` type that TypeORM supports. Is there a benefit to decimal vs money? https://www.postgresql.org/docs/13/datatype-money.html

packages/api/orm/entity/Environment.ts
```diff
@@ -0,0 +1,22 @@
...
+@Entity()
+export class Environment extends ProvisionableEntity {
+  @Column({ type: String, nullable: false, length: 100 })
+  name: string;
+
+  @ManyToOne(() => Application, (application) => application.environments)
+  application: Application;
+
+  // operators - simple-array or AppEnvOperator[] ?
+  @Column({ type: "simple-array", default: "" })
```
kylelaker-ccpo, line 14: The docs make me a little worried about this type because of the limitation that values must not contain commas. Is that a constraint we're sure we're okay with?

packages/api/orm/entity/Portfolio.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  @Column({ type: "enum", enum: CloudServiceProvider, nullable: false })
+  csp: CloudServiceProvider;
+
+  @Column({ type: String, nullable: false, array: true })
+  dodComponents: Array<string>;
+
+  @Column({ type: String, nullable: false, array: true })
+  portfolioManagers: Array<string>;
+
```
kylelaker-ccpo, line 26: Any reason to prefer `Array<string>` here over `string[]` that gets used later?

kylelaker-ccpo (changes requested):
I don't really have many specific changes that I am requesting but I do have a few questions about the tradeoffs of some of this design.

packages/api/orm/entity/TaskOrder.ts
```diff
@@ -0,0 +1,34 @@
...
+@Entity()
+export class TaskOrder extends BaseEntity {
+  @Column({
+    type: String,
+    nullable: false,
+    length: 17,
+    comment: "TO numbers are 13 characters. TO modifications are 17 characters.",
+  })
+  taskOrderNumber: string;
+
```
zachclark-ccpo, line 16: See Jeff's PR, there is a comment about the business rules for Task Order.
Link: https://github.com/dod-ccpo/atat-web-api/pull/467#discussion_r755612885

packages/api/orm/entity/Environment.ts
jasonburkert-ccpo: Yes, I think that is a good improvement.  Those three entities are the totality.

packages/api/orm/entity/TaskOrder.ts
```diff
@@ -0,0 +1,34 @@
...
+  @Column({ type: "enum", nullable: false, enum: FileScanStatus, default: FileScanStatus.PENDING })
+  fileScanStatus: FileScanStatus;
+
+  @ManyToOne(() => Portfolio, (portfolio) => portfolio.taskOrders)
+  portfolio: Portfolio;
+
+  @OneToMany(() => Clin, (clin) => clin.taskOrder)
+  clins: Clin[];
+}
```
zachclark-ccpo, line 34: Going through the examples here:
https://github.com/typeorm/typescript-example/blob/master/src/entity/Post.ts
I noticed that they used: 
```
@JoinTable()
```
Is that something we need?

packages/api/orm/entity/Clin.ts
```diff
@@ -0,0 +1,51 @@
+import { BaseEntity } from "./BaseEntity";
+import { Column, Entity, ManyToOne } from "typeorm";
+import { TaskOrder } from "./TaskOrder";
+
+@Entity()
+export class Clin extends BaseEntity {
+  @Column({
+    type: String,
```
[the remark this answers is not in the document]
jasonburkert-ccpo: A good question.  I don't have a good answer, and have wondered the same.  I think the source may have been some example.  It's probably better to remove the type specification from the column options and rely on the conversion of the TypeScript type to the compatible postgres type in these cases.  I'll look more deeply at this and either explain or replace.

packages/api/orm/entity/Clin.ts
```diff
@@ -0,0 +1,51 @@
...
+    nullable: false,
```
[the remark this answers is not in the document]
jasonburkert-ccpo: I'll remove any `nullable: false` from the column options in preference of relying on the default.

tsconfig.json
```diff
@@ -25,6 +25,7 @@
     "strictPropertyInitialization": false,
     "resolveJsonModule": true,
     "useUnknownInCatchVariables": false,
+    "emitDecoratorMetadata": true,
```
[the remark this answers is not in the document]
jasonburkert-ccpo: It's actually part of the documented config on typeorm.io

packages/api/package.json
```diff
@@ -5,15 +5,17 @@
   "scripts": {
     "test": "jest --config=jest.config.js",
     "lint": "eslint --ext .js,.ts,.json -c ../../.eslintrc.json .",
-    "lint:fix": "npm run lint -- --fix"
+    "lint:fix": "npm run lint -- --fix",
+    "start": "ts-node orm/index.ts",
+    "typeorm": "node --require ts-node/register ./node_modules/typeorm/cli.js"
```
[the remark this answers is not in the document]
jasonburkert-ccpo: The `start` command here can be removed.  I simply missed this during my cleanup.
And while the `typeorm` script was useful to get familiar with the tool, I didn't use it a great deal, and it can probably be removed also.

packages/api/orm/entity/Clin.ts
jasonburkert-ccpo: No, I don't believe it would require a schema update.  The table and column comments can be updated discretely with the [comment command](https://www.postgresql.org/docs/13/sql-comment.html).  They can be viewed in Postgres within `psql` using the `\d+ <tablename>` command.

I think it's worth potential maintenance burden for fields containing important business information that may not be clear from field names.  I purposefully didn't place them everywhere, because some things are self evident.  I imagine that if I was maintaining (or querying) this system far in the future that it might be helpful to have descriptions for fields immediately next to the data, at least for entities such as `clin`.

packages/api/orm/entity/Clin.ts
jasonburkert-ccpo: Hmm.  We'd been using decimal/number in the API spec and TypeScript to this point.  I hadn't considered the money type.

packages/api/orm/entity/Portfolio.ts
jasonburkert-ccpo: Nope.  Will make consistent.

packages/api/orm/entity/TaskOrder.ts
jasonburkert-ccpo: All of the relations that I created have the parent with a `@OneToMany` relationship with a child, and then in the child specifying the reverse `@ManyToOne` relationship.  This results in the child effectively having a foreign key column pointing to the parent.  So two entities, with the child pointing back to the parent.

A `@JoinTable` is not needed in these cases, as the example you've linked illustrates a `@ManyToMany` relationship.  Imagine that between the parent and child tables exists a third table, which is necessary to model that relationship in SQL.

> @JoinTable() is required for @ManyToMany relations. You must put @JoinTable on one (owning) side of relation.

jasonburkert-ccpo: @zachclark-ccpo Good question.

packages/api/orm/entity/Clin.ts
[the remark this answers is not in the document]
juliusfitzhugh-ccpo: I know i'm late and it seems like the `type` will be removed, but in case we want to keep it `varchar` might be the `type` that best fits according to the [Postgres docs](https://www.postgresql.org/docs/9.3/datatype-character.html).

packages/api/package.json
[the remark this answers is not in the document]
jasonburkert-ccpo: Removed in 7dd84c15da6cd3fd23c1a3aa51879a92a5d3b96d

packages/api/orm/entity/Portfolio.ts
```diff
@@ -0,0 +1,42 @@
...
+  @OneToMany(() => Application, (application) => application.portfolio)
+  applications: Application[];
+
+  // operators - simple-array or PortfolioOperator[] ?
+  @Column({ type: "simple-array", default: "" })
+  administrators: string[];
+
+  @Column({ type: "simple-array", default: "" })
+  contributors: string[];
```
jeffsegal-ccpo, line 38: Technically we don't support access levels other than administrator for Portfolios at the API level, but if we want to go ahead and re-use these three fields via inheritance that's probably fine since the API wouldn't allow using the other two access levels anyway.

packages/api/orm/entity/Clin.ts
jeffsegal-ccpo: Yeah as long as we can easily convert to and from money & Number between the DB and API then that would be a nice option.

packages/api/orm/entity/Clin.ts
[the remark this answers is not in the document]
jasonburkert-ccpo: Removed in 5d71ae47d8f42cb180d7bda879989da916e39fdf

packages/api/orm/entity/Clin.ts
jasonburkert-ccpo: Tried it.  Went smoothly.  Didn't even need to adjust my insert.

[This reference](https://wiki.postgresql.org/wiki/Don%27t_Do_This#Don.27t_use_money) includes the following, to which we generally conform.  This should be ok for our use case.
> When should you?
If you're only working in a single currency, aren't dealing with fractional cents and are only doing addition and subtraction then money might be the right thing.

jasonburkert-ccpo: Changed in a0c0142b756741ad4379072e0f6efe80f57efa25

packages/api/orm/entity/Portfolio.ts
jasonburkert-ccpo: Our other models tend to use the `Array<type>` syntax, so will use that.

jasonburkert-ccpo: Updated in d0cb229be7a30f5bfe8a4e027345dcd518cff62a

packages/api/orm/entity/TaskOrder.ts
jasonburkert-ccpo: Thanks for the note.

packages/api/orm/entity/Portfolio.ts
jasonburkert-ccpo: Good point.  Will make a note in the inherited class.

packages/api/orm/entity/Clin.ts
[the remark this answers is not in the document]
jasonburkert-ccpo: For simple TypeScript `string` fields, I removed the type from the column options.  For two that are arrays of strings, I changed the type in the column options to `varchar`.

Updated in f14ec95b8c703c52846010133b574130fa777f79

packages/api/orm/entity/Environment.ts
jasonburkert-ccpo: Moved in 54e87a13ffaeead8556cbb1c46f680dc2bcb2bd6

packages/api/orm/entity/Environment.ts
jasonburkert-ccpo: I've replaced these with arrays of varchars in 6d56cf589554dff6723f2478a0caca08682a7c68
This makes them similar to `dodComponents` and `portfolioManagers` from `Portfolio`, but with a default value.

packages/api/orm/entity/Clin.ts
jasonburkert-ccpo: I'm not sure.  Storage wise perhaps better to store the integers, but these numbers are always presented this way in this business domain.  We've got existing validation to ensure incoming data is 4 chars.

packages/api/orm/entity/Portfolio.ts
```diff
@@ -0,0 +1,32 @@
...
+@Entity()
+export class Portfolio extends ProvisionableEntity {
+  @Column({ type: String, length: 100 })
+  name: string;
+
+  @Column({ type: String, nullable: true, length: 300 })
+  description: string;
+
+  @Column({ type: String })
+  owner: string;
```
juliusfitzhugh-ccpo, line 16: I noticed you removed the type in the `column` for string, should these be removed as well?

packages/api/package.json
```diff
@@ -26,8 +26,11 @@
     "http-errors": "^1.8.0",
     "lambda-multipart-parser": "^1.0.1",
     "middy-middleware-json-error-handler": "^3.0.0",
-    "serialize-error": "^8.1.0",
```
juliusfitzhugh-ccpo, line 29: Was this deleted intentionally?

packages/api/orm/entity/Portfolio.ts
jasonburkert-ccpo: Yes.  Nice catch.

packages/api/package.json
jasonburkert-ccpo: No, looks like this is due to reordering.  This same package is listed on line 32.

packages/api/orm/entity/Portfolio.ts
jasonburkert-ccpo: Updated in 06b254a386c831c08e7d400788fa2bf6d17fc6a7

packages/api/package.json
juliusfitzhugh-ccpo: Sorry my mistake, I must have over looked it.

juliusfitzhugh-ccpo (approved):
Nice job, we are setup nicely for the next steps!

PR #464, opened 2021-11-23
Define a model for abstracting shared behaviors across lambdas

zachclark-ccpo:
## Overview
This PR is dedicated to fixing our custom cdk constructs, allowing for more flexibility when composing functions by moving away from the rigid traditional inheritance model (e.g. Animal -> Feline -> Cat).

### One construct to rule them all
All of the existing `props` of the various cdk constructs have been gathered together into a single file.
They are declared as optional `props` (for example, s3 buckets):
```
/**
   * Optional - Secure bucket for S3 functions
   */
```
Inside the actual constructor, there is an if statement to check if the prop is defined. If the prop is defined, then the resource functionality is added:
```
// Optional - s3 bucket
    if (props.bucket) {
      this.bucket = props.bucket.bucket;
      this.fn.addEnvironment("DATA_BUCKET", this.bucket.bucketName);
      switch (props.bucketPermissions) {
        case "READ":
          this.bucket.grantRead(this.fn);
          break;
        case "PUT":
          this.bucket.grantPut(this.fn);
          break;
        case "DELETE":
          this.bucket.grantDelete(this.fn);
          break;
        default:
          this.bucket.grantRead(this.fn);
      }
    }
```
These if statements replace the existing inheritance model, eliminating the need for `api-s3-function.ts` because all of its features are now included in an if statement.

### Flexible Functions  (an Animal that can `purr` and `bark` at the same time)
The composition of the construct(ApiFunction) is based on the user defined `props`. 
Example for the following use-cases:
Function which can access the data store but is not exposed by API Gateway:
```
 example = new ApiFunction(this, "CreateExample", {
    table: this.table,
    handlerPath: this.determineApiHandlerPath("createExample", "examples/"),
  });
```
Function which can access an S3 Bucket and the data store:
```
 example = new ApiFunction(this, "CreateExample", {
    table: this.table,
    bucket: taskOrderManagement.pendingBucket,
    handlerPath: this.determineApiHandlerPath("createExample", "examples/"),
  });
```
Function which can access Secrets Manager but does not need the data store:
```
 example = new ApiFunction(this, "CreateExample", {
    smtpSecrets: smtpSecrets,
    handlerPath: this.determineApiHandlerPath("createExample", "examples/"),
  });
```
### Notes

- Added `tablePermissions`, before these permissions were determined by what `HTTPMethod` the function was using. Using `HTTPMethod` is a problem for Functions that are not connected to the internet via API Gateway.
- It might be worth it to make the if statements into private fns that still live in the file, that way it would be less of mess for future implementations.

## TODO

- [x] Replace existing (0-1)functions in `atat-web-api-stack.ts` with the new model.
- [x] Move tables/s3/sqs/etc if statement logic into private functions???These would still live in the same file, but it would appear cleaner.

#### Resources
I got quite a bit of inspiration from the [AWS Construct Library Design Guidelines](https://github.com/aws/aws-cdk/blob/master/docs/DESIGN_GUIDELINES.md#vpc)

**TICKET:**[AT-6845](https://ccpo.atlassian.net/browse/AT-6845)


kylelaker-ccpo:
Overall, I like this. Some ideas:
 - what if these took arrays instead of single items? or even `T |  T[]`?
 - it feels like it may get awkward to "jam" all the resources into one `new` call. is a construct right for this? could we add a generic `grant()` method to add more stuff later?
 - can we at least change _one_ example to use this new construct to determine it's viability? if it trivial can we change all/many of them?

zachclark-ccpo:
> Overall, I like this. Some ideas:
> 
> * what if these took arrays instead of single items? or even `T |  T[]`?

That is a good point, what if someone needs to define multiple SQS queues.
> * it feels like it may get awkward to "jam" all the resources into one `new` call. is a construct right for this? could we add a generic `grant()` method to add more stuff later?

Do you mean something like:
```
private grantRequiredQueuePermissions(): iam.Grant | undefined {
    switch (this.method) {
      // Read-Only operations
      case HttpMethod.GET:
      case HttpMethod.HEAD:
        return this.queue.grantConsumeMessages(this.fn);
      // Read-Write operations
      case HttpMethod.POST:
      case HttpMethod.PUT:
      case HttpMethod.DELETE:
        return this.queue.grantSendMessages(this.fn);
      default:
        // This will allow Synthesis to continue; however, the error will stop the
        // CDK from moving forward with a diff or deploy action.
        cdk.Annotations.of(this).addError("Unknown HTTP method " + this.method);
        return undefined;
    }
  }
}
```
but with the various resources? I was playing around with that today, have a TODO to address that.

> * can we at least change _one_ example to use this new construct to determine it's viability? if it trivial can we change all/many of them?

Yes that won't be a problem at all, I have a TODO to actually change one of the fns and its also an AC on the ticket

zachclark-ccpo:
If you are reviewing this:

> Give me a better name for `ApiFunctiontest` and `lambda-fn.ts`.
> Note that `addDatabaseApiFunction` and `addQueueDatabaseApiFunction` on the `atat-web-api-stack.ts` **are** using the new function construct.

infrastructure/lib/constructs/lambda-fn.ts
```diff
@@ -0,0 +1,220 @@
...
+   * @param table - DynamoDB table connection
+   * @param bucket - S3 bucket connection
+   * @param queue - SQS queue connection
+   * @param stateMachine - State Machine connection
+   */
+  public readonly method: HttpMethod;
+  public readonly table: dynamodb.ITable;
+  public readonly bucket: s3.IBucket;
+  public readonly queue: sqs.IQueue;
+  public readonly stateMachine?: sfn.IStateMachine;
```
juliusfitzhugh-ccpo, line 127: Do we needed the `?` for `stateMachine` only? Just wondering.

juliusfitzhugh-ccpo (approved):
I'm liking this. One small question, but not blocking. 

Not sure I can help with the names, but I how about `baseFunction`, `genericFunction`, `generalFunction` or `generalLambda`?? And for the `lambda-fn.ts` I think that kind of fits or something like `services-fn.ts`. Feel free to disregard, others might have better suggestions.

infrastructure/lib/constructs/lambda-fn.ts
zclarkEDC: Good catch, forgot to remove the `?`.

jasonburkert-ccpo: Was removed in 1145be5180547470cc60ea395e8fe493755a7acc

infrastructure/lib/constructs/lambda-fn.ts
```diff
@@ -0,0 +1,219 @@
...
+  readonly batchSize?: number;
+
+  /**
+   * The State Machine that will start execution when this function
+   * is called.
+   */
+  readonly stateMachine?: sfn.IStateMachine;
+}
+
+export class ApiFunctiontest extends cdk.Construct {
```
jasonburkert-ccpo, line 108: OptionalResourcesApiFunction, ExtensibleFunction, LegoFunction, SwissArmyFunction?  /s

jasonburkert-ccpo (approved):

infrastructure/lib/constructs/lambda-fn.ts
```diff
@@ -0,0 +1,219 @@
...
+      case "PUT":
+        return this.bucket.grantPut(this.fn);
+      case "DELETE":
+        return this.bucket.grantDelete(this.fn);
+      default:
+        return this.bucket.grantRead(this.fn);
+    }
+  }
+
+  private grantRequiredSQSPermissions(queuePermissions: string | undefined): iam.Grant | undefined {
```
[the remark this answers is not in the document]
zachclark-ccpo: Good call, I will shorten them

infrastructure/lib/atat-web-api-stack.ts
```diff
@@ -125,27 +127,48 @@ export class AtatWebApiStack extends cdk.Stack {
...
+    );
 
     // Submission operations (all files live in the submit/ folder)
-    this.addQueueDatabaseApiFunction("submitPortfolioDraft", "portfolioDrafts/submit/", props.vpc);
+    this.addQueueDatabaseApiFunction(
+      "submitPortfolioDraft",
+      "portfolioDrafts/submit/",
+      props.vpc,
+      TablePermissions.READ_WRITE,
+      QueuePermissions.SEND
```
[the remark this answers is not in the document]
zachclark-ccpo: Enums are awesome

infrastructure/lib/constructs/lambda-fn.ts
zachclark-ccpo: Going to go with ApiFlexFunction

zachclark-ccpo: But Lego function was the runner up

infrastructure/lib/atat-web-api-stack.ts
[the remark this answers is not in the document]
kylelaker-ccpo: I really wish that as part of this we were able to get rid of these additional wonky functions. If we're going to define a helper wrapper, why isn't it helpful enough? I thought the whole point was to remove all of these bespoke QueueDatabaseApi vs DatabaseApi things?

infrastructure/lib/constructs/lambda-fn.ts
```diff
@@ -0,0 +1,219 @@
...
+import { SqsEventSource } from "@aws-cdk/aws-lambda-event-sources";
+import * as sfn from "@aws-cdk/aws-stepfunctions";
+
+/**
+ * An IAM service principal for the API Gateway service, used to grant Lambda
+ * invocation permissions.
+ */
+const APIGW_SERVICE_PRINCIPAL = new iam.ServicePrincipal("apigateway.amazonaws.com");
+
+export interface ApiFunctionPropstest {
```
kylelaker-ccpo, line 21: Why is `test` at the end of this name? Why do the `Props` not match the name of the class as we do everywhere else?

infrastructure/lib/constructs/lambda-fn.ts
```diff
@@ -0,0 +1,219 @@
...
+   */
+  readonly table?: dynamodb.ITable;
+
+  /**
+   * Optional table permissions to grant to the Lambda function.
+   * One of the following may be specified: "ALL", "READ", "READ_WRITE", "WRITE".
+   *
+   * @default - Read/write access is given to the Lambda function if no value is specified.
+   */
+  readonly tablePermissions?: string;
```
kylelaker-ccpo, line 62: Just accept the enum `TablePermissions` a type so you don't have to define the specific allowed values.

infrastructure/lib/constructs/lambda-fn.ts
```diff
@@ -0,0 +1,219 @@
...
+   */
+  readonly bucket?: SecureBucket;
+
+  /**
+   * Optional bucket permissions to grant to the Lambda function.
+   * One of the following may be specified: "READ", "PUT", "DELETE".
+   *
+   * @default - Read access is given to the Lambda function if no value is specified.
+   */
+  readonly bucketPermissions?: string;
```
kylelaker-ccpo, line 75: Same here. This really needs to not be a string

infrastructure/lib/constructs/lambda-fn.ts
```diff
@@ -0,0 +1,219 @@
...
+   */
+  readonly queue?: sqs.IQueue;
+
+  /**
+   * Optional bucket permissions to grant to the Lambda function.
+   * One of the following may be specified: "SEND", "CONSUME".
+   *
+   * @default - Read access is given to the Lambda function if no value is specified.
+   */
+  readonly queuePermissions?: string;
```
kylelaker-ccpo, line 88: Again. Strings should not hold enumerated values.

infrastructure/lib/queue-permissions.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Supported Queue Permissions for the application.
+ */
+export enum QueuePermissions {
+  SEND = "SEND",
+  CONSUME = "CONSUME",
+}
```
kylelaker-ccpo, line 7: Why not just make a file `infrastructure/lib/permissions.ts`? Why do queue and tables need their own files? Can we define a more general `CREATE`, `READ`, `UPDATE`, `DELETE` set of values that can be used across resources?

infrastructure/lib/table-permissions.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Supported Table Permissions for the application.
+ */
+export enum TablePermissions {
+  ALL = "ALL",
+  READ = "READ",
+  READ_WRITE = "READ_WRITE",
+  WRITE = "WRITE",
+}
```
kylelaker-ccpo, line 9: if you notice yourself creating an enum where some members are combinations of other members (like `READ_WRITE` or `ALL`) then you almost certainly would be better served by integers that you can `|`. So:

```ts
class TablePermissions {
  public const READ = 1 >> 0;
  public const WRITE = 1 >> 1;
  public const FOO = 1 >> 2;

  const get ALL() {
    return READ | WRITE | FOO;
  }
}


if (resource.permissions & TablePermissions.WRITE) {
  // resource.grantWriteAccess();
}
if (resource.permissions & TablePermissions.READ) {
  resource.grantReadAccess();
}
```

infrastructure/lib/constructs/lambda-fn.ts
```diff
@@ -0,0 +1,219 @@
...
+  }
+
+  private grantSQSPermissions(queuePermissions: string | undefined): iam.Grant | undefined {
+    switch (queuePermissions) {
+      case "SEND":
+        return this.queue.grantSendMessages(this.fn);
+      case "CONSUME":
+        return this.queue.grantConsumeMessages(this.fn);
+      default:
+        return this.queue.grantSendMessages(this.fn);
```
kylelaker-ccpo, line 216: Is it really appropriate to just grant `send` permissions on any invalid value? Why is that not an error?

infrastructure/lib/constructs/lambda-fn.ts
```diff
@@ -0,0 +1,219 @@
...
+    switch (bucketPermissions) {
+      case "READ":
+        return this.bucket.grantRead(this.fn);
+      case "PUT":
+        return this.bucket.grantPut(this.fn);
+      case "DELETE":
+        return this.bucket.grantDelete(this.fn);
+      default:
+        return this.bucket.grantRead(this.fn);
+    }
```
kylelaker-ccpo, line 206: What if someone were to pass in `"FOO"` should they really be allowed to read?

infrastructure/lib/constructs/lambda-fn.ts
```diff
@@ -0,0 +1,219 @@
...
+      case "ALL":
+        return this.table.grantFullAccess(this.fn);
+      case "READ":
+        return this.table.grantReadData(this.fn);
+      case "READ_WRITE":
+        return this.table.grantReadWriteData(this.fn);
+      case "WRITE":
+        return this.table.grantWriteData(this.fn);
+      default:
+        return this.table.grantReadWriteData(this.fn);
```
kylelaker-ccpo, line 192: We should not default to granting write permissions.

infrastructure/lib/constructs/lambda-fn.ts
```diff
@@ -0,0 +1,219 @@
...
+  readonly batchSize?: number;
+
+  /**
+   * The State Machine that will start execution when this function
+   * is called.
+   */
+  readonly stateMachine?: sfn.IStateMachine;
+}
+
+export class ApiFlexFunction extends cdk.Construct {
```
kylelaker-ccpo, line 108: We really need docs on what this does. I mean, we're granting permissions, gluing random stuff together, and adding env vars all over the place. And we have to customize the properties and constructor every time we want to change anything. We haven't decorated anything; we've just crammed 4 classes into one.

`ApiFlexFunction` is not a great name for this. But we've done far too much here to even call it `ApiLambaPermissionGranter`

kylelaker-ccpo (commented):
Was there really not any code to delete as part of this? Like did we not remove all the references to at least one of the classes/constructs this PR/ticket were created to remove?

infrastructure/lib/atat-web-api-stack.ts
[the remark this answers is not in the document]
jasonburkert-ccpo: @kylelaker-ccpo To clarify, would you expect that there would be more direct invocations of `ApiFunctiontest()` or its equivalent and the elimination of `addDatabaseApiFunction()` and `addQueueDatabaseApiFunction()`?

infrastructure/lib/constructs/lambda-fn.ts
jasonburkert-ccpo: I think `ApiFunction` was an obvious choice but it's used in infrastructure, and I'm guessing that this was a lazy way to create a unique name.  Zach was looking for a better name for the construct.  I made a few suggestions elsewhere.

Re: props name, I suppose that the props for `ApiFunctiontest` should be `ApiFunctiontestProps` if keeping with the convention.

infrastructure/lib/constructs/lambda-fn.ts
jasonburkert-ccpo: I should have caught this.  Type should be `TablePermissions`

infrastructure/lib/constructs/lambda-fn.ts
jasonburkert-ccpo: Type should be `QueuePermissions`

infrastructure/lib/constructs/lambda-fn.ts
jasonburkert-ccpo: Type should be `BucketPermissions` or equivalent

infrastructure/lib/constructs/lambda-fn.ts
jasonburkert-ccpo: I should have caught this as well.  Agree this default should be `grantReadData()`, or maybe nothing at all.

infrastructure/lib/constructs/lambda-fn.ts
jasonburkert-ccpo: Yeah, agree not a great default behavior upon viewing this with fresh eyes.

infrastructure/lib/constructs/lambda-fn.ts
jasonburkert-ccpo: Yeah, agree not a great default behavior upon viewing this with fresh eyes.

PR #466, opened 2021-11-23
Update dependencies

kylelaker-ccpo:
- cdk-nag is updated to 1.0.4
 - aws-cdk is updated to 1.134
 - aws-sdk v3 is updated to 3.42.0
 - typescript is unified at 4.5.2

Some of these were already in use in some packages but not all. This
syncs the dependencies up everywhere.


kylelaker-ccpo:
It looks like some of these didn't really need the `package.json` update (which is why `package-lock.json` didn't update) but this clarifies that we may be using the features in the newer versions of the libraries.

juliusfitzhugh-ccpo (approved):
Nice! I think not having to worry about too many `package-lock.json` files helps because they can become large and are not as straight forward (maybe I just need more time looking at them).

PR #467, opened 2021-11-23
ATAT Internal API spec + APIGW instance

jeffsegal-ccpo:
This includes the first cut of new ATAT Internal API, which will eventually supplant the ATAT Provisioning Wizard API. No operations are implemented yet. We expect to start filling them out next Sprint.

Ticket: AT-6875

### DIFF
--- a/atat_internal_api.yaml
+++ b/atat_internal_api.yaml
@@ -2,7 +2,7 @@ openapi: 3.0.2
 info:
   description: >-
     This is the ATAT Internal API. It has operations for creating, updating and destroying ATAT resources, as well as
-    for obtaining reporting data on provisioned Portfolios.
+    for obtaining reporting data about provisioned Portfolios.
   version: 1.0.0
   title: ATAT Internal API
   contact:
@@ -51,7 +51,7 @@ paths:
     post:
       operationId: createPortfolio
       requestBody:
-        description: Use an empty String to create an empty Portfolio or pass a complete object to create a complete version.
+        description: Use an empty request body to create an empty Portfolio or pass a complete object to create a complete version.
         content:
           application/json:
             schema:
@@ -189,7 +189,7 @@ paths:
             - cognitoAuthorizer: []
           - Ref: AWS::NoValue
     delete:
-      description: Deletes a Portfolio (hard delete if not provisioned, soft delete otherwise)
+      description: Deletes a Portfolio (soft delete if provisioned, hard delete otherwise)
       operationId: deletePortfolio
       parameters:
         - name: portfolioId
@@ -710,7 +710,7 @@ paths:
               schema:
                 $ref: '#/components/schemas/EnvironmentDetail'
         '404':
-          description: Portfolio, Application ot Environment with the given ID does not exist
+          description: Portfolio, Application or Environment with the given ID does not exist
       x-amazon-apigateway-request-validator: "request-params-validator"
       x-amazon-apigateway-integration:
         httpMethod: "POST"
@@ -1735,8 +1735,8 @@ components:
         code:
           type: string
           enum:
-          - INVALID_INPUT
-          - OTHER
+            - INVALID_INPUT
+            - OTHER
         message:
           type: string
       required:
@@ -1973,8 +1973,9 @@ components:
           items:
             $ref: "#/components/schemas/Clin"
         taskOrderFile:
-          description: "Metadata associated with file which was previously uploaded\
-            \ by POST /taskOrderFiles"
+          description: >-
+            Metadata associated with file which was previously uploaded
+            by POST /taskOrderFiles
           allOf:
             - $ref: "#/components/schemas/FileMetadataSummary"
         taskOrderNumber:
@@ -1998,9 +1999,10 @@ components:
       additionalProperties: false
       description: "Metadata describing an uploaded file"
     PortfolioAccess:
-      description: "Represents a set of Operators who should be granted access to\
-        \ a Portfolio at a specific access level. Uses an enum construct for forward\
-        \ compatibility should other access levels be specified."
+      description: >-
+        Represents a set of Operators who should be granted access to
+        a Portfolio at a specific access level. Uses an enum construct for forward
+        compatibility should other access levels be specified.
       properties:
         portfolioAdministrators:
           type: array
@@ -2008,9 +2010,10 @@ components:
             type: string
             format: email
     AppEnvAccess:
-      description: "Represents a set of Operators who should be granted access to\
-        \ an Application or Environment at a specific access level. The same access\
-        \ levels are available for both Applications & Environments."
+      description: >-
+        Represents a set of Operators who should be granted access to
+        an Application or Environment at a specific access level. The same access
+        levels are available for both Applications & Environments.
       properties:
         administrators:
           type: array
--- a/atat_internal_api.yaml
+++ b/atat_internal_api.yaml
@@ -1,0 +1,2197 @@
+openapi: 3.0.2
+info:
+  description: >-
+    This is the ATAT Internal API. It has operations for creating, updating and destroying ATAT resources, as well as
+    for obtaining reporting data on provisioned Portfolios.
+  version: 1.0.0
+  title: ATAT Internal API
+  contact:
+    email: atat-dev+provisioning_api@ccpo.mil
+tags:
+  - name: CRUD
+    description: >-
+      Operations related to managing Portfolios, Applications and Environments
+paths:
+  /portfolios:
+    get:
+      operationId: getPortfolios
+      parameters:
+        - $ref: '#/components/parameters/offsetParam'
+        - $ref: '#/components/parameters/limitParam'
+      responses:
+        '400':
+          description: Invalid request parameter
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+        '200':
+          description: Should return up to {limitParam} items
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: "#/components/schemas/PortfolioSummary"
+              examples:
+                PortfolioSummaries:
+                  $ref: '#/components/examples/PortfolioSummaries'
+      x-amazon-apigateway-request-validator: "request-params-validator"
+      x-amazon-apigateway-integration:
+        httpMethod: "POST"
+        uri:
+          Fn::Sub: "arn:${AWS::Partition}:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${NotImplementedFunction.Arn}/invocations"
+        type: "aws_proxy"
+      security:
+        Fn::If:
+          - IsAuthorizationRequired
+          -
+            - cognitoAuthorizer: []
+          - Ref: AWS::NoValue                  
+    post:
+      operationId: createPortfolio
+      requestBody:
+        description: Use an empty String to create an empty Portfolio or pass a complete object to create a complete version.
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/Portfolio'
+          text/plain:
+            schema:
+              type: string
+              example: ''
+      responses:
+        '201':
+          description: Successful operation
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Portfolio"
+              examples:
+                PortfolioRequestBody:
+                  $ref: '#/components/examples/PortfolioResponseBody'
+        '400':
+          description: Invalid input
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ValidationError'
+      x-amazon-apigateway-request-validator: "request-params-validator"
+      x-amazon-apigateway-integration:
+        httpMethod: "POST"
+        uri:
+          Fn::Sub: "arn:${AWS::Partition}:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${NotImplementedFunction.Arn}/invocations"
+        type: "aws_proxy"
+      security:
+        Fn::If:
+          - IsAuthorizationRequired
+          -
+            - cognitoAuthorizer: []
+          - Ref: AWS::NoValue
+    options:
+      tags:
+        - cors
+      description: CORS headers
+      responses:
+        '200':
+          $ref: '#/components/responses/CorsHeaders'
+      x-amazon-apigateway-integration:
+        type: mock
+        requestTemplates:
+          application/json: |
+            {
+              "statusCode" : 200
+            }
+        responses:
+          default:
+            statusCode: "200"
+            responseParameters:
+              method.response.header.Access-Control-Allow-Headers: '''*'''
+              method.response.header.Access-Control-Allow-Methods: '''*'''
+              method.response.header.Access-Control-Allow-Origin: '''*'''
+            responseTemplates:
+              application/json: "{}"
+  /portfolios/{portfolioId}:
+    get:
+      description: Gets a detailed view of a Portfolio
+      operationId: getPortfolio
+      parameters:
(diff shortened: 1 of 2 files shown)


infrastructure/lib/atat-web-api-stack.ts
```diff
@@ -50,7 +50,8 @@ export interface AtatWebApiStackProps extends cdk.StackProps {
 
 export class AtatWebApiStack extends cdk.Stack {
   private readonly environmentId: string;
-  public readonly restApi: apigw.IRestApi;
+  public readonly draftApi: apigw.IRestApi;
```
kylelaker-ccpo, line 53:
```suggestion
  /**
   * @deprecated Use {@link internaApi} instead
   */
  public readonly draftApi: apigw.IRestApi;
```

infrastructure/lib/atat-web-api-stack.ts
```diff
@@ -149,22 +150,35 @@ export class AtatWebApiStack extends cdk.Stack {
...
+      Location: internalApiAsset.s3ObjectUrl,
+    });
 
     // And with the data now loaded from the template, we can use ApiDefinition.fromInline to parse it as real
     // OpenAPI spec (because it was!) and now we've got all our special AWS values and variables interpolated.
     // This will get used as the `Body:` parameter in the underlying CloudFormation resource.
-    const apiGateway = new SecureRestApi(this, "AtatSpecTest", {
-      restApiName: `${props.environmentId} API`,
-      apiDefinition: apigw.ApiDefinition.fromInline(apiSpecAsTemplateInclude),
+    const draftApiGateway = new SecureRestApi(this, "AtatDraftApi", {
```
kylelaker-ccpo, line 172: This `id` change (`AtatSpecTest`->`AtatDraftApi`) is going to result in a change in the underlying API URL. Will be interesting to see how sturdy our configuration endpoint is :)

infrastructure/lib/atat-web-api-stack.ts
```diff
@@ -269,14 +283,22 @@ export class AtatWebApiStack extends cdk.Stack {
...
+        description: "URL for the Draft API Gateway instance",
+        stringValue: draftApiGateway.urlForPath(),
+        parameterName: `/atat/${this.environmentId}/draftApi/url`,
+      })
+    );
+    this.ssmParams.push(
+      new ssm.StringParameter(this, "InternalApiGatewayUrl", {
+        description: "URL for the Internal API Gateway instance",
+        stringValue: draftApiGateway.urlForPath(),
+        parameterName: `/atat/${this.environmentId}/internalApi/url`,
```
kylelaker-ccpo, line 301: ahh nvm we've broken the configuration endpoint's ability to determine the API URL with this change. but I do like that we'll move to `internalApi`-only eventually which is more descriptive anyway

kylelaker-ccpo (changes requested):

infrastructure/lib/atat-web-api-stack.ts
jeffsegal-ccpo: Ahhh - if I understand the design correctly then I think it should still work but I could be wrong.

atat_internal_api.yaml
```diff
@@ -0,0 +1,2197 @@
...
+            $ref: "#/components/schemas/Clin"
+        taskOrderFile:
+          description: "Metadata associated with file which was previously uploaded\
+            \ by POST /taskOrderFiles"
+          allOf:
+            - $ref: "#/components/schemas/FileMetadataSummary"
+        taskOrderNumber:
+          pattern: "^[\\d]{13,17}$"
+          type: string
+          description: "TO numbers are 13 characters. TO modifications are 17 characters."
```
zachclark-ccpo, line 1977: I just want to point out that the business rules for this changed, and I believe there is tech debt to change the pattern to reflect it. Biggest change is making character 9 an F, but full list is on 2.1 of: https://ccpo.atlassian.net/wiki/spaces/AT/pages/2031747073/ATAT+Business+Rules+Data+Fields

atat_internal_api.yaml
```diff
@@ -0,0 +1,2197 @@
...
+        type: "aws_proxy"
+      security:
+        Fn::If:
+          - IsAuthorizationRequired
+          -
+            - cognitoAuthorizer: []
+          - Ref: AWS::NoValue
+    put:
+      operationId: updateApplication
+      parameters:
```
zachclark-ccpo, line 423: Missing description for updateApplication

atat_internal_api.yaml
```diff
@@ -0,0 +1,2197 @@
...
+          description: Invalid input
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ValidationError'
+      x-amazon-apigateway-request-validator: "request-params-validator"
+      x-amazon-apigateway-integration:
+        httpMethod: "GET"
+        uri:
+          Fn::Sub: "arn:${AWS::Partition}:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${NotImplementedFunction.Arn}/invocations"
```
zachclark-ccpo, line 458: https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-swagger-extensions-integration.html

> The HTTP method used in the integration request. For Lambda function invocations, the value must be POST.
```suggestion
      x-amazon-apigateway-integration:
        httpMethod: "POST"
        uri:
          Fn::Sub: "arn:${AWS::Partition}:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${NotImplementedFunction.Arn}/invocations"
```
It should be a POST for everywhere in this file, won't work as a lambda proxy request otherwise.

atat_internal_api.yaml
```diff
@@ -0,0 +1,2197 @@
...
+        type: "aws_proxy"
+      security:
+        Fn::If:
+          - IsAuthorizationRequired
+          -
+            - cognitoAuthorizer: []
+          - Ref: AWS::NoValue
+    post:
+      operationId: createEnvironment
+      parameters:
```
zachclark-ccpo, line 614: Add description here

atat_internal_api.yaml
```diff
@@ -0,0 +1,2197 @@
...
+        type: "aws_proxy"
+      security:
+        Fn::If:
+          - IsAuthorizationRequired
+          -
+            - cognitoAuthorizer: []
+          - Ref: AWS::NoValue
+    put:
+      operationId: updateEnvironment
+      parameters:
```
zachclark-ccpo, line 726: Add description here

atat_internal_api.yaml
```diff
@@ -0,0 +1,2197 @@
...
+        type: "aws_proxy"
+      security:
+        Fn::If:
+          - IsAuthorizationRequired
+          -
+            - cognitoAuthorizer: []
+          - Ref: AWS::NoValue
+    post:
+      operationId: createTaskOrder
+      parameters:
```
zachclark-ccpo, line 931: Add description here

atat_internal_api.yaml
```diff
@@ -0,0 +1,2197 @@
...
+      security:
+        Fn::If:
+          - IsAuthorizationRequired
+          -
+            - cognitoAuthorizer: []
+          - Ref: AWS::NoValue
+    put:
+      operationId: updateTaskOrder
+      parameters:
+        - name: portfolioId
```
zachclark-ccpo, line 1032: Add description here

atat_internal_api.yaml
```diff
@@ -0,0 +1,2197 @@
...
+      security:
+        Fn::If:
+          - IsAuthorizationRequired
+          -
+            - cognitoAuthorizer: []
+          - Ref: AWS::NoValue
+    post:
+      operationId: createClin
+      parameters:
+        - name: portfolioId
```
zachclark-ccpo, line 1183: Add description here

atat_internal_api.yaml
```diff
@@ -0,0 +1,2197 @@
...
+        contributors:
+          type: array
+          items:
+            type: string
+            format: email
+        readOnlyOperators:
+          type: array
+          items:
+            type: string
+            format: email
```
zachclark-ccpo, line 2023:
```suggestion
    AppEnvAccess:
      description: "Represents a set of Operators who should be granted access to\
        \ an Application or Environment at a specific access level. The same access\
        \ levels are available for both Applications & Environments."
      required:
          - administrators
          - contributors
          - readOnlyOperators
      properties:
        administrators:
          type: array
          items:
            type: string
            format: email
        contributors:
          type: array
          items:
            type: string
            format: email
        readOnlyOperators:
          type: array
          items:
            type: string
            format: email
```
Are any of these properties required? If so:
I suggest adding a required field validation check, so the frontend knows to match the `AppEnvAccess` properties of adminitrators, contributors, and readOnlyOperators. They could send empty arrays, which is probably fine, but if it isn't fine I recommend adding a `minItems: 1` to the arrays.

atat_internal_api.yaml
```diff
@@ -0,0 +1,2197 @@
...
+    User:
+      properties:
+        email:
+          type: string
```
zachclark-ccpo, line 2027: Should we have 
```
required:
- email
- firstName
- lastName
```
here?

zachclark-ccpo (changes requested):
The number one change that needs to be addressed is changing every instance of
```
x-amazon-apigateway-integration:
        httpMethod: "GET"
```
with
```
x-amazon-apigateway-integration:
        httpMethod: "POST"
```
because it must always be a POST for lambda proxy integration to work.

https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-swagger-extensions-integration.html

> The HTTP method used in the integration request. For Lambda function invocations, the value must be POST.

atat_internal_api.yaml
kylelaker-ccpo: Some something more like

```regex
(?<DoDAAC>[a-z0-9]{6})
(?<year>\d{2})
F
(?<ToId>\d{4})
(?<ModId>\d{4})?
```

But probably without the named capture groups, so:

```
[a-z0-9]{6}\d{2}F\d{4}(?:\d{4})?
```

Technically that last group probably could be capturing as well (so remove `?:`) if we really wanted to simply it further. But it has to be grouped because `\d{4}?` means something different.

But is this a spec problem or a business logic problem?

atat_internal_api.yaml
```diff
@@ -0,0 +1,2197 @@
...
+        - $ref: "#/components/schemas/BaseObject"
+      properties:
+        clins:
+          minItems: 1
+          type: array
+          items:
+            $ref: "#/components/schemas/Clin"
+        taskOrderFile:
+          description: "Metadata associated with file which was previously uploaded\
+            \ by POST /taskOrderFiles"
```
kylelaker-ccpo, line 1971:
```suggestion
          description: >-
            Metadata associated with file which was previously uploaded
            by POST /taskOrderFiles
```

https://yaml-multiline.info/

There are a few other places throughout this file where we should do this too. A search for `\s+\\` should turn most of them up.

atat_internal_api.yaml
```diff
@@ -0,0 +1,2197 @@
...
+          description: "name of file when uploaded"
+        id:
+          type: string
+          description: "Should be a UUID. Will be the S3 Object Key."
+      additionalProperties: false
+      description: "Metadata describing an uploaded file"
+    PortfolioAccess:
+      description: "Represents a set of Operators who should be granted access to\
+        \ a Portfolio at a specific access level. Uses an enum construct for forward\
+        \ compatibility should other access levels be specified."
```
kylelaker-ccpo, line 1997:
```suggestion
      description: >-
        Represents a set of Operators who should be granted access to
        a Portfolio at a specific access level. Uses an enum construct for forward
        compatibility should other access levels be specified.
```

atat_internal_api.yaml
```diff
@@ -0,0 +1,2197 @@
...
+      properties:
+        portfolioAdministrators:
+          type: array
+          items:
+            type: string
+            format: email
+    AppEnvAccess:
+      description: "Represents a set of Operators who should be granted access to\
+        \ an Application or Environment at a specific access level. The same access\
+        \ levels are available for both Applications & Environments."
```
kylelaker-ccpo, line 2007:
```suggestion
      description: >-
        Represents a set of Operators who should be granted access to
        an Application or Environment at a specific access level. The same access
        levels are available for both Applications & Environments.
```

atat_internal_api.yaml
```diff
@@ -0,0 +1,2197 @@
...
+      gatewayresponse.header.Access-Control-Allow-Methods: '''*'''
+      gatewayresponse.header.Access-Control-Allow-Origin: '''*'''
+    responseTemplates:
+      application/json: |
+        {
+          "code": "OTHER",
+          "message": "Authentication is required for this endpoint"
+        }
+  # UNSUPPORTED_MEDIA_TYPE is handled by DEFAULT_4XX
+  # WAF_FILTERED is handled by DEFAULT_4XX
```
kylelaker-ccpo, line 2197:
```suggestion
  # WAF_FILTERED is handled by DEFAULT_4XX

```

An EOL at the end of the file keeps things a little cleaner when interacting with CLI and other tools.

atat_internal_api.yaml
```diff
@@ -0,0 +1,2197 @@
...
+    Error:
+      type: object
+      additionalProperties: false
+      description: Generic error model
+      properties:
+        code:
+          type: string
+          enum:
+          - INVALID_INPUT
+          - OTHER
```
kylelaker-ccpo, line 1733:
```suggestion
          enum:
            - INVALID_INPUT
            - OTHER
```

Minor formatting thing in a few places. Can we be consistent and indent list items?

kylelaker-ccpo (changes requested):
Minor formatting things

atat_internal_api.yaml
```diff
@@ -0,0 +1,2197 @@
...
+tags:
+  - name: CRUD
+    description: >-
+      Operations related to managing Portfolios, Applications and Environments
+paths:
+  /portfolios:
+    get:
+      operationId: getPortfolios
+      parameters:
+        - $ref: '#/components/parameters/offsetParam'
```
robertmccardell-ccpo, line 19: Is there any way to add a parameter that will support returning  portofolios based on lifecycle_status (Active, Draft, Archived)?

atat_internal_api.yaml
jasonburkert-ccpo: Agree the old pattern was based on limited business rules that have since been replaced.  Good note Zach.

Thanks for breaking down the new rules in a regular expression Kyle.

I think through his final question Kyle means that the purpose of this ticket/PR is new API spec, and yes, the old API spec did have the latent business rule problem.  Let's not address this here.  I've added a comment on business rules page to point to Kyle's regex comment for future reference when we do address this problem.

atat_internal_api.yaml
jeffsegal-ccpo: I don't think we need to make them required on the API side; we just assume that there are no operators with that particular access level if we never get them in a PATCH.

jeffsegal-ccpo: These aren't required so let's leave it as-is for now. Plus we may end up redesigning this depending on how the discussions go next week.

atat_internal_api.yaml
jeffsegal-ccpo: We'll be designing the filtering and sorting capabilities soon. Most likely, it'll either be a `q` parameter which will contain a full query object, or specifically named set of queries for each filter that we need to support.

zachclark-ccpo (approved):

juliusfitzhugh-ccpo (approved):
Feedback from platform team has been addressed.

PR #480, opened 2021-11-29
Add local typeorm config to gitignore

jasonburkert-ccpo:
Add file `ormconfig.json` to `.gitignore`

This file is for configuration of typeorm for local development.
Reference: https://typeorm.io/#/using-ormconfig/using-ormconfigjson

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,4 +25,4 @@ tags-private.json
 .scannerwork
 
 # typeorm local config
-packages/api/ormconfig.json
+ormconfig.json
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -22,4 +22,7 @@ tags-private.json
 
 # Sonar Scanner
 *.ucfg
-.scannerwork
\ No newline at end of file
+.scannerwork
+
+# typeorm local config
+packages/api/ormconfig.json
```
kylelaker-ccpo, line 28:
```suggestion
# typeorm local config
ormconfig.json
```

This should have it ignore the file regardless of what directory it's in. Thoughts?

jasonburkert-ccpo: Yeah, that makes sense.

kylelaker-ccpo (approved):

PR #481, opened 2021-11-29
Generate 'schema.json' before build and test

jasonburkert-ccpo:
- Adds `scripts/generate-schemas.ts` to create file `schema.json`
- The new script is called before _build_ and _test_ via [npm Pre Scripts](https://docs.npmjs.com/cli/v6/using-npm/scripts#pre--post-scripts) `prebuild` and `pretest`
- Modifies github workflow to run `prebuild` before cdk synth
- Removes `schema.json` from repo and adds to .gitignore
- Removes generation of `schema.json` from web api stack


kylelaker-ccpo (changes requested):
Can you add a note about the need to run `npm prebuild` in README.md? It seems like it's only necessary if/when you skip `npm run build` and jump straight into CDK commands? Or do I misunderstand?

package.json
```diff
@@ -7,7 +7,9 @@
     "watch": "lerna run watch",
     "test": "lerna run test",
     "lint": "lerna run lint",
-    "lint:fix": "lerna run lint:fix"
+    "lint:fix": "lerna run lint:fix",
+    "prebuild": "npx ts-node scripts/generate-schemas.ts",
+    "pretest": "npx ts-node scripts/generate-schemas.ts"
```
juliusfitzhugh-ccpo, line 12: `prebuild` and `pretest` are the same,  are they suppose to be different or is there a reason for having both?

zachclark-ccpo (commented):
So before doing a `cdk deploy`, one must do an `npm run pre build`?

package.json
kylelaker-ccpo: These are `npm` `pre-` hooks. So you can create a `prefoo` that runs before `foo`. Theoretically, it may be possible to do:

```
"generate-schema": "npx ts-node scripts/generate-schemas.ts",
"prebuild": "npm run generate-schema",
"pretest": npm run generate-schema",
```

to somewhat de-duplicate. https://www.yld.io/blog/using-npm-pre-and-post-hooks/

kylelaker-ccpo:
> So before doing a `cdk deploy`, one must do an `npm run pre build`?

One option we may have would be to run `npm run cdk ...` instead of just `cdk`. Then we can have a `precdk` script/hook to automate this for us!

package.json
jasonburkert-ccpo: Yep.  I was just happy to get it to work.
Would `precdk` be similar?

jasonburkert-ccpo:
> Can you add a note about the need to run `npm prebuild` in README.md? It seems like it's only necessary if/when you skip `npm run build` and jump straight into CDK commands? Or do I misunderstand?

@kylelaker-ccpo 
Yes, that's right.  The two prescripts handle build and test, but since the cdk synth didn't call either it was necessary to add because schema.json wasn't present.

I like your precdk idea as an alternative.

kylelaker-ccpo:
> > Can you add a note about the need to run `npm prebuild` in README.md? It seems like it's only necessary if/when you skip `npm run build` and jump straight into CDK commands? Or do I misunderstand?
> 
> @kylelaker-ccpo Yes, that's right. The two prescripts handle build and test, but since the cdk synth in the GH workflow didn't call either it was necessary to add because schema.json wasn't present.
> 
> I like your precdk idea as an alternative.

We'd have to change other things to make that work. I'd rather just get the README addition in and have this fix ASAP. We can reconsider `npm run cdk` later.

zachclark-ccpo (changes requested):
We need to update the convertSchema function, because it is currently hardcoded to use:
```
    const schema = utils.packageRoot() + "/../atat_provisioning_wizard_api.yaml";
```
Now that we have 2 apis, which have _slightly_ different schema, we need to generate another schema for the `atat_internal_api.yaml`

jasonburkert-ccpo:
@kylelaker-ccpo Added the notes to `README.md` in 23d3b383dd02f536cd6f4bc4d38856fe6cc0b28f

zachclark-ccpo (approved):
Generating 2 schemas issue now resolved see the commit [880210c](https://github.com/dod-ccpo/atat-web-api/pull/481/commits/880210c668edb4953fa9a3fdddb1858a885ec906)

jasonburkert-ccpo:
@zachclark-ccpo Thanks for noticing the need and adding that change.

PR #489, opened 2021-11-30
Add initial TypeORM migration

jasonburkert-ccpo:
Creates tables `portfolio`, `application`, `environment`, `task_order`, `clin` each with primary key constraint named `pk_<table>`

Foreign key constraints are named `fk_<childtable_parenttable>`

Creates types `provisioning_status_enum`, `csp_enum`, `file_scan_status_enum`, `dod_component_enum`

Adds model `DodComponent` agreeing with values from API spec

Migration file created with `npx typeorm migration:create -n initial`
Initial DDL from pg_dump output, then revised in response to PR comments.

>The migration:run and migration:revert commands only work on .js files.
Thus the typescript files need to be compiled before running the commands.

After transpiling the migration file with `tsc -w`, executed `npx typeorm migration:run` and `npx typeorm migration:revert` many times in succession successfully changing a local postgres instance running in docker.

Migration pulls in SQL statements from files named `<migrationfile_base>-up.sql` and `<migrationfile_base>-down.sql`


packages/api/orm/migration/1638312144983-initial.ts
```diff
@@ -0,0 +1,232 @@
...
+    );
+    
+    ALTER TYPE public.environment_status_enum OWNER TO postgres;
+    
+    CREATE TYPE public.portfolio_csp_enum AS ENUM (
+        'CSP A',
+        'CSP B'
+    );
+    
+    ALTER TYPE public.portfolio_csp_enum OWNER TO postgres;
```
kylelaker-ccpo, line 63: How miserable is it to migrate enum's later? Let's say CSPs are later "Foo" and "Bar". We're looking at a migration that:

1. Adds the new variants
2. Does something with all As and Bs (either renaming to Foo or Bar or dropping them)
3. Deletes the old variants

?

packages/api/orm/migration/1638312144983-initial.ts
```diff
@@ -0,0 +1,232 @@
...
+    CREATE EXTENSION IF NOT EXISTS "uuid-ossp" WITH SCHEMA public;
+    
+    COMMENT ON EXTENSION "uuid-ossp" IS 'generate universally unique identifiers (UUIDs)';
+    
+    CREATE TYPE public.application_provisioningstatus_enum AS ENUM (
+        'not_started',
+        'in_progress',
+        'failed',
+        'complete'
+    );
```
kylelaker-ccpo, line 26: Can all of these share a `public.provisioning_status_enum` rather than defining one type per field?

packages/api/orm/migration/1638312144983-initial.ts
```diff
@@ -0,0 +1,232 @@
...
+    
+    ALTER TYPE public.task_order_filescanstatus_enum OWNER TO postgres;
+    
+    SET default_tablespace = '';
+    SET default_table_access_method = heap;
+    
+    CREATE TABLE public.application (
+        id uuid DEFAULT public.uuid_generate_v4() NOT NULL,
+        "createdAt" timestamp without time zone DEFAULT now() NOT NULL,
+        "updatedAt" timestamp without time zone DEFAULT now() NOT NULL,
```
kylelaker-ccpo, line 97: These times are always going to be UTC per the DB config I believe. I don't even know if you can configure that in the RDS Parameter Group. Is that okay?

packages/api/orm/migration/1638312144983-initial.ts
```diff
@@ -0,0 +1,232 @@
...
+    
+    CREATE TYPE public.application_provisioningstatus_enum AS ENUM (
+        'not_started',
+        'in_progress',
+        'failed',
+        'complete'
+    );
+    
+    
+    ALTER TYPE public.application_provisioningstatus_enum OWNER TO postgres;
```
kylelaker-ccpo, line 29: We probably want to talk about whether this should be `postgres` or `atat_api_admin`.

packages/api/orm/migration/1638312144983-initial.ts
```diff
@@ -0,0 +1,232 @@
...
+    CREATE TYPE public.task_order_filescanstatus_enum AS ENUM (
+        'pending',
+        'accepted',
+        'rejected'
+    );
+    
+    ALTER TYPE public.task_order_filescanstatus_enum OWNER TO postgres;
+    
+    SET default_tablespace = '';
+    SET default_table_access_method = heap;
```
kylelaker-ccpo, line 92: I *think* we can omit these because they're defaults. https://www.postgresql.org/docs/13/runtime-config-client.html

packages/api/orm/migration/1638312144983-initial.ts
kylelaker-ccpo: The default is owner is the user who performs the `CREATE TYPE` so we may want to consider whether `ALTER` is required https://www.postgresql.org/docs/13/sql-createtype.html

packages/api/orm/migration/1638312144983-initial.ts
```diff
@@ -0,0 +1,232 @@
...
+        name character varying,
+        value text
+    );
+    
+    
+    ALTER TABLE public.typeorm_metadata OWNER TO postgres;
+    
+    ALTER TABLE ONLY public.task_order
+        ADD CONSTRAINT "PK_15e973a09676013165565bc6243" PRIMARY KEY (id);
+
```
kylelaker-ccpo, line 202: We also should decide, stylistically, whether we want to add constraints as ALTERs after the fact or as part of the `CREATE TABLE`. https://www.postgresql.org/docs/13/sql-createtable.html

packages/api/orm/migration/1638312144983-initial.ts
```diff
@@ -0,0 +1,232 @@
...
+    );
+    
+    ALTER TABLE public.clin OWNER TO postgres;
+    
+    COMMENT ON COLUMN public.clin."clinNumber" IS 'contract line item number from task order, 0001 through 9999';
+    COMMENT ON COLUMN public.clin."idiqClin" IS 'indefinite-delivery, indefinite-quantity CLIN specific to JWCC';
+    COMMENT ON COLUMN public.clin."totalClinValue" IS 'full amount requested by customer from USG for project or contract';
+    COMMENT ON COLUMN public.clin."obligatedFunds" IS 'amount already disbursed by USG for project or contract to be spent during CLIN POP';
+    COMMENT ON COLUMN public.clin."popStartDate" IS 'start of POP during which funds can be spent';
+    COMMENT ON COLUMN public.clin."popEndDate" IS 'end of POP during which funds can be spent';
```
kylelaker-ccpo, line 131: It looks like modifying comments in TS/models _will_ require a TypeORM migration in order to update which gives me pause about whether so many fields should have them. Will this result in a significant maintenance burden? Does having these comments in the database provide value? The answer very well may be "No" and "Yes" respectively but it feels worth considering.

packages/api/orm/migration/1638312144983-initial.ts
```diff
@@ -0,0 +1,232 @@
...
+    SET statement_timeout = 0;
+    SET lock_timeout = 0;
+    SET idle_in_transaction_session_timeout = 0;
+    SET client_encoding = 'UTF8';
+    SET standard_conforming_strings = on;
+    SELECT pg_catalog.set_config('search_path', '', false);
+    SET check_function_bodies = false;
+    SET xmloption = content;
+    SET client_min_messages = warning;
+    SET row_security = off;
```
kylelaker-ccpo, line 15: Should these be part of the initial migration? Or the initial bootstrapping? Are these even really necessary?

kylelaker-ccpo (commented):
I've got a few thoughts/questions on this. Overall this first migration isn't nearly as terrifying as I had expected it to be so I'm looking forward to setting up the custom resource to run it and get to work.

Because of how long/involved the SQL is, I wonder if we should think about ways to store those as files and read them in. This means our IDEs (and GitHub) could do syntax checking/highlighting for us. May give us a little more freedom for comments too.

```ts
export class initial1638312144983 implements MigrationInterface {
  private static readonly migrationId = "1638312144983-initial";

  public async up(queryRunner: QueryRunner): Promise<void> {
    const sql = fs.readFileSync(`${this.migrationId}-up.sql`);
    await queryRunner.query(sql);
  }

  public async down(queryRunner: QueryRunner): Promise<void> {
    const sql = fs.readFileSync(`${this.migrationId}-down.sql`);
    await queryRunner.query(sql);
  }
}
```

packages/api/orm/migration/1638312144983-initial.ts
juliusfitzhugh-ccpo: If I had to choose, I would probably go with `atat_api_admin` to be clear and explcit. I am not sure `postgres` adds much value or context.

packages/api/orm/migration/1638312144983-initial.ts
juliusfitzhugh-ccpo: Do you suggest finding a different approach for the enums due to complex migration changes, or that we should think a bit more before we go down this path? Doing a quick search I found [PostgreSQL: ENUM is no Silver Bullet](https://medium.com/swlh/postgresql-3-ways-to-replace-enum-305861e089bc) that give alternate options that we can consider.

packages/api/orm/migration/1638312144983-initial.ts
juliusfitzhugh-ccpo: I like how all the constraints are in one place, but at the same time being added with the CREATE TABLE will probably reduce the code and keep the contraints closer to the table.

packages/api/orm/migration/1638312144983-initial.ts
jasonburkert-ccpo: Makes sense to use a more specific name.

jasonburkert-ccpo: Changed in c0482b7d84df671b89e575e58c0422fdc5bb6a29

packages/api/orm/migration/1638312144983-initial.ts
jasonburkert-ccpo: Removed in 3130a36b464102d41d157d051b8c705b1acdd983

packages/api/orm/migration/1638312144983-initial.ts
jasonburkert-ccpo: I wonder how many of these are exports of defaults.

packages/api/orm/migration/1638312144983-initial.ts
```diff
@@ -0,0 +1,232 @@
...
+        ADD CONSTRAINT "FK_4934116062d67f9aef3725972ff" FOREIGN KEY ("applicationId") REFERENCES public.application(id);
+    
+    ALTER TABLE ONLY public.clin
+        ADD CONSTRAINT "FK_b919a2f2b8cc2d03fbc3d3d9456" FOREIGN KEY ("taskOrderId") REFERENCES public.task_order(id);
+    
+    `;
+    await queryRunner.query(sql);
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {}
```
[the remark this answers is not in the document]
jasonburkert-ccpo: Should have many DROP statements to revert what up does.

packages/api/orm/migration/1638312144983-initial.ts
jasonburkert-ccpo: Working on the initial models led me to believe TZ would come from DB config.

packages/api/orm/migration/1638312144983-initial.ts
kylelaker-ccpo: Probably most? Certainly the two `SET`s I highlighted later on are defaults. So I get the feeling a lot of these may be too.

packages/api/orm/migration/1638312144983-initial.ts
jeffsegal-ccpo: Interesting food for thought, but I am still inclined to stick with `TYPE` rather than any of those options or build the validation at the app layer. If the values do need to change then we'll have to fix it with a migration regardless, but with a `TYPE` we can be sure that before our code even executes that our existing data set will be valid.

kylelaker-ccpo: I think we've decided that for now we're set on `enum`s. But I just want to make sure we have some idea of what we're going to do for this since we _know_ that that we're eventually going to have to.

packages/api/orm/migration/1638312144983-initial.ts
jasonburkert-ccpo: Replaced all with `provisioning_status_enum` in 8d94b985e33b4e93a9ef12e9a18d0a9abf0010b3

packages/api/orm/migration/1638312144983-initial.ts
jasonburkert-ccpo: Removed `default_tablespace` in a8639253ecd206acabff6bacc6d11b5e8939a0bc

jasonburkert-ccpo: Per linked docs, _heap_ is the default value for `default_table_access_method`.  Will remove.
> This parameter specifies the default table access method to use when creating tables or materialized views if the CREATE command does not explicitly specify an access method, or when SELECT ... INTO is used, which does not allow to specify a table access method. The default is heap.

jasonburkert-ccpo: Removed `default_table_access_method` in 75274233e973428925996f95b810cfffecb3ccea

jeffsegal-ccpo:
I was going to request that we come up with a naming convention that incorporated version numbers like `V1_1_0001_AddPortfolios.ts` but then looked at the TypeORM code and realized that unlike Flyway, it does not support this type of scheme. So, I think we'll just have to rely on timestamps for now. Most of the time this will be fine, but it may require some fudging to accommodate dependencies between simultaneous tickets or insert updates to ATAT versions that have already been released as part of a hotfix.

packages/api/orm/migration/1638312144983-initial.ts
jasonburkert-ccpo: After discussion, expecting UTC coming from DB config.

packages/api/orm/migration/1638312144983-initial.ts
jasonburkert-ccpo: Will include constraints in the CREATE TABLE statements while giving them more meaningful names as @jeffsegal-ccpo has suggested.

jasonburkert-ccpo: Moved PK constraints in 5828e63801f4228ab9b073cca7803543aae51d10

jasonburkert-ccpo: Moved FK constraints in a98c2c9d35fe3835f46e4e6faa67c81edd4c935b

packages/api/orm/migration/1638312144983-initial.ts
jasonburkert-ccpo: I've looked each setting in docs and found they are either already defaults or settings used by _pg_dump_.  Will remove all.

jasonburkert-ccpo: Removed in a08ce9d6ff25b8e25ad86c2b481b7c68b7c6a28b

packages/api/orm/migration/1638312144983-initial.ts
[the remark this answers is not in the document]
jasonburkert-ccpo: @juliusfitzhugh-ccpo It's a good question.
Added initial DROP statements in acce871fa62233fe4b65052621d5174fb99f08c2

jasonburkert-ccpo:
Addressed most feedback at this point.
Must test with `migration:run` and `migration:revert` commands.

Some questions:
1. Include table `typeorm_metadata` in this migration?
2. Should `dodComponents` enum values be included in a database type? At least for now? I'm leaning toward _yes_.
3. Same question as above but for `popStatus`, `fundingStatus`, `lifecycleStatus`.  These are computed, so leaning toward _no_.

packages/api/orm/migration/1638312144983-initial.ts
```diff
@@ -0,0 +1,140 @@
...
+        type character varying NOT NULL,
+        database character varying,
+        schema character varying,
+        "table" character varying,
+        name character varying,
+        value text
+    );
+    
+    ALTER TABLE typeorm_metadata OWNER TO atat_api_admin;
+    `;
```
zachclark-ccpo, line 121: Could this live in it's own file? And we just import it to this file?

zachclark-ccpo:
Should we be trying to use the typeorm migration API to do this? https://typeorm.io/#/migrations/using-migration-api-to-write-migrations

packages/api/orm/migration/1638312144983-initial.ts
jasonburkert-ccpo: It could.

packages/api/orm/migration/1638312144983-initial.ts
```diff
@@ -0,0 +1,128 @@
...
+        "createdAt" timestamp without time zone DEFAULT now() NOT NULL,
+        "updatedAt" timestamp without time zone DEFAULT now() NOT NULL,
+        "archivedAt" timestamp without time zone,
+        "provisioningStatus" provisioning_status_enum DEFAULT 'PENDING'::provisioning_status_enum NOT NULL,
+        administrators character varying[] DEFAULT '{}'::character varying[] NOT NULL,
+        contributors character varying[] DEFAULT '{}'::character varying[] NOT NULL,
+        "readOnlyOperators" character varying[] DEFAULT '{}'::character varying[] NOT NULL,
+        name character varying(100) NOT NULL,
+        description character varying(300),
+        owner character varying NOT NULL,
```
zachclark-ccpo, line 87:
```suggestion
        owner character varying NOT NULL,
```
Owner is a User with an email, first name, and last name
```
  User:
      required:
      - email
      - firstName
      - lastName
      properties:
        email:
          type: string
          format: email
        firstName:
          type: string
        lastName:
          type: string
```

packages/api/orm/migration/1638312144983-initial.ts
```diff
@@ -0,0 +1,128 @@
...
+        csp csp_enum NOT NULL,
+        "dodComponents" character varying[] NOT NULL,
```
zachclark-ccpo, line 89: Is there an ENUM for this? Because in the API spec it is:
```
dodComponents:
          type: array
          items:
            type: string
            enum:
              - AIR_FORCE
              - ARMY
              - MARINE_CORPS
              - NAVY
              - SPACE_FORCE
              - COMBATANT_COMMAND
              - JOINT_STAFF
              - DAFA
              - OSD_PSAS
       ```

Comment (jasonburkert-ccpo) [reply packages/api/orm/migration/1638312144983-initial.ts]: Added in 2d86e944efb2152e569e51936b737f0275e87e1f

File: packages/api/orm/migration/1638312144983-initial.ts [hunk 13]
```diff
@@ -0,0 +1,142 @@
...
+    
+    CREATE TABLE portfolio (
+        id uuid DEFAULT uuid_generate_v4() CONSTRAINT "pk_portfolio" PRIMARY KEY,
+        "createdAt" timestamp without time zone DEFAULT now() NOT NULL,
+        "updatedAt" timestamp without time zone DEFAULT now() NOT NULL,
+        "archivedAt" timestamp without time zone,
+        "provisioningStatus" provisioning_status_enum DEFAULT 'PENDING'::provisioning_status_enum NOT NULL,
+        administrators character varying[] DEFAULT '{}'::character varying[] NOT NULL,
+        contributors character varying[] DEFAULT '{}'::character varying[] NOT NULL,
+        "readOnlyOperators" character varying[] DEFAULT '{}'::character varying[] NOT NULL,
```
juliusfitzhugh-ccpo, line 47: Does the `portfolio` table need `administrators`, `contributors` and `readOnlyOperators`? And does `administrators` equate to `portfolioAdministrators` as shown in the new internal API Spec? I am only asking since  only `portfolioAdministrators` was available at the top level of the portfolio.

packages/api/orm/entity/Application.ts
```diff
@@ -3,7 +3,7 @@ import { Environment } from "./Environment";
 import { Portfolio } from "./Portfolio";
 import { ProvisionableEntity } from "./ProvisionableEntity";
 
-@Entity()
+@Entity("application")
```
juliusfitzhugh-ccpo, line 6: This explicitly define the name of the entity, rather than the default given by TypeORM?

juliusfitzhugh-ccpo (commented):
I had a few questions asking for clarity, otherwise nothing stood out to me. I like this version much better than the original one. Great Job!

packages/api/orm/migration/1638312144983-initial.ts
jasonburkert-ccpo: Good eye.  `administrators` does equate to `portfolioAdministrators`.  `contributors` and `readOnlyOperators` are not necessary per spec, but present due to entity inheritance from `ProvisionableEntity`.  This was by design, but a potential area of improvement in the entity modeling.

juliusfitzhugh-ccpo: Thanks for clearing that up.

packages/api/orm/migration/1638312144983-initial.ts
jasonburkert-ccpo: Implemented in e09df7c71ef60c48406483fd248e3188346dc508

jasonburkert-ccpo:
> Because of how long/involved the SQL is, I wonder if we should think about ways to store those as files and read them in. This means our IDEs (and GitHub) could do syntax checking/highlighting for us. May give us a little more freedom for comments too.

Implemented in e09df7c71ef60c48406483fd248e3188346dc508

jasonburkert-ccpo:
Since we don't yet have a SQL formatter configured, I used the default behavior of the [SQLTools extension](https://marketplace.visualstudio.com/items?itemName=mtxr.sqltools) for vscode.

packages/api/orm/migration/1638312144983-initial.ts
```diff
@@ -0,0 +1,15 @@
+import * as fs from "fs";
+import { join } from "path";
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class initial1638312144983 implements MigrationInterface {
+  private readonly migrationId = "1638312144983-initial";
```
jeffsegal-ccpo, line 6: This is fine for now but I am guessing that we can use reflection to avoid having to copy the name here next time.

jeffsegal-ccpo (approved):

kylelaker-ccpo (approved):

PR #491, opened 2021-12-01
Update to AWS CDK v2

kylelaker-ccpo:
v2 of the AWS CDK is expected to be released just about any time now.
This is prep work to make sure that we're ready for it. All that was
really necessary was updating package.json and imports as well as
changing around where a few classes live. We were not using any
experimental packages which is nice because none need to be pulled in.
Hopefully this greatly simplifies dependency management going forward.

We were already using modern values in `cdk.json`'s `context` block so
we don't need to worry about that. We may need to re-run `cdk bootstrap`
in order to use the newer version of that but even that seems to be
optional.


infrastructure/lib/atat-iam-stack.ts
```diff
@@ -79,14 +78,14 @@ export class AtatIamStack extends cdk.Stack {
       description: "Grants read/write access to developer-specific actions and resources",
       statements: [
         // During a deployment, outside of CloudFormation (where the role will be used),
-        // the CDK needs to read from and write to the S3 buckets for the CDK.
+        // the CDK needs to read from and write to the S3 buckets for the
```
juliusfitzhugh-ccpo, line 81: Not sure if you wanted to remove more, add something else, or if it was accidentally deleted. I made a suggestion for what I thought you were trying to do. 
```suggestion
        // the CDK needs to read from and write to the S3 buckets.
```

kylelaker-ccpo: If I told that I did a find and replace to do `s/cdk\.//i` would it surprise you that this happened :D Great catch. And honestly, yeah, this suggestion to just end the sentence makes more sense.

infrastructure/lib/constructs/database.ts
```diff
@@ -82,13 +88,13 @@ export class Database extends cdk.Construct {
   // Using a ConcreteDependable means that we can change the underlying definition of what it
   // means for this to be "ready" without having to update the API. We just need to add/change
   // resources in the ConcreteDependable.
```
juliusfitzhugh-ccpo, line 90: I suggest updating the name to avoid any confusion. Looking at the cdk docs for `DependencyGroup` it does not appear to have much of a difference (other than accepting an array of dependencies) and it appears your comment still holds true with the name change.
```suggestion
  // Using a DependencyGroup means that we can change the underlying definition of what it
  // means for this to be "ready" without having to update the API. We just need to add/change
  // resources in the DependencyGroup.
```

juliusfitzhugh-ccpo (approved):
I'm really liking the consolidation of packages into the `aws-cdk-lib` with v2. Nice work!

PR #496, opened 2021-12-01
Reuse connections

kylelaker-ccpo:


zachclark-ccpo:
> Once you created a connection you can obtain it anywhere from your app, using getConnection function:
```
import {getConnection} from "typeorm";

// can be used once createConnection is called and is resolved
const connection = getConnection();

// if you have multiple connections you can get connection by name
const secondConnection = getConnection("test2-connection");
```
https://typeorm.io/#/connection

kylelaker-ccpo:
@zachclark-ccpo This should be mostly equivalent to getting the `default` connection, right? Do you have specific code you'd prefer to see?

zachclark-ccpo (approved):

PR #500, opened 2021-12-02
Implement createEnvironment operation

juliusfitzhugh-ccpo:
Implement `POST /portfolios/:portfolioId/applications/:applicationId/environments`

Add an environment into the RDS postgres database for a given 
application of a portfolio using the new `atat-internal-api` spec. 
This is a part of the switch from DynamoDB to RDS database.

- Implement `createEnvironment` function
  - connect with `atat-web-api-stack` infrastructure
  - update `atat-internal-api` to use func
- Add `EnvironmentRepository` for simple and reusable interactions with the database
  - Include methods in `EnvironmentRepository` to be used with other endpoints
- Include new models/interfaces with the `Environment` entity
- Update `dbTest.ts` to create a portfolio and application where the environment could be included
- Implement Business Rule 3.3 logic to prevent duplicate environment names in an application
- Add 404 errors to `atat-internal-api` spec to account for non-validation errors (e.g., portfolio not found)
- All tests are skipped since they are only used for local deployment until a concrete solution is decided base on [AT-6969](https://ccpo.atlassian.net/browse/AT-6969).

### Timeout performance
It is known that a `timeout` of 5 seconds is too short for the lambda to execute and was increased to 10 seconds. This needs to be optimize for better UI performance. From basic testing of the `createEnvironment` func, it appears that this might be due to a cold start, since the initial invocation is a little over 6s, and the second invocation is about 500ms.

###  AJV validator issue
The combination of `allOf` and `additionalProperties: false` defined in the internal spec causes the middy validator to error because properties defined in `allOf` are ignored by `additionalProperties` and only accounts for properties defined under `properties` in the Environment Object (see [ajv - additionalProperties]( https://ajv.js.org/faq.html#additional-properties-inside-compound-keywords-anyof-oneof-etc) and [Schema composition with allOf](https://json-schema.org/understanding-json-schema/reference/combining.html#properties-of-schema-composition)).   

To get around this issue, `allOf` was removed and all the properties in `allOf` were placed under the properties of the Environment object. This allows the use of `additionalProperties: false` and ensures that there are no issues with the middy validator failing due to properties inside of `allOf`.  This also prevents manually structuring a schema to be used by the validator.

Ticket: AT-6891

### DIFF
--- a/atat_internal_api.yaml
+++ b/atat_internal_api.yaml
@@ -602,7 +602,7 @@ paths:
       x-amazon-apigateway-integration:
         httpMethod: "POST"
         uri:
-          Fn::Sub: "arn:${AWS::Partition}:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${NotImplementedFunction.Arn}/invocations"
+          Fn::Sub: "arn:${AWS::Partition}:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${CreateEnvironmentFunction.Arn}/invocations"
         type: "aws_proxy"
       security:
         Fn::If:
--- a/atat_internal_api.yaml
+++ b/atat_internal_api.yaml
@@ -644,6 +644,12 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ValidationError'
+        '404':
+          description: Portfolio or Application with the given ID does not exist
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
       x-amazon-apigateway-request-validator: "request-params-validator"
       x-amazon-apigateway-integration:
         httpMethod: "POST"
--- a/atat_internal_api.yaml
+++ b/atat_internal_api.yaml
@@ -1918,13 +1918,43 @@ components:
       required:
         - name
       type: object
-      allOf:
-        - $ref: "#/components/schemas/BaseObject"
-        - $ref: '#/components/schemas/AppEnvAccess'
       properties:
         name:
           pattern: "^[a-zA-Z\\d _-]{1,100}$"
           type: string
+      # Remove the addition of 'allOf' while still using 'additionalProperties'
+      # which adds duplication to the spec and is not ideal, but allows for
+      # validation to occur for the middy validator
+        updatedAt:
+          type: string
+          format: "date-time"
+          readOnly: true
+        createdAt:
+          type: string
+          format: "date-time"
+          readOnly: true
+        archivedAt:
+          type: string
+          format: "date-time"
+          readOnly: true
+        id:
+          type: string
+          readOnly: true
+        administrators:
+          type: array
+          items:
+            type: string
+            format: email
+        contributors:
+          type: array
+          items:
+            type: string
+            format: email
+        readOnlyOperators:
+          type: array
+          items:
+            type: string
+            format: email
       additionalProperties: false
       description: "Represents an Environment for a specific Application"
     Clin:
--- a/atat_internal_api.yaml
+++ b/atat_internal_api.yaml
@@ -602,7 +602,7 @@ paths:
       x-amazon-apigateway-integration:
         httpMethod: "POST"
         uri:
-          Fn::Sub: "arn:${AWS::Partition}:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${CreateEnvironmentFunction.Arn}/invocations"
+          Fn::Sub: "arn:${AWS::Partition}:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${NotImplementedFunction.Arn}/invocations"
         type: "aws_proxy"
       security:
         Fn::If:
--- a/atat_internal_api.yaml
+++ b/atat_internal_api.yaml
@@ -648,7 +648,7 @@ paths:
       x-amazon-apigateway-integration:
         httpMethod: "POST"
         uri:
-          Fn::Sub: "arn:${AWS::Partition}:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${NotImplementedFunction.Arn}/invocations"
+          Fn::Sub: "arn:${AWS::Partition}:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${CreateEnvironmentFunction.Arn}/invocations"
         type: "aws_proxy"
       security:
         Fn::If:
@@ -764,7 +764,7 @@ paths:
               schema:
                 $ref: '#/components/schemas/ValidationError'
         '404':
-          description: Portfolio, Application ot Environment with the given ID does not exist
+          description: Portfolio, Application or Environment with the given ID does not exist
       x-amazon-apigateway-request-validator: "request-params-validator"
       x-amazon-apigateway-integration:
         httpMethod: "POST"
--- a/packages/api/orm/repository/EnvironmentRepository.ts
+++ b/packages/api/orm/repository/EnvironmentRepository.ts
@@ -1,0 +1,61 @@
+import { EntityRepository, Repository, InsertResult, UpdateResult } from "typeorm";
+import { Environment, IEnvironment, EnvironmentUpdate } from "../entity/Environment";
+
+@EntityRepository(Environment)
+export class EnvironmentRepository extends Repository<Environment> {
+  // GET environments/:id
+  getEnvironment(id: string): Promise<Environment> {
+    return this.findOneOrFail({
+      select: [
+        "name",
+        "id",
+        "createdAt",
+        "updatedAt",
+        "archivedAt",
+        "administrators",
+        "contributors",
+        "readOnlyOperators",
+      ],
+      where: { id },
+    });
+  }
+
+  // GET all environments in an application
+  getEnvironmentsByApplicationId(applicationId: string): Promise<[Array<Environment>, number]> {
+    return this.createQueryBuilder("environment")
+      .select([
+        "environment.name",
+        "environment.id",
+        "environment.createdAt",
+        "environment.updatedAt",
+        "environment.archivedAt",
+      ])
+      .where("environment.applicationId = :applicationId", {
+        applicationId,
+      })
+      .getManyAndCount();
+  }
+
+  // POST create new environment
+  createEnvironment(environments: Array<IEnvironmentCreate>): Promise<InsertResult> {
+    return this.insert(environments);
+  }
+
+  // PUT update environment
+  updateEnvironment(id: string, changes: IEnvironmentUpdate): Promise<UpdateResult> {
+    return this.update(id, { ...changes });
+  }
+
+  // DELETE environment (hard delete)
+  async deleteEnvironment(id: string): Promise<IEnvironment> {
+    console.log("Deleting: " + id);
+    const environment = await this.findOneOrFail({
+      select: ["name", "id", "createdAt", "updatedAt", "archivedAt"],
+      where: { id },
+    });
+    const deleteResult = await this.delete(id);
+    console.log(`Deleted: ${environment}. Results: ${deleteResult}`);
+
+    return environment;
+  }
+}
--- a/packages/api/portfolios/environments/createEnvironment.test.ts
+++ b/packages/api/portfolios/environments/createEnvironment.test.ts
@@ -1,13 +1,13 @@
 import { handler } from "./createEnvironment";
 import { Context } from "aws-lambda";
 import { ApiGatewayEventParsed } from "../../utils/eventHandlingTool";
-import { IEnvironment } from "../../../orm/entity/Environment";
+import { IEnvironmentCreate } from "../../../orm/entity/Environment";
 import { SuccessStatusCode } from "../../utils/response";
 
 describe("createEnvironment", () => {
   it.skip("successful operation testing locally only", async () => {
-    const validRequest: ApiGatewayEventParsed<IEnvironment> = {
-      body: { name: "local test" },
+    const validRequest: ApiGatewayEventParsed<IEnvironmentCreate> = {
+      body: { name: "new local testing and a unique name" },
       pathParameters: {
         portfolioId: "91c6bd1f-5ed8-413b-8f85-55a62dd50ad3",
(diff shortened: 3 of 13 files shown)


packages/api/orm/repository/EnvironmentRepository.ts
```diff
@@ -0,0 +1,11 @@
+import { EntityRepository, Repository, InsertResult } from "typeorm";
+import { Environment } from "../entity/Environment";
+import { Application } from "../entity/Application";
+
+@EntityRepository(Environment)
+export class EnvironmentRepository extends Repository<Environment> {
+  // POST create new environment(s)
+  createEnvironment(environments: { name: string; application: Application }[]): Promise<InsertResult> {
```
jasonburkert-ccpo, line 8: Would 'parentApplication' add clarity?

atat_internal_api.yaml
```diff
@@ -1725,7 +1725,6 @@ components:
         id:
           type: string
           readOnly: true
-      additionalProperties: false
```
jasonburkert-ccpo, line 1728: Not such a great modifier for a BaseObject, huh?

packages/api/models/Environment.ts
```diff
@@ -1,6 +1,8 @@
+import { Application } from "../orm/entity/Application";
```
zachclark-ccpo, line 1: Do not import this Application, it is not the interface file, it is the Entity

packages/api/models/Environment.ts
```diff
@@ -1,6 +1,8 @@
...
 import { AppEnvOperator } from "./Operator";
 
 export interface Environment {
   name: string;
-  operators: Array<AppEnvOperator>;
+  application: Application;
```
zachclark-ccpo, line 6: Why do you have application here?
It is not required on the internal API yaml spec.

juliusfitzhugh-ccpo: Good point. I was probably thinking of it from the database side (which includes application) rather than the spec side. I did run into a conflict when trying to create an environment and `Application` was needed for a simple interface of `name` and `application`. I'll update accordingly.

packages/api/portfolios/environments/createEnvironment.ts
```diff
@@ -0,0 +1,143 @@
...
+  context?: Context
+): Promise<APIGatewayProxyResult> {
+  const parameters = Object.entries(event.pathParameters as APIGatewayProxyEventPathParameters);
+  if (
+    !parameters.every(
+      ([param, paramValue]) => ["portfolioId", "applicationId"].includes(param) && isValidUuidV4(paramValue)
+    )
+  ) {
+    throw createError(404, "Shape validation failed, invalid UUIDv4");
+  }
```
zachclark-ccpo, line 41: Do we have an error we can return for this?

packages/api/portfolios/environments/createEnvironment.ts
```diff
@@ -0,0 +1,143 @@
...
+  const applicationId = event.pathParameters!.applicationId;
+
+  if (event.body === undefined) {
+    throw createError(400, "Shape validation failed, invalid request body");
+  }
+  const environmentBody = event.body as IEnvironmentCreate;
+
+  const connection = await createConnection();
+  const insertedEnvironments: IEnvironmentCreate[] = [];
+
```
zachclark-ccpo, line 52: Could you put a comment here with "Set up DB connection" or something.

packages/api/portfolios/environments/createEnvironment.ts
```diff
@@ -0,0 +1,143 @@
...
+  type: internalSchema.Environment.type,
+  additionalProperties: internalSchema.Environment.additionalProperties,
+  required: internalSchema.Environment.required,
+  properties: {
+    ...internalSchema.Environment.properties,
+    ...internalSchema.Environment.allOf[0].properties,
+    ...internalSchema.Environment.allOf[1].properties,
+  },
+};
+
```
zachclark-ccpo, line 135: Is this still a blocker?

packages/api/portfolios/environments/createEnvironment.ts
zachclark-ccpo: Or is this just here to throw an error to the middleware

zachclark-ccpo (dismissed):
Preliminary review completed, I think you should try moving some of the portfolioId/applicationId validation logic into a separate file so we don't have to repeat it for the the other API requests, and update the API spec so we don't have to do this workaround:
```
// temporary work around for providing validation shape that does not fail
// due to the 'allOf' and 'additionalProperties' conflict when composing schemas
// see https://json-schema.org/understanding-json-schema/reference/combining.html#properties-of-schema-composition
const environmentSchema = {
  description: internalSchema.Environment.description,
  type: internalSchema.Environment.type,
  additionalProperties: internalSchema.Environment.additionalProperties,
  required: internalSchema.Environment.required,
  properties: {
    ...internalSchema.Environment.properties,
    ...internalSchema.Environment.allOf[0].properties,
    ...internalSchema.Environment.allOf[1].properties,
  },
};

File: packages/api/portfolios/environments/createEnvironment.ts (hunk 5 above)
Comment (juliusfitzhugh-ccpo) [reply packages/api/portfolios/environments/createEnvironment.ts:41]: This was intended to throw the error to the middleware. It's similar to the `validateRequestShape` function.

File: packages/api/portfolios/environments/createEnvironment.ts (hunk 7 above)
Comment (juliusfitzhugh-ccpo) [reply packages/api/portfolios/environments/createEnvironment.ts:135]: yeah, the same issue came up later after removing the `additionalProperties` that was in `BaseObject`.  I reverted the `BaseObject` and went with this approach. I tried the approach you suggested with the `metadata` property in the `atat-internal-api.yml`, but was not successful (not sure I did it right, though).

File: packages/api/portfolios/environments/createEnvironment.ts (hunk 5 above)
Comment (juliusfitzhugh-ccpo) [reply packages/api/portfolios/environments/createEnvironment.ts]: Updated to remove the `catch` block and all errors handled by `errorHandlingMiddleware`. it's not pretty, but it works.

File: packages/api/portfolios/environments/createEnvironment.ts [hunk 8]
```diff
@@ -0,0 +1,109 @@
...
+ *
+ * @param event - The POST request from API Gateway
+ */
+export async function baseHandler(
+  event: ApiGatewayEventParsed<IEnvironment>,
+  context?: Context
+): Promise<APIGatewayProxyResult> {
+  const setupResult = validateRequestShape<IEnvironment>(event);
+  const { portfolioId, applicationId } = setupResult.path;
+  const environmentBody = setupResult.bodyObject;
```
zachclark-ccpo, line 35: You don't need to set this, `.use(jsonBodyParser())` parses the event.body before anything inside the baseHandler actually runs

juliusfitzhugh-ccpo: I'm not sure I completely follow, should I just use `event.body`? Also, do we not want to use what is coming from `validateRequestShape`?

packages/api/portfolios/environments/createEnvironment.ts
juliusfitzhugh-ccpo: I've updated the `atat-internal-api.yml` to have this structure, and this was removed. It causes duplication with `BaseObject` and `AppEnvAccess`, but ensures the schema does not have to be restructured here.

packages/api/portfolios/environments/createEnvironment.ts
```diff
@@ -0,0 +1,109 @@
...
+    // ensure the environment name is unique for the application
+    // according to business rule 3.3
+    for (const environment of environments) {
+      if (environment.name === environmentBody.name) {
+        throw createError(400, "Duplicate Environment name in application", {
+          errorName: "DuplicateEnvironmentName",
+          environmentName: environmentBody.name,
+        });
+      }
+    }
```
zachclark-ccpo, line 67: We should pull this into Environment repository, we need this for the UPDATE and we need it for Application operations.

packages/api/portfolios/environments/createEnvironment.ts
```diff
@@ -0,0 +1,109 @@
...
+          "archivedAt",
+          "administrators",
+          "contributors",
+          "readOnlyOperators",
+        ],
+        where: { id: environment.id },
+      });
+
+      insertedEnvironments.push(queryInsertedEnvironment);
+      console.log("Inserted Environment: " + JSON.stringify(queryInsertedEnvironment));
```
zachclark-ccpo, line 93: Instead of inserting, and then doing a query, you could try to do a `.save` and then wrapping the response to only the fields you specified in your select

packages/api/portfolios/environments/createEnvironment.ts
```diff
@@ -49,11 +49,8 @@ export async function baseHandler(
 
     // get all environment names for application
     const environments = await connection
-      .getRepository(Environment)
-      .createQueryBuilder("environment")
-      .select(["environment.name"])
-      .where("environment.applicationId = :applicationId", { applicationId: application.id })
-      .getMany();
+      .getCustomRepository(EnvironmentRepository)
+      .getAllEnvironmentNames(application.id);
```
jasonburkert-ccpo, line 53: Good change

packages/api/portfolios/environments/createEnvironment.ts
juliusfitzhugh-ccpo: Yeah I think I will try that in a separate refactor PR to keep the moment going.

packages/api/portfolios/environments/createEnvironment.ts
juliusfitzhugh-ccpo: Added a method to get all the env names in the Environment repository to assist with that. You can maybe do an .include() on names to see if the name is taken to simplify the code.

jasonburkert-ccpo (approved):
Added some comments to discussion, but I don't see anything requiring changes.

packages/api/portfolios/environments/createEnvironment.ts
juliusfitzhugh-ccpo: Understood from our conversation earlier. I left as-is to avoid typescript issues.

packages/api/portfolios/environments/createEnvironment.test.ts
```diff
@@ -0,0 +1,73 @@
...
+import { IEnvironmentCreate } from "../../../orm/entity/Environment";
+import { ErrorStatusCode, SuccessStatusCode } from "../../utils/response";
+
+describe("createEnvironment", () => {
+  it.skip("successful operation testing locally only", async () => {
+    const validRequest: ApiGatewayEventParsed<IEnvironmentCreate> = {
+      body: { name: "new local testing and a unique name" },
+      pathParameters: {
+        portfolioId: "91c6bd1f-5ed8-413b-8f85-55a62dd50ad3",
+        applicationId: "7bc938ca-4c1c-4740-8ccf-18c940a70862",
```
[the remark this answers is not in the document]
juliusfitzhugh-ccpo: I would but those are hardcoded values that I use on my local db.

packages/api/portfolios/environments/createEnvironment.ts
```diff
@@ -0,0 +1,106 @@
...
+    // get all environment names for application
+    const environments = await connection
+      .getCustomRepository(EnvironmentRepository)
+      .getAllEnvironmentNames(application.id);
+
+    // ensure the environment name is unique for the application
+    // according to business rule 3.3
+    for (const environment of environments) {
+      if (environment.name === environmentBody.name) {
+        throw createError(400, "Duplicate Environment name in application", {
```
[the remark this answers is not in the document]
juliusfitzhugh-ccpo: Oh I didn’t notice. I’ll change to lower case.

packages/api/portfolios/environments/createEnvironment.ts
```diff
@@ -0,0 +1,106 @@
+import "reflect-metadata";
```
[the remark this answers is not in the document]
juliusfitzhugh-ccpo: Yeah that was my thought. It came up in enough errors so I added it to make sure.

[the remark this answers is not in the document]
jasonburkert-ccpo: Saw that suggested when troubleshooting the entity meta data not found errors, and just want to be explicit that that's what we're doing

packages/api/portfolios/environments/createEnvironment.test.ts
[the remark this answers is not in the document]
jasonburkert-ccpo: ok fair

atat_internal_api.yaml
```diff
@@ -1912,13 +1918,43 @@ components:
...
-      allOf:
-        - $ref: "#/components/schemas/BaseObject"
-        - $ref: '#/components/schemas/AppEnvAccess'
       properties:
         name:
           pattern: "^[a-zA-Z\\d _-]{1,100}$"
           type: string
+      # Remove the addition of 'allOf' while still using 'additionalProperties'
+      # which adds duplication to the spec and is not ideal, but allows for
+      # validation to occur for the middy validator
```
zachclark-ccpo, line 1927: This is an ajv specific issue: https://ajv.js.org/faq.html#additional-properties-inside-compound-keywords-anyof-oneof-etc @jeffsegal-ccpo @jasonburkert-ccpo

zachclark-ccpo: Tried to resolve this using various open api patterns, but still receiving errors.

juliusfitzhugh-ccpo: Thanks for pointing to the ajv link for the source. Here is another link (you pointed me to previously) that is kind of related as well. 
- [Schema Composition with allOf](https://json-schema.org/understanding-json-schema/reference/combining.html#properties-of-schema-composition)

jasonburkert-ccpo: If we want the increased security of `additionalProperties: false`, then compound keyword `allOf` must be removed.  This makes us repeat ourselves in the API spec, but will allow validation via [middy validator](https://middy.js.org/packages/validator/) which has a dependency on ajv.

PR #504, opened 2021-12-02
Refactor models and entities to prevent name collisions

juliusfitzhugh-ccpo:
Refactor of models and entities in the `api/package` to prevent name collisions and add clarity. The only changes done are the name of the models or entities, and the locations they are being used (a lot of the files in `api/packages`). For example `Application` becomes `EnvironmentModel` for the model and `EnvironmentEntity` for the typeORM entity. 

Since a large amount of files were updated, a deployed stack was tested to ensure all the original routes worked as expected and nothing was found.

### DIFF
--- a/packages/api/models/Error.ts
+++ b/packages/api/models/Error.ts
@@ -3,12 +3,12 @@ export enum ErrorCode {
   OTHER = "OTHER",
 }
 
-export interface Error {
+export interface ErrorModel {
   code: ErrorCode;
   message: string;
 }
 
-export interface ValidationError extends Error {
+export interface ValidationErrorModel extends ErrorModel {
   code: ErrorCode.INVALID_INPUT;
   error_map: Record<string, unknown>;
 }
--- a/packages/api/models/Application.ts
+++ b/packages/api/models/Application.ts
@@ -1,8 +1,8 @@
-import { AppEnvOperator } from "./Operator";
-import { Environment } from "./Environment";
-export interface Application {
+import { AppEnvOperatorModel } from "./Operator";
+import { EnvironmentModel } from "./Environment";
+export interface ApplicationModel {
   name: string;
   description?: string;
-  environments: Array<Environment>;
-  operators: Array<AppEnvOperator>;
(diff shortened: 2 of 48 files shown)


jasonburkert-ccpo:
Thanks for opening this PR!  This renaming of the exported interfaces and types and classes from the TypeScript files does eliminate the ambiguity of code like const clins: Array<Clin>;

**Team question**: Should the file names also be changed?  Does `models/Clin.ts` become `models/ClinModel.ts`, and `orm/entity/Clin.ts` become `orm/entity/ClinEntity.ts` ?  How about `models/Operator.ts`?

Also, I think we may need to specify table names in the `@Entity` decorators, which is probably a good idea anyway.

jasonburkert-ccpo:
> Also, I think we may need to specify table names in the `@Entity` decorators, which is probably a good idea anyway.
I've included this in #489

packages/api/models/Error.ts
```diff
@@ -3,12 +3,12 @@ export enum ErrorCode {
...
+export interface ErrorModel {
   code: ErrorCode;
   message: string;
 }
 
-export interface ValidationError extends Error {
+export interface ValidationErrorModel extends ErrorModel {
   code: ErrorCode.INVALID_INPUT;
   error_map: Record<string, unknown>;
 }
```
zachclark-ccpo, line 14: Why do we need to change the name of this to include Model? It shouldn't conflict with an `orm/Entity` because there is no entity for Error.

kylelaker-ccpo:
Why do we need to totally rename the types? Why not just `import { X as XModel }` in the cases where there would be a conflict?

kylelaker-ccpo:
We can also do the `I<Type>` convention for the interfaces if we want

jasonburkert-ccpo:
Closing PR following team discussion.

packages/api/models/Error.ts
juliusfitzhugh-ccpo: I agree and thought the same. However, I wasn't entirely sure were to draw the line, so I included it since I changed most of the others.

PR #507, opened 2021-12-03
Run migrations during bootstrapping

kylelaker-ccpo:
#### Modified RDS initial-bootstrap
- specified entity and migration dirs in typeorm connection
- added function `migrate` to run all typeorm migrations on database create or update

#### Modified `DatabaseBootstrapper` construct
- added new lambda layer `DatabaseSupportLayer`
- added new script `prepare-db-layer.sh` to transpile typeorm entities and migrations and download the RDS CA bundle
- new script is executed by the new layer

#### Modified `ApiFlexFunction` construct
- removed afterBundling command hook that used to download the RDS CA bundle

I have no idea if this works as-is.


infrastructure/lib/constructs/database.ts
```diff
@@ -36,12 +36,24 @@ class DatabaseBootstrapper extends cdk.Construct {
   public readonly backingLambda: lambda.IFunction;
   constructor(scope: cdk.Construct, id: string, props: BootstrapProps) {
     super(scope, id);
+
+    const layer = new lambda.LayerVersion(this, "DatabaseSupportLayer", {
+      compatibleRuntimes: [lambda.Runtime.NODEJS_14_X],
```
jeffsegal-ccpo, line 41: Is this an alternative to specify an exact image where the bundling will be executed?

packages/rds/initial-bootstrap.ts
```diff
@@ -95,6 +98,13 @@ async function handleDelete(connectConfig: Config): Promise<void> {
   await connection.query(`DROP DATABASE IF EXISTS ${connectConfig.databaseName} WITH FORCE`);
 }
 
+async function migrate(connectConfig: Config): Promise<void> {
+  const connection = await connect(connectConfig);
+  const migrations = await connection.runMigrations({ transaction: "none" });
+  await connection.close();
+  migrations.forEach((migration) => console.log(migration));
```
jeffsegal-ccpo, line 105: Looks like this method should only return migrations that we actually ran, although TypeORM also uses its own internal logger to log each one as well.

infrastructure/lib/constructs/database.ts
kylelaker-ccpo: This just decreases the value from "all runtimes" https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-lambda.LayerVersion.html#compatibleruntimes. It doesn't _really_ matter. But I didn't want to add things we don't know work.

packages/rds/initial-bootstrap.ts
kylelaker-ccpo: Yeah, my plan was to remove this before going non-draft.

jasonburkert-ccpo (commented):
👍

scripts/prepare-db-layer.sh
```diff
@@ -0,0 +1,29 @@
...
+export ASSET_INPUT_DIR="/asset-input"
+export ASSET_OUTPUT_DIR="/asset-output"
+
+## Transpile entities and migrations for TypeORM 
+prepare-orm-files() (
+    npm ci \
+        && npm run lerna bootstrap -- --ci \
+        && npm run build \
+        && cd "$ASSET_INPUT_DIR/build/packages/api" \
+        && cp --parents -aur orm/**/**.js "$ASSET_OUTPUT_DIR" \
```
jasonburkert-ccpo, line 19: I understand the first double asterisk globbing operator.  What is the significance of the second?  Does it do something different than `*.js` ?

jeffsegal-ccpo (approved):

scripts/prepare-db-layer.sh
kylelaker-ccpo: Honestly, not sure. `orm/**.js` didn't catch the things under subdirectories. So `orm/**` should catch `entities` and `migrations` and then `orm/**/**.js` should catch all the JS files within each of those directories. Any suggestions on cleaning up this glob? I was just sorta going with the thing that worked first here :D

packages/orm/entity/ProvisionableEntity.ts
```diff
@@ -1,14 +1,20 @@
 import { BaseEntity } from "./BaseEntity";
 import { Column } from "typeorm";
-import { ProvisioningStatus } from "../../models/ProvisioningStatus";
+
+export enum ProvisioningStatus {
+  PENDING = "pending",
+  IN_PROGRESS = "in_progress",
+  FAILED = "failed",
+  COMPLETE = "complete",
```
jasonburkert-ccpo, line 8: The values are uppercase now per spec.   Ordered like spec.
```suggestion
  PENDING = "PENDING",
  IN_PROGRESS = "IN_PROGRESS",
  COMPLETE = "COMPLETE",
  FAILED = "FAILED",
```

packages/orm/entity/TaskOrder.ts
```diff
@@ -1,9 +1,14 @@
 import { BaseEntity } from "./BaseEntity";
 import { Clin } from "./Clin";
 import { Column, Entity, ManyToOne, OneToMany } from "typeorm";
-import { FileScanStatus } from "../../models/FileMetadata";
 import { Portfolio } from "./Portfolio";
 
+export enum FileScanStatus {
+  PENDING = "pending",
+  ACCEPTED = "accepted",
+  REJECTED = "rejected",
```
jasonburkert-ccpo, line 9:
```suggestion
  PENDING = "PENDING",
  ACCEPTED = "ACCEPTED",
  REJECTED = "REJECTED",
```

scripts/prepare-db-layer.sh
jasonburkert-ccpo: Nah, this is good.  I was just curious.

juliusfitzhugh-ccpo (approved):
This was a quick turn around thank you, and nice job! Nothing stood out to me. There were a lot of good things in this PR to bring everything together. I like the orm package helping to separate the switch from the old to new internal spec.

zachclark-ccpo (approved):

PR #516, opened 2021-12-07
Add database test route

kylelaker-ccpo:
We previously used `POST /portfolioDrafts` for this but that caused
problems because it was used by the UI directly so if there were issues
with talking to the database, we took down the app. This lets us test
them somewhere outside the actual app code.


zachclark-ccpo (approved):

jeffsegal-ccpo (approved):
If we think this will come in handy and it's not duplicative with the other real API methods that are hopefully about to be merged, then I'm all for this.

PR #517, opened 2021-12-07
Added  CspInvocation and CspResponse objects

jeffsegal-ccpo:
The CSP Provisioning process will be executed roughly as follows:

1. User interacts with ATAT UI / Internal API to update ATAT resources (Portfolios, Applications, Environments & Task Orders)
2. When ready, User submits all pending resources for provisioning. Pending resources are defined as new rows in the database whose status is not `COMPLETE` or `IN_PROGRESS`
3. The POST /portfolios/:idsubmit Lambda walks the tree of the given Portfolio and creates one or more `CspInvocation` objects and passing each to their own instance of a Step Function workflow
4. The  CSP Invocation Lambda reads the given `CspInvocation` object and makes a matching invocation to the endpoint contained within it (initial implementation in Phase 1 will be totally mocked out, but include ability to test all paths)
    1.  If a 5xx code is received, retry up to 6 times
    2. If a 4xx code is received, invoke the Reject CSP Response Lambda. This will mark each resource under the parent with `provisioning_status` = `FAILED`.
    3. If a 2xx code is received, invoke the Persist CSP Response Lambda. This will mark each resource under the parent with `provisioning_status` = `COMPLETE`.

We'll need a proper formal schema In order to properly exchange information between these Lambdas. It should also facilitate parallel development of these Lambdas. If any other objects need to be created, they can be added here later. In theory we could add them to a different spec file but this seems fine for now. 

Ticket: AT-6707 (CSP Provisioning Epic)

### DIFF
--- a/atat_internal_api.yaml
+++ b/atat_internal_api.yaml
@@ -2038,6 +2038,37 @@ components:
   #          type: string
   #        lastName:
   #          type: string
+    CspInvocation:
+      required:
+        - method
+        - endpoint
+        - payload
+      properties:
+        method:
+          type: string
+          enum:
+            - GET
+            - POST
+            - PUT
+            - DELETE
+        headers:
+          type: object
+          additionalProperties:
+            type: string
+        endpoint:
+          type: string
+          format: uri
+        payload:
+          type: object
+    CspResponse:
+      properties:
+        code:
+          type: integer
+        content:
+          type: object
+            additionalProperties: true
+        request:
+          $ref: '#/components/schemas/CspInvocation'
   examples:
     PortfolioSummaries:
       value:
--- a/atat_internal_api.yaml
+++ b/atat_internal_api.yaml
@@ -2066,7 +2066,7 @@ components:
           type: integer
         content:
           type: object
-            additionalProperties: true
+          additionalProperties: true
         request:
           $ref: '#/components/schemas/CspInvocation'
   examples:


atat_internal_api.yaml
```diff
@@ -2038,6 +2038,37 @@ components:
...
+          format: uri
+        payload:
+          type: object
+    CspResponse:
+      properties:
+        code:
+          type: integer
+        content:
+          type: object
+            additionalProperties: true
```
zachclark-ccpo, line 2069: `additionalProperties: true` is true by default, could probably get away with removing it.

atat_internal_api.yaml
```diff
@@ -2038,6 +2038,37 @@ components:
...
+        method:
+          type: string
+          enum:
+            - GET
+            - POST
+            - PUT
+            - DELETE
+        headers:
+          type: object
+          additionalProperties:
```
zachclark-ccpo, line 2056: I was a bit confused seeing additionalProperties used this, but it checks out.
Link to docs if anyone is interested: https://swagger.io/docs/specification/data-models/dictionaries/

atat_internal_api.yaml
```diff
@@ -2038,6 +2038,37 @@ components:
   #          type: string
   #        lastName:
   #          type: string
+    CspInvocation:
+      required:
+        - method
+        - endpoint
+        - payload
```
zachclark-ccpo, line 2045: Are `headers` not required?

atat_internal_api.yaml
jeffsegal-ccpo: Yeah - I believe this is the best practice for defining an object as a `string` -> `string` map (i.e. what we would want for HTTP Headers).

atat_internal_api.yaml
jeffsegal-ccpo: Nope - I didn't want to impose anything on the process but in theory we might need to pass some headers in some day that might require some dynamic calculation or other reason for including in this object (vs just hard-coding it in the invocation Lambda).

zachclark-ccpo (approved):

atat_internal_api.yaml
jasonburkert-ccpo: Thanks for linking those docs

atat_internal_api.yaml
zachclark-ccpo, line 2069:
```suggestion
        content:
          type: object
          additionalProperties: true
```

zachclark-ccpo: Nested mapping fix. from @jasonburkert-ccpo  slack comment

jasonburkert-ccpo (approved):
Looks good after

PR #518, opened 2021-12-08
Implement createPortfolio operation

jasonburkert-ccpo:
- Adds operation `createPortfolio` to the api stack
- Adds unit tests (some skipped for local DB testing) for `createPortfolio`
- Adds typeorm repository `PortfolioRepository`
- Sets ARN for CreatePortfolioFunction in API spec file
- Changes `validateRequestShape()` to allow optional checking of path parameters.  Requests to this operation shouldn't have any.
- Adds unit test to ensure requests that include audit properties `id`, `createdAt`, `updatedAt`, `archivedAt` fail with status 400 BAD REQUEST.  This ensures that those properties won't be overwritten even if implementation should change
- Adds `EmptyPortfolioMiddleware` to handle empty string request body that creates an empty portfolio using default values
- Adds migration to make column `portfolio.owner` nullable so it is not required to be included in requests
- Implements business rule to enforce unique portfolio names across all portfolios.  Duplicate names result in 400 BAD REQUEST and an appropriate message.
- Fix: Corrects two ARNs in old API spec that should have been changed in #557

Ticket: AT-6878


packages/orm/entity/TaskOrder.ts
```diff
@@ -20,10 +21,10 @@ export class TaskOrder extends BaseEntity {
   @Column({ type: "uuid", comment: "S3 object key of task order pdf" })
   fileId: string;
 
-  @Column({ length: 256, comment: "name of pdf file when uploaded" })
+  @Column({ type: "varchar", length: 256, comment: "name of pdf file when uploaded" })
   fileName: string;
 
-  @Column({ nullable: true, comment: "pdf file size in bytes" })
+  @Column({ type: "integer", nullable: true, comment: "pdf file size in bytes" })
```
kylelaker-ccpo, line 27: Is this informing TypeORM of the DB structure? Or do we need to write a migration for this as well?

packages/api/utils/database.ts
```diff
@@ -80,7 +85,7 @@ export async function createConnection(): Promise<Connection> {
     },
     logging: "all",
     database: databaseName,
-    entities: ["/opt/nodejs/orm/entity/**.js"],
+    entities: [Portfolio, Application, Environment, TaskOrder, Clin, "/opt/nodejs/orm/entity/**.js"],
     migrations: ["/opt/nodejs/orm/migration/**.js"],
```
kylelaker-ccpo, line 89: Wait a second... has this seriously been an option the whole time? To just put the types directly in this array?

infrastructure/lib/atat-web-api-stack.ts
```diff
@@ -537,6 +540,9 @@ export class AtatWebApiStack extends cdk.Stack {
       handlerPath: this.determineApiHandlerPath(operationId, handlerFolder),
       database: this.database,
       ormLayer: this.ormLayer,
+      functionPropsOverride: {
+        timeout: cdk.Duration.seconds(10),
+      },
```
kylelaker-ccpo, line 545: Fair but we do need to debug this. We can't actually support 10s for each query; the UI will crawl

kylelaker-ccpo (commented):
Some early thoughts

infrastructure/lib/atat-web-api-stack.ts
jasonburkert-ccpo: Usually I see it taking about 6 seconds initially.

packages/api/utils/database.ts
jasonburkert-ccpo: Yep.  See the examples for _entities_ at [Common connection options](https://typeorm.io/#/connection-options/common-connection-options).  I struggled with EntityMetadataNotFoundError until I did this.

packages/orm/entity/TaskOrder.ts
jasonburkert-ccpo: Good question.  Had been encountering `ColumnTypeUndefinedError` until adding these.  This is manually specifying the column type to TypeORM.  Attempts to get column type automatically must have been failing in these cases.

You had me concerned that we might need a migration, however, I checked each column changed against the migration file and found the expected type for each within the CREATE TABLE statements.  So I think we're ok.

infrastructure/lib/atat-web-api-stack.ts
jeffsegal-ccpo: Well, we need the timeout to be big enough for a cold start right? So technically the first query could be unacceptably slow as long as subsequent ones are fine. Granted, we'll have to pursue a warming strategy if this is the case.

packages/api/portfolios/createPortfolio.ts
```diff
@@ -0,0 +1,86 @@
...
+  context?: Context
+): Promise<APIGatewayProxyResult> {
+  validateRequestShape<IPortfolio>(event);
+
+  // set up database connection
+  const connection = await createConnection();
+  const insertedPortfolios: IPortfolioCreate[] = [];
+
+  try {
+    const setupResult = validateRequestShape<IPortfolio>(event);
```
zachclark-ccpo, line 39: Doing this twice? see line 32

packages/api/portfolios/createPortfolio.ts
```diff
@@ -0,0 +1,86 @@
...
+          "archivedAt",
+          "administrators",
+          "provisioningStatus",
+        ],
+        where: { id: portfolio.id },
+      });
+
+      insertedPortfolios.push(queryInsertedPortfolio);
+      console.log("Inserted Portfolio: " + JSON.stringify(queryInsertedPortfolio));
+    }
```
zachclark-ccpo, line 68: This for loop executes one time, because you are only inserting a single portfolio, right?

packages/api/portfolios/createPortfolio.ts
```diff
@@ -0,0 +1,86 @@
...
+  } catch (error) {
+    console.error("Database error: " + error);
+    return DATABASE_ERROR;
```
juliusfitzhugh-ccpo, line 71: This can be removed since this error is caught by the switch `default` case in the `errorHandlingMiddleware`. I did that in commit https://github.com/dod-ccpo/atat-web-api/pull/500/commits/3fa3537e7856684d610ef247b43f563c4fbbcf56 since there was no default and this seemed like a good default to catch our unknown errors.

packages/api/repository/PortfolioRepository.ts
```diff
@@ -0,0 +1,55 @@
...
+      where: { id },
+    });
+  }
+
+  // GET all portfolios
+  getPortfolios(): Promise<[Array<Portfolio>, number]> {
+    return this.createQueryBuilder("portfolio")
+      .select(["portfolio.name", "portfolio.id", "portfolio.createdAt", "portfolio.updatedAt", "portfolio.archivedAt"])
+      .getManyAndCount();
+  }
```
juliusfitzhugh-ccpo, line 31: I ended up changing this to `.getMany()` because `.getManyAndCount()` seemed to be unnecessary and dealing with the return structure seemed bulky (see changes in `EnvironmentRepositroy` on commit https://github.com/dod-ccpo/atat-web-api/pull/537/commits/0afbda327ab96b1a102cad2ec3f6dcbb076ac8f9). If an array is returned we can simply use the `.length` property to get the count. 

Both can be used, but I thought this could open a small discussion conversation as if we want `.getMany()` or `.getManyAndCount()`.

packages/api/portfolios/createPortfolio.ts
jasonburkert-ccpo: Good catch. Fixed in 2b24834ba61dbf1e311de8432f4c80ed09cf7337

packages/api/portfolios/createPortfolio.ts
jasonburkert-ccpo: Yep.  Removed in 2b24834ba61dbf1e311de8432f4c80ed09cf7337

packages/api/repository/PortfolioRepository.ts
jasonburkert-ccpo: That change and explanation makes sense.  Changed in 34db71956811264f3ed5daa2c1e79dccf7d2bebf

packages/api/portfolios/createPortfolio.ts
jasonburkert-ccpo: Removed in 137037d0e45e3765b608ea4642285f17ceab04bd

packages/api/portfolios/createPortfolio.test.ts
```diff
@@ -0,0 +1,29 @@
...
+        owner: "jane.manager@example.mil",
+        csp: CloudServiceProvider.CSP_A,
+        dodComponents: [DodComponent.ARMY, DodComponent.NAVY],
+        portfolioManagers: ["joe.manager@example.mil", "jane.manager@example.mil"],
+        description: "Mock portfolio description",
+        provisioningStatus: ProvisioningStatus.PENDING,
+      },
+      pathParameters: {
+        portfolioId: uuidv4(), // this is required by validateRequestShape, but discarded.  id generated by DB.
+      },
```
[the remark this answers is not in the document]
jasonburkert-ccpo: Made this exact change and unit tested.  Very nice suggestion!  Changed in 286517cd41860f4622c0ddf1043eadf3b2ff5ef1

infrastructure/lib/atat-web-api-stack.ts
```diff
@@ -545,7 +545,7 @@ export class AtatWebApiStack extends cdk.Stack {
       database: this.database,
       ormLayer: this.ormLayer,
       functionPropsOverride: {
-        timeout: Duration.seconds(10),
+        timeout: cdk.Duration.seconds(10),
```
jasonburkert-ccpo, line 548: This makes it similar to other uses of same

juliusfitzhugh-ccpo: I thought at one point we were going the other way, but fair enough the others are also using `cdk.XX` too, so better to stay consistent.

packages/api/portfolios/createPortfolio.test.ts
```diff
@@ -0,0 +1,38 @@
...
+    const result = await handler(validRequest, {} as Context, () => null);
+    expect(result?.statusCode).toBe(SuccessStatusCode.CREATED);
+  });
+  it("requests with audit properties should cause 400 BAD REQUEST", async () => {
+    const validRequest: ApiGatewayEventParsed<IPortfolioCreate> = {
+      body: {
+        id: "",
+        createdAt: "",
+        updatedAt: "",
+        archivedAt: "",
```
juliusfitzhugh-ccpo, line 31: Should dummy values be in these?

packages/api/portfolios/createPortfolio.ts
```diff
@@ -0,0 +1,88 @@
...
+
+const portfolioSchema = {
+  additionalProperties: schema.Portfolio.additionalProperties,
+  type: schema.Portfolio.type,
+  required: schema.Portfolio.required,
+  properties: {
+    ...schema.PortfolioBase.properties,
+    ...schema.Portfolio.properties,
+    ...schema.PortfolioSummary.properties,
+    ...schema.PortfolioAccess.properties,
```
juliusfitzhugh-ccpo, line 77: While this doesn't have the `BaseObject` properties (e.g., createdAt), should we allow other properties to pass by the validator like properties related to applications, funding and possibly other things by spreading everything? 

The spec only allows the request body to have `csp`, `dodComponents`, `name`, `portfolioAdministrators`, `portfolioManagers`, and `description`. I don't think the `schema.PortfolioSummary.properties` would be needed in that case and the `schema.Portfolio.properties.applications` might not be wanted either. I might suggest to align with the spec a little closer.

```suggestion
    ...schema.PortfolioBase.properties,
    ...schema.PortfolioAccess.properties,
    description: schema.Portfolio.properties.description,
    owner: schema.Portfolio.properties.owner,
    portfolioManagers: schema.Portfolio.properties.portfolioManagers,
```

packages/api/repository/PortfolioRepository.ts
```diff
@@ -0,0 +1,55 @@
...
+  // GET all portfolios
+  async getPortfolios(): Promise<Array<Portfolio>> {
+    return await this.createQueryBuilder("portfolio")
+      .select(["portfolio.name", "portfolio.id", "portfolio.createdAt", "portfolio.updatedAt", "portfolio.archivedAt"])
+      .getMany();
+  }
+
+  // POST create new portfolio
+  async createPortfolio(portfolio: IPortfolio): Promise<InsertResult> {
+    return await this.insert(portfolio);
```
juliusfitzhugh-ccpo, line 35: I know you mentioned about seeing the updates display old data previously. Did you see a difference once the `async/await` was added? 

I know previously I added `async/await` on all the methods, but sonar cloud complained about a lot of them.

juliusfitzhugh-ccpo (commented):
I added a few comments, but things are looking good! I'm ready to see all the `portfolio`, `application` and `environment` endpoints work together. I was trying to run tests against your deployed stack, but I was getting `"message": "Forbidden"`  status `403`. Was the stack updated, or was I just doing something wrong?

packages/api/portfolios/createPortfolio.test.ts
jasonburkert-ccpo: I think their mere presence should trigger the 400.  Being an unexpected additional property shouldn't depend on the value of the property.

The test could be improved to iterate over the collection of disallowed properties.  This will trip on the first one, and not evaluate them all.

packages/api/portfolios/createPortfolio.ts
jasonburkert-ccpo: Thanks for the critical look.  Will revisit this, especially seeing Zach's extensive example.

packages/api/portfolios/createPortfolio.test.ts
juliusfitzhugh-ccpo: Yeah, that's true. 

You should be okay without the iteration since the validator checks and gets all the `additionalProperties` that should not be in the request (I did the `it.each()` then remembered the validator capture all invalid properties). It would look something like this.
```typescript
{
  "code": "INVALID_INPUT",
  "message": "Request failed validation",
  "error_map": [
    {
      "instancePath": "/body",
      "schemaPath": "#/properties/body/additionalProperties",
      "keyword": "additionalProperties",
      "params": {
        "additionalProperty": "id"
      },
      "message": "must NOT have additional properties"
    },
    {
      "instancePath": "/body",
      "schemaPath": "#/properties/body/additionalProperties",
      "keyword": "additionalProperties",
      "params": {
        "additionalProperty": "createdAt"
      },
      "message": "must NOT have additional properties"
    },
    {
      "instancePath": "/body",
      "schemaPath": "#/properties/body/additionalProperties",
      "keyword": "additionalProperties",
      "params": {
        "additionalProperty": "updatedAt"
      },
      "message": "must NOT have additional properties"
    },
    {
      "instancePath": "/body",
      "schemaPath": "#/properties/body/additionalProperties",
      "keyword": "additionalProperties",
      "params": {
        "additionalProperty": "archivedAt"
      },
      "message": "must NOT have additional properties"
    }
  ]
}
```

packages/orm/entity/Portfolio.ts
```diff
@@ -20,13 +20,13 @@ export enum DodComponent {
...
 export interface IPortfolio {
   name: string;
-  description?: string;
+  administrators?: Array<string>;
 }
 
-export interface PortfolioUpdate {
-  name: string;
+export interface IPortfolioCreate extends IPortfolio {
   description: string;
```
juliusfitzhugh-ccpo, line 30: I saw you had `description` as optional in `IPortfolio` previously, should that have carried over to this `description`?

packages/api/utils/emptyPortfolioMiddleware.ts
```diff
@@ -0,0 +1,40 @@
+import middy from "@middy/core";
+import { APIGatewayProxyEvent, APIGatewayProxyResult } from "aws-lambda";
+import { CloudServiceProvider } from "../models/CloudServiceProvider";
+
+function isRequestBodyEmptyString(body: string | null): boolean {
+  return body?.length === 0;
+}
```
juliusfitzhugh-ccpo, line 7: Do we want to cover an empty JSON body as well (e.g., `"{}"`)? In previous routes, an empty request body would be sent such as`{}`, so I was wondering if we wanted to follow that pattern here or simply send an empty string.

packages/api/utils/emptyPortfolioMiddleware.ts
```diff
@@ -0,0 +1,40 @@
...
+  return JSON.parse(JSON.stringify(emptyPortfolio));
+}
+
+/**
+ * Middleware helper for createPortfolio operation
+ * API spec allows empty Portfolio to be created when empty string request body is recieved.
+ * This handles that case.  When request body is an empty or populated JSON object that is handled elsewhere.
+ * @returns
+ */
+export const EmptyPortfolioMiddleware = (): middy.MiddlewareObj<APIGatewayProxyEvent, APIGatewayProxyResult> => {
```
juliusfitzhugh-ccpo, line 29: Nice use of middy to create a default portfolio with middleware!

packages/orm/entity/Portfolio.ts
jasonburkert-ccpo: Applied in a8f122e87e60c2d0aa1e8b3e00e1a38c57e7e2c5

packages/api/utils/emptyPortfolioMiddleware.ts
jasonburkert-ccpo: The API spec lists two individual content/content types for request bodies, one for JSON, and one for plain text.  An empty JSON body will get caught by the validator and return 400.

I've included examples of the empty string request and a full portfolio JSON request in the ticket.

```
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/Portfolio'
          text/plain:
            schema:
              type: string
              example: ''
```

packages/api/portfolios/createPortfolio.ts
```diff
@@ -0,0 +1,102 @@
...
+        "csp",
+        "description",
+        "dodComponents",
+        "owner",
+        "portfolioManagers",
+        "createdAt",
+        "updatedAt",
+        "archivedAt",
+        "administrators",
+        "provisioningStatus",
```
juliusfitzhugh-ccpo, line 57: According to the spec, it does not seem the `provisioningStatus` is returned. This might need to be removed, unless I missed something.
```suggestion
```

packages/api/portfolios/createPortfolio.ts
jasonburkert-ccpo: This was improved in 620f1e7468095690dabfd867b7970dedc240456d

packages/api/repository/PortfolioRepository.ts
jasonburkert-ccpo: I've not looked at this since then.  Will check sonar cloud and follow guidance there.

packages/api/portfolios/createPortfolio.ts
jasonburkert-ccpo: Agreed.  That property is from entity `PortfolioSummary`.  Will remove.

jasonburkert-ccpo: Removed in 4e7dccc903692b313f201aae7b6c84fa654f3b8e

packages/api/utils/shapeValidator.ts
```diff
@@ -18,7 +18,10 @@ export function validateRequestShape<T>(
   event: ApiGatewayEventParsed<T>,
   ...extraValidators: Array<(obj: unknown) => obj is T>
 ): SetupSuccess<T> {
-  if (!Object.values(event.pathParameters as APIGatewayProxyEventPathParameters).every(isValidUuidV4)) {
+  if (
+    event.pathParameters &&
+    !Object.values(event.pathParameters as APIGatewayProxyEventPathParameters).every(isValidUuidV4)
+  ) {
```
zachclark-ccpo, line 24: Why change this?
Have you tested if this breaks the createApplication and createEnvironment requests?

zachclark-ccpo (changes requested):
Where is the business rules validation for "This portfolio name already exists. Please try another name." (Rule 1.1)?
Check out `createApplication.ts` for an example of how to validate this:
```
   // Business rules validation
    // 3.1 unique applicaiton name
    // Search for all applications in the portfolio where the name matches the incoming name
    const existingApplications = await connection
      .getCustomRepository(ApplicationRepository)
      .getAllMatchingApplicationNames(portfolioId, applicationBody.name);
    console.log("existing applications below");
    console.log(existingApplications);

    // existing applications is either an empty array or contains a matching application.name
    // if it contains any matches, throw an error
    if (existingApplications.length > 0) {
      throw createError(400, "Duplicate application name in portfolio", {
        errorName: "DuplicateApplicationName",
        applicationName: applicationBody.name,
      });
    }
```

juliusfitzhugh-ccpo (dismissed):
Nice job! It's good to see a migration besides the initial one work. Tested in a deployed stack and the expected behavior was observed for `createPortfolio`.

packages/api/utils/shapeValidator.ts
zachclark-ccpo:
```
if(event.pathParameters)
```
Doesn't need to be included because
```
    x-amazon-apigateway-request-validator: "request-params-validator"
```
Catches empty path parameters on the API gateway level.

packages/api/repository/PortfolioRepository.ts
jasonburkert-ccpo: Corrected smells in 8b9aca2fa37b114313e9bd7785ffbd4cfbe2eaa4

packages/api/utils/shapeValidator.ts
jasonburkert-ccpo: This was changed because requests to this operation shouldn't have any path parameters.  I don't want to catch empty path parameters, I want to allow empty path parameters.  Request body shape checking was failing validation when path parameters were not included.

jasonburkert-ccpo: This should not alter the behavior of `createApplication` or `createEnvironment` because the presence of the path parameters is merely checked prior to their individual evaluation in the remainder of the boolean logic.

The initial presence check allows my operation to not provide path parameters, but continues to allow their evaluation by other operations.

What's more, the unit tests in `shapeValidator.test.ts` cover the altered function.

jasonburkert-ccpo:
> Where is the business rules validation for "This portfolio name already exists. Please try another name." (Rule 1.1)?

@zachclark-ccpo 
Good catch.  I had a question about that business rule, and Tom just confirmed names should be unique across all portfolios.  I'll implement that.  Thanks for the code reference.

packages/api/utils/shapeValidator.ts
juliusfitzhugh-ccpo: I tried it out on my deployed stack, which has the `createApplication` and `createEnvironment`, and it worked as expected.

packages/api/portfolios/createPortfolio.ts
```diff
@@ -0,0 +1,101 @@
...
+}
+
+// a workaround for ensuring validation does not give a false failure due to 'allOf'
+// also a band-aid for ensuring BaseObject properties (e.g., id) are not accepted
+// when sent in the request body
+const { type, additionalProperties, required: portfolioRequired } = schema.Portfolio;
+const { required: baseRequired } = schema.PortfolioBase;
+const { portfolioManagers, description, owner /*, applications */ } = schema.Portfolio.properties;
+const { csp, dodComponents, name } = schema.PortfolioBase.properties;
+const { portfolioAdministrators } = schema.PortfolioAccess.properties;
```
juliusfitzhugh-ccpo, line 76: This allows `portfolioAdministrators` through the validator, but there is no mapping to `administrators` which is returned after creating the portfolio.  The simplest solution appears to be update the spec like below.  Otherwise it appears the `Portfolio` entity (or another class) has to be updated with `portfolioAdministrators`.

```yaml
PortfolioAccess:
      description: >-
        Represents a set of Operators who should be granted access to
        a Portfolio at a specific access level. Uses an enum construct for forward
        compatibility should other access levels be specified.
      properties:
        administrators:
          type: array
          items:
            type: string
            format: email
```

zachclark-ccpo (approved):

packages/api/portfolios/createPortfolio.ts
jasonburkert-ccpo: Agree this is the simplest solution.  Changed in 88d13a0c85d67b37279c6ebb41b677da58c467cc

jasonburkert-ccpo:
> > Where is the business rules validation for "This portfolio name already exists. Please try another name." (Rule 1.1)?
> 
> @zachclark-ccpo Good catch. I had a question about that business rule, and Tom just confirmed names should be unique across all portfolios. I'll implement that. Thanks for the code reference.

@zachclark-ccpo This was implemented in 32849d97fee2bdc3ade2ac9c11182afcdf56a284

packages/api/portfolios/createPortfolio.ts
```diff
@@ -0,0 +1,111 @@
...
+const { administrators } = schema.PortfolioAccess.properties;
+const portfolioSchema = {
+  type,
+  additionalProperties,
+  required: baseRequired.concat(portfolioRequired),
+  properties: {
+    portfolioManagers,
+    description,
+    owner,
+    // applications,
```
zachclark-ccpo, line 95: Should this still be commented out?

zachclark-ccpo (approved):

packages/api/portfolios/createPortfolio.ts
jasonburkert-ccpo: It was my intent to leave it commented for this initial operation.

PR #529, opened 2021-12-13
Fix Step Functions LogGroup Error and other database errors

juliusfitzhugh-ccpo:
This PR fixes a few common errors that were seen across the team and implements the changes that fixes those errors (RDS endpoints implementation).
- Add `logGroupName` for the state machine in the stack and use the prefix `/aws/vendedlogs/states/` to over come the `Invalid Logging Configuration: The CloudWatch Logs Resource Policy size was exceeded.` when trying to `cdk deploy`.
- Add entities classes for the connection configuration to prevent `EntityMetadataNotFoundError`.
- Add column types to entities to prevent `CloumnTypeUndefinedError`.

### DIFF
--- a/packages/api/utils/database.ts
+++ b/packages/api/utils/database.ts
@@ -3,6 +3,11 @@ import * as rdsIam from "atat-web-api-rds/auth/iam";
 import path from "path";
 import * as fs from "fs";
 import { TlsOptions } from "tls";
+import { Portfolio } from "../../orm/entity/Portfolio";
+import { Application } from "../../orm/entity/Application";
+import { Environment } from "../../orm/entity/Environment";
+import { TaskOrder } from "../../orm/entity/TaskOrder";
+import { Clin } from "../../orm/entity/Clin";
 
 let CONNECTION: Connection | undefined;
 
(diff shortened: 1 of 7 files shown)


packages/api/utils/database.ts
```diff
@@ -80,7 +85,7 @@ export async function createConnection(): Promise<Connection> {
     },
     logging: "all",
     database: databaseName,
-    entities: ["/opt/nodejs/orm/entity/**.js"],
+    entities: [Portfolio, Application, Environment, TaskOrder, Clin, "/opt/nodejs/orm/entity/**.js"],
```
jeffsegal-ccpo, line 88: Will we need to do this forever for all new entities or is this a stop-gap to get through the initial cut of implementations?

jeffsegal-ccpo (commented):
Thanks Julius. It's super helpful to collate all these things together in a separate PR so everyone can move together as we get past them.

jasonburkert-ccpo (approved):
Well done.  Thanks.

packages/api/utils/database.ts
jasonburkert-ccpo: This is a documented capability.  Reference [this comment](https://github.com/dod-ccpo/atat-web-api/pull/518#discussion_r767985259) where I addressed similar question on #518

jasonburkert-ccpo: I think we also need the *.js references when running migrations.

juliusfitzhugh-ccpo: I'm not completely sure. @jasonburkert-ccpo found this work around and might have a better idea as if this is needed or if there is a better way.

jeffsegal-ccpo: Got it. No big deal for now but we can get to the bottom of it later. Not ideal to have to remember to put new entities in a config file but also not the end of the world.

PR #534, opened 2021-12-15
Added SAML IdP for GD

jeffsegal-ccpo:
This PR sets us up to integrate with the IdP that Robert Branch from GD set up for us. It defines custom attributes using the names he gave us and then sets up the mappings accordingly.

Ticket: AT-6874

### DIFF
--- a/infrastructure/lib/constructs/authentication.ts
+++ b/infrastructure/lib/constructs/authentication.ts
@@ -61,7 +61,7 @@ export interface SAMLIdentityProviderProps extends IdentityProviderProps {
   /**
    * The URL where the SAML Metadata file can be fetched from.
    */
-  metadataURL: string;
+  metadataURL: cdk.SecretValue | string;
 }
 
 /**
--- a/infrastructure/lib/constructs/authentication.ts
+++ b/infrastructure/lib/constructs/authentication.ts
@@ -108,9 +108,14 @@ export interface CognitoAuthenticationProps {
 
 function samlAttributeMapping(props: CognitoAuthenticationProps): { [key: string]: string } {
   return {
-    email: "email",
-    family_name: "lastName",
-    given_name: "firstName",
+    email: "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/emailaddress",
+    family_name: "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/surname",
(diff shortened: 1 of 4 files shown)


kylelaker-ccpo (commented):
This looks reasonable to me! I won't approve until we've seen it run.

infrastructure/lib/constructs/authentication.ts
```diff
@@ -215,22 +220,17 @@ export class CognitoAuthentication extends cdk.Construct {
         requireDigits: true,
         requireLowercase: true,
         requireSymbols: true,
-        requireUppercase: true,
+      },
+      customAttributes: {
+        groups: new cognito.StringAttribute({ mutable: true, maxLen: 2000 }),
+        dod_id: new cognito.StringAttribute({ mutable: true }),
```
kylelaker-ccpo, line 226: Should this be mutable?

infrastructure/lib/constructs/authentication.ts
```diff
@@ -108,9 +108,14 @@ export interface CognitoAuthenticationProps {
...
-    family_name: "lastName",
-    given_name: "firstName",
+    email: "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/emailaddress",
+    family_name: "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/surname",
+    given_name: "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/givenname",
+    phone_number: "telephone",
+    "custom:dod_id": "employeeID",
+    "custom:rank": "rank",
+    "custom:citizenship": "citizenship",
+    "custom:branch": "branch",
```
kylelaker-ccpo, line 118: How do you feel about the LOE to make these attributes passed in to `idpProps` to the AtatWebApiStack? It can get added to the tech debt list. But maybe it's fine if we consider these just won't be mapped if they aren't in the assertion.

kylelaker-ccpo (approved):

infrastructure/lib/constructs/authentication.ts
kylelaker-ccpo, line 226:
```suggestion
        dod_id: new cognito.StringAttribute({ minLen: 10, maxLen: 10, mutable: false }),
```

Feels like it'd enforce constraints a little better. But since we can't modify the field later, maybe we don't want to do that.

PR #536, opened 2021-12-15
Add an updated validateRequestShape func

juliusfitzhugh-ccpo:
Add a validateRequestShape func in a new file to provide basic
validation for the requests sent using the 'atat-internal-api'
spec. The new internal api spec requires a slight change in the
path parameters and portfolioDraftId is no longer being used.

All of the pathParameters are being validated by API Gateway, so
there is no need to include validation for the required
pathParameters (e.g., portoflioId, environmentId, applicationId).

### DIFF
--- a/packages/api/utils/validateRequestShape.test.ts
+++ b/packages/api/utils/validateRequestShape.test.ts
@@ -1,0 +1,37 @@
+import { validateRequestShape } from "../utils/validateRequestShape";
+import { ApiGatewayEventParsed } from "./eventHandlingTool";
+import { v4 as uuidv4 } from "uuid";
+import { Environment } from "../../orm/entity/Environment";
+
+describe("Shape validation tests", () => {
+  jest.spyOn(console, "error").mockImplementation(() => jest.fn()); // suppress output
+  const validEventRequest: ApiGatewayEventParsed<Environment> = {
+    body: { name: "env name" },
+    pathParameters: { portfolioId: uuidv4(), applicationId: uuidv4(), environmentId: uuidv4() },
+  } as any;
+  const undefinedBodyEvent: ApiGatewayEventParsed<Environment> = {
+    body: undefined,
+    pathParameters: { portfolioId: uuidv4() },
+  } as any;
+  const invalidUuidv4: ApiGatewayEventParsed<Environment> = {
+    body: { name: "env name" },
+    pathParameters: { portfolioId: uuidv4(), applicationId: "not uuidv4" },
+  } as any;
+
+  it("should return a SetupSuccess object", async () => {
+    expect(validateRequestShape<Environment>(validEventRequest)).toEqual({
+      bodyObject: validEventRequest.body,
+      path: validEventRequest.pathParameters,
+    });
+  });
+  it("should throw error if request body invalid", async () => {
+    expect(() => {
+      validateRequestShape<Environment>(undefinedBodyEvent);
+    }).toThrow(Error("Shape validation failed, invalid request body"));
+  });
+  it("should throw error if invalid UUIDv4", async () => {
+    expect(() => {
+      validateRequestShape<Environment>(invalidUuidv4);
+    }).toThrow(Error("Shape validation failed, invalid UUIDv4"));
+  });
+});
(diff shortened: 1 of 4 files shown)


packages/api/utils/validateRequestShape.ts
```diff
@@ -0,0 +1,35 @@
...
+ * @param event - The incoming API Gateway Request proxied to Lambda
+ * @param extraValidators - Additional validators that check whether the body is a valid object of Type T
+ * @returns SetUpSuccess object if event passes validation, otherwise it throws an error
+ */
+export function validateRequestShape<T>(
+  event: ApiGatewayEventParsed<T>,
+  ...extraValidators: Array<(obj: unknown) => obj is T>
+): SetupSuccess<T> {
+  if (!Object.values(event.pathParameters as APIGatewayProxyEventPathParameters).every(isValidUuidV4)) {
+    throw createError(404, "Shape validation failed, invalid UUIDv4");
```
jasonburkert-ccpo, line 22: Let's make this more generic, like _invalid path parameter_ or _invalid identifier_.  Even in the new API spec the only place we specify uuid is the S3 object key.

packages/api/utils/validateRequestShape.test.ts
```diff
@@ -0,0 +1,37 @@
...
+  const undefinedBodyEvent: ApiGatewayEventParsed<Environment> = {
+    body: undefined,
+    pathParameters: { portfolioId: uuidv4() },
+  } as any;
+  const invalidUuidv4: ApiGatewayEventParsed<Environment> = {
+    body: { name: "env name" },
+    pathParameters: { portfolioId: uuidv4(), applicationId: "not uuidv4" },
+  } as any;
+
+  it("should return a SetupSuccess object", async () => {
```
jasonburkert-ccpo, line 21: Maybe a couple of additional tests that check for `SetupSuccess` / `SetupError`
```suggestion
  it("should return a SetupSuccess object", async () => {
    expect(validateRequestShape<Environment>(validEventRequest)).toBeInstanceOf(SetupSuccess);
  });
  it("should return this exact object", async () => {
```

jasonburkert-ccpo (changes requested):
Looking good.  I have a couple small comments, but in particular let's remove mention of uuid from the error message.

packages/api/utils/validateRequestShape.ts
```diff
@@ -0,0 +1,35 @@
...
+ * Check if incoming POST Request passes basic shape validation
+ *
+ * This shape validation checks the pathParameter to ensure it is not null or undefined, and that is
+ * a valid UUIDv4.
+ *
+ * @param event - The incoming API Gateway Request proxied to Lambda
+ * @param extraValidators - Additional validators that check whether the body is a valid object of Type T
+ * @returns SetUpSuccess object if event passes validation, otherwise it throws an error
+ */
+export function validateRequestShape<T>(
```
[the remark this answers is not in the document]
juliusfitzhugh-ccpo: There were a few other files related to validation, so I named it after the function to avoid confusion. But `shapeValidator.ts` works for me.

packages/api/utils/validateRequestShape.ts
juliusfitzhugh-ccpo: Sure, it can be changed. Originally the message was linked to the `switch` statement in `errorHandlingMiddleware`, but is currently being caught with `error.name && error.name.startsWith("EntityNotFoundError")` instead of the `switch` statement in PR #500.

packages/api/utils/validateRequestShape.test.ts
juliusfitzhugh-ccpo: Added an additional test for `SetupSuccess`. I do not think there are any `SetupError`, but the `.toThrow` catches the errors thrown when there is an error.

packages/api/utils/validateRequestShape.ts
jasonburkert-ccpo: If this isn't making its way up to the user, nevermind.  What you've got in #500 in `errorHandlingMiddleware` 
 returning `NO_SUCH_PORTFOLIO_OR_APPLICATION` is fine.

juliusfitzhugh-ccpo: Made the change, shouldn't make a difference.

jasonburkert-ccpo (approved):

zachclark-ccpo (commented):
Looks good to me

zachclark-ccpo (approved):

PR #542, opened 2021-12-17
Implement updateEnvironment func (RDS endpoint)

juliusfitzhugh-ccpo:
Implement `PUT /portfolios/:portfolioId/applications/:applicationId/environments/:environmentId`

Add `updateEnvironment` operation to completely replace an environment by id 
(internal api - RDS endpoint) with tests only for local development to ensure the 
expected behavior.

Connect function with infrastructure.

Ticket: AT-6893

### DIFF
--- a/atat_internal_api.yaml
+++ b/atat_internal_api.yaml
@@ -775,7 +775,7 @@ paths:
       x-amazon-apigateway-integration:
         httpMethod: "POST"
         uri:
-          Fn::Sub: "arn:${AWS::Partition}:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${NotImplementedFunction.Arn}/invocations"
+          Fn::Sub: "arn:${AWS::Partition}:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${UpdateEnvironmentFunction.Arn}/invocations"
         type: "aws_proxy"
       security:
         Fn::If:
--- a/packages/api/portfolios/environments/createEnvironment.test.ts
+++ b/packages/api/portfolios/environments/createEnvironment.test.ts
@@ -18,6 +18,21 @@ describe("createEnvironment", () => {
     console.log(result?.body);
     expect(result?.statusCode).toBe(SuccessStatusCode.CREATED);
   });
+  it.skip("requests with properties from BaseObject should cause status code 400", async () => {
+    const validRequest: ApiGatewayEventParsed<IEnvironmentCreate> = {
+      body: {
+        name: "flaw check 1",
+        id: "7bc938ca-4c1c-4740-8ccf-18c940a70865",
+        createdAt: "2020-12-17T18:29:42.769Z",
+        updatedAt: "2020-12-17T18:29:42.769Z",
+        archivedAt: "2099-12-17T18:29:42.769Z",
+      },
+      requestContext: { identity: { sourceIp: "7.7.7.7" } },
+    } as any;
+    const result = await handler(validRequest, {} as Context, () => null);
+    console.log(result?.body);
+    expect(result?.statusCode).toBe(ErrorStatusCode.BAD_REQUEST);
+  });
   it.skip("not a unique env name error, testing locally only", async () => {
     const validRequest: ApiGatewayEventParsed<IEnvironmentCreate> = {
       body: { name: "naming makes a big difference" },
--- a/packages/api/portfolios/environments/createEnvironment.test.ts
+++ b/packages/api/portfolios/environments/createEnvironment.test.ts
@@ -5,21 +5,22 @@ import { IEnvironmentCreate } from "../../../orm/entity/Environment";
 import { ErrorStatusCode, SuccessStatusCode } from "../../utils/response";
 
 describe("createEnvironment", () => {
+  const validRequest: ApiGatewayEventParsed<IEnvironmentCreate> = {
+    body: { name: "new local testing after refactor" },
+    pathParameters: {
+      portfolioId: "91c6bd1f-5ed8-413b-8f85-55a62dd50ad3",
+      applicationId: "7bc938ca-4c1c-4740-8ccf-18c940a70862",
+    },
+    requestContext: { identity: { sourceIp: "7.7.7.7" } },
+  } as any;
   it.skip("successful operation testing locally only", async () => {
-    const validRequest: ApiGatewayEventParsed<IEnvironmentCreate> = {
-      body: { name: "new local testing and a unique name" },
-      pathParameters: {
-        portfolioId: "91c6bd1f-5ed8-413b-8f85-55a62dd50ad3",
-        applicationId: "7bc938ca-4c1c-4740-8ccf-18c940a70862",
-      },
-      requestContext: { identity: { sourceIp: "7.7.7.7" } },
-    } as any;
     const result = await handler(validRequest, {} as Context, () => null);
     console.log(result?.body);
     expect(result?.statusCode).toBe(SuccessStatusCode.CREATED);
   });
   it.skip("requests with properties from BaseObject should cause status code 400", async () => {
-    const validRequest: ApiGatewayEventParsed<IEnvironmentCreate> = {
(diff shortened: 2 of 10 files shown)


packages/api/utils/businessRulesValidation.test.ts
```diff
@@ -0,0 +1,9 @@
+import { uniqueNameValidator } from "./businessRulesValidation";
+
+describe("uniqueNameValidator", function () {
+  it("should throw an error if name is not unique", async () => {
+    const name = "unique name";
+    const names = ["not unique", "unique name"];
```
jasonburkert-ccpo, line 6: The test is effective, but I think the specific values are confusing.
```suggestion
    const name = "orange";
    const names = ["apple", "orange"];
```

jasonburkert-ccpo (approved):
Looks good.  Suggest adding catch to `updateEnvironment`

packages/api/portfolios/environments/createEnvironment.ts
```diff
@@ -36,6 +36,7 @@ export async function baseHandler(
 
   // set up database connection
   const connection = await createConnection();
+  const environmentRepository = connection.getCustomRepository(EnvironmentRepository);
```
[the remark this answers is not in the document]
juliusfitzhugh-ccpo: Stole the idea from @zachclark-ccpo when I was looking over his draft for `createApplication`.

packages/api/portfolios/environments/updateEnvironment.ts
```diff
@@ -0,0 +1,73 @@
...
+    const environment = await environmentRepository.findOneOrFail({ id: environmentId });
+
+    const environmentNames = await environmentRepository.getAllEnvironmentNames(application.id);
+
+    // Throws error if duplicate name found
+    uniqueNameValidator(requestBody.name, environmentNames);
+
+    response = await environmentRepository.updateEnvironment(environment.id, requestBody);
+    console.log("Updated Environment: " + JSON.stringify(response));
+  } finally {
```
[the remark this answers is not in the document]
juliusfitzhugh-ccpo: Do we want to have a `catch` block here instead and remove this error from `errorHandlingMiddleware` as a `default`?

packages/api/utils/businessRulesValidation.test.ts
juliusfitzhugh-ccpo: Yeah, good point.

packages/api/portfolios/environments/updateEnvironment.ts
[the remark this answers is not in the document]
jasonburkert-ccpo: Per our discussion, skip the catch block and we'll rely on `errorHandlingMiddleware`.  Thanks for the great advice.

atat_internal_api.yaml
```diff
@@ -1920,7 +1920,7 @@ components:
       type: object
       allOf:
         - $ref: "#/components/schemas/BaseObject"
-        - $ref: '#/components/schemas/AppEnvAccess'
+        - $ref: "#/components/schemas/AppEnvAccess"
```
jasonburkert-ccpo, line 1923: We seem to have this same pattern in a number of places near `$ref`s

packages/api/portfolios/environments/createEnvironment.test.ts
```diff
@@ -18,6 +18,21 @@ describe("createEnvironment", () => {
...
   });
+  it.skip("requests with properties from BaseObject should cause status code 400", async () => {
+    const validRequest: ApiGatewayEventParsed<IEnvironmentCreate> = {
+      body: {
+        name: "flaw check 1",
+        id: "7bc938ca-4c1c-4740-8ccf-18c940a70865",
+        createdAt: "2020-12-17T18:29:42.769Z",
+        updatedAt: "2020-12-17T18:29:42.769Z",
+        archivedAt: "2099-12-17T18:29:42.769Z",
+      },
```
jasonburkert-ccpo, line 29: Just noticing in my own branch that this should be otherwise a valid body so the 400 isn't caused by not including required properties.

packages/api/portfolios/environments/createEnvironment.ts
```diff
@@ -96,11 +89,29 @@ export async function baseHandler(
   return new ApiSuccessResponse<Array<IEnvironmentCreate>>(insertedEnvironments, SuccessStatusCode.CREATED);
 }
 
+// a workaround for ensuring validation does not give a false failure due to 'allOf'
+// also a band-aid for ensuring BaseObject properties (e.g., id) are not accepted
+// when sent in the request body
+const { description, type, additionalProperties, required } = internalSchema.Environment;
+const { administrators, contributors, readOnlyOperators } = internalSchema.Environment.allOf[1].properties;
```
jasonburkert-ccpo, line 96: Would this be clearer if it referred directly to `internalSchema.AppEnvAccess.properties`?

packages/api/utils/businessRulesValidation.test.ts
jasonburkert-ccpo, line 6: The old values are back!

packages/api/portfolios/environments/updateEnvironment.ts
```diff
@@ -0,0 +1,78 @@
...
+    const application = await connection.getRepository(Application).findOneOrFail({
+      id: applicationId,
+    });
+    const environment = await connection
+      .getCustomRepository(EnvironmentRepository)
+      .findOneOrFail({ id: environmentId });
+
+    const environmentNames = await connection
+      .getCustomRepository(EnvironmentRepository)
+      .getAllEnvironmentNames(application.id);
```
jasonburkert-ccpo, line 55: Since these are both reads before the update, we could probably run them in parallel with await `Promise.all()`.  https://stackoverflow.com/a/35612484

jasonburkert-ccpo (approved):
Left some observations. Overall, nice work.

packages/api/portfolios/environments/createEnvironment.ts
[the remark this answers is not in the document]
juliusfitzhugh-ccpo: Yeah, I agree. Changed.

packages/api/portfolios/environments/updateEnvironment.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    // Throws error if duplicate name found
+    uniqueNameValidator(requestBody.name, environmentNames);
+
+    response = await connection
+      .getCustomRepository(EnvironmentRepository)
+      .updateEnvironment(environment.id, requestBody);
```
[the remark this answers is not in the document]
juliusfitzhugh-ccpo: Weird, I thought I changed that... Might have happened during a rebase or something locally. Changed back.

packages/api/utils/businessRulesValidation.test.ts
juliusfitzhugh-ccpo: Yeah, I think something happened, because I remember accepting your changes previously.

atat_internal_api.yaml
juliusfitzhugh-ccpo: Yeah, that is true. I just happened to see that one. I reverted it back. I see that there are a good amount with both `' '` and `" "` being used. I figured since we will probably be making updates to the spec anyway that i'll leave it as is for now and only change the adding of the function to use.

packages/api/portfolios/environments/createEnvironment.test.ts
juliusfitzhugh-ccpo: Thanks, for pointing that out. Updated to ensure it is a valid request and only errors for the added properties.

packages/api/portfolios/environments/updateEnvironment.ts
juliusfitzhugh-ccpo: I think this should definitely be considered to run these calls in parallel. I originally considering doing that with the calls to check for the `ids`, but the error handling was thrown off slightly so I removed them. To keep the errors clean for now I left this as-is, but as a part of refactoring we should try to run calls in parallel and figure out a nice way to deal with `error` handling when a `reject` does happen in `Promise.all()`.

PR #543, opened 2021-12-17
Implement patchEnvironment operation (RDS endpoint)

juliusfitzhugh-ccpo:
Implement PATCH /portfolios/:portfolioId/applications/:applicationId/environments/:environmentId

Add `patchEnvironment` operation to update operators in an environment by id
(internal api - RDS endpoint) with tests only for local development to ensure the
expected behavior.

Connect function with infrastructure.

Ticket: AT-6893

### DIFF
--- a/packages/api/portfolios/environments/patchEnvironment.ts
+++ b/packages/api/portfolios/environments/patchEnvironment.ts
@@ -6,24 +6,24 @@ import { Environment, IEnvironmentOperators } from "../../../orm/entity/Environm
 import { EnvironmentRepository } from "../../repository/EnvironmentRepository";
 import { APIGatewayProxyResult, Context } from "aws-lambda";
 import { ApiSuccessResponse, SuccessStatusCode } from "../../utils/response";
-import internalSchema = require("../../models/internalSchema.json");
+import { ApiGatewayEventParsed } from "../../utils/eventHandlingTool";
+import { validateRequestShape } from "../../utils/shapeValidator";
 import middy from "@middy/core";
-import xssSanitizer from "../../utils/xssSanitizer";
+import internalSchema = require("../../models/internalSchema.json");
+import { wrapSchema } from "../../utils/schemaWrapper";
+import { IpCheckerMiddleware } from "../../utils/ipLogging";
 import jsonBodyParser from "@middy/http-json-body-parser";
+import xssSanitizer from "../../utils/xssSanitizer";
 import validator from "@middy/validator";
 import JSONErrorHandlerMiddleware from "middy-middleware-json-error-handler";
-import cors from "@middy/http-cors";
-import { ApiGatewayEventParsed } from "../../utils/eventHandlingTool";
-import { CORS_CONFIGURATION } from "../../utils/corsConfig";
-import { wrapSchema } from "../../utils/schemaWrapper";
 import { errorHandlingMiddleware } from "../../utils/errorHandlingMiddleware";
-import { IpCheckerMiddleware } from "../../utils/ipLogging";
-import { validateRequestShape } from "../../utils/shapeValidator";
+import cors from "@middy/http-cors";
+import { CORS_CONFIGURATION } from "../../utils/corsConfig";
 
 /**
- * Submits an update to an environment of an application
+ * Submits a request to update the operators of an environment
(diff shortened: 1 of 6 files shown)


packages/api/portfolios/environments/patchEnvironment.ts
```diff
@@ -0,0 +1,67 @@
...
+import { CORS_CONFIGURATION } from "../../utils/corsConfig";
+import { wrapSchema } from "../../utils/schemaWrapper";
+import { errorHandlingMiddleware } from "../../utils/errorHandlingMiddleware";
+import { IpCheckerMiddleware } from "../../utils/ipLogging";
+import { validateRequestShape } from "../../utils/shapeValidator";
+
+/**
+ * Submits an update to an environment of an application
+ *
+ * @param event - The PUT request from API Gateway
```
zachclark-ccpo, line 26: The PATCH operation is used to only update the operators, not the entire environment.

packages/api/portfolios/environments/patchEnvironment.ts
```diff
@@ -0,0 +1,67 @@
...
+  let response: Environment;
+
+  try {
+    // ensures the portfolio, application, and environment id exists
+    await connection.getRepository(Portfolio).findOneOrFail({
+      id: portfolioId,
+    });
+    await connection.getRepository(Application).findOneOrFail({
+      id: applicationId,
+    });
```
zachclark-ccpo, line 48: If you are using findOneOrFail for environmentRepository, I don't think you need to search for the portfolio and app beforehand.

packages/api/portfolios/environments/updateEnvironment.ts
```diff
@@ -0,0 +1,78 @@
...
+    // Throws error if duplicate name found
+    uniqueNameValidator(requestBody.name, environmentNames);
+
+    response = await connection
+      .getCustomRepository(EnvironmentRepository)
+      .updateEnvironment(environment.id, requestBody);
+    console.log("Updated Environment: " + JSON.stringify(response));
+  } finally {
+    connection.close();
+  }
```
zachclark-ccpo, line 66: How do you stop the validator from throwing an error if you don't update the name in the request? Say fish environment exists in the db, and you want to update fish environment. how do you prevent validation from throwing an error if you don't rename fish environment (but update something else)?

packages/api/repository/EnvironmentRepository.ts
```diff
@@ -49,8 +50,20 @@ export class EnvironmentRepository extends Repository<Environment> {
...
 
   // PUT update environment
-  updateEnvironment(id: string, changes: IEnvironment): Promise<UpdateResult> {
-    return this.update(id, { ...changes });
+  async updateEnvironment(id: string, overwrites: IEnvironment): Promise<Environment> {
+    await this.update(id, {
+      administrators: [],
+      contributors: [],
+      readOnlyOperators: [],
+      ...overwrites,
```
zachclark-ccpo, line 58: Does this reset the adminstrators to []?

packages/api/portfolios/environments/patchEnvironment.ts
juliusfitzhugh-ccpo: Corrected, I forgot to update the JS doc. Thanks!

packages/api/portfolios/environments/updateEnvironment.ts
juliusfitzhugh-ccpo: I believe this is captured by using `internalSchema.AppEnvAccess` in the validator. Only the `operators` should be updated according to the Spec. I believe the `PUT` operator handles changes to the environment name.

packages/api/repository/EnvironmentRepository.ts
juliusfitzhugh-ccpo: Well it depends on what is sent with the request body. If no `administrators` are sent then yes the administrators are reset (same with all the operators). I went with replacing all operators because I was unsure if the operators should appended to the previous values or be replaced entirely. 

I decided on replacing because if some operators are removed (not included in the request) and others added, I assumed we did not want to keep the older operators. I think I see where you are coming from and we should iron out it should be handled this way or not.

PR #545, opened 2021-12-17
Fix 500 errors on file upload

kylelaker-ccpo:
There were errors during file upload due to a missing S3 bucket name
parameter. This ensures that is added by setting permissions between the
function(s) and the bucket.


juliusfitzhugh-ccpo (approved):
Looks good, those permissions came back to haunt us.

jasonburkert-ccpo (approved):

PR #546, opened 2021-12-17
Update access policies

kylelaker-ccpo:
This updates the permissions for the CloudFormation execution role while
also making improvements to other policies as well.


infrastructure/lib/atat-iam-stack.ts
```diff
@@ -113,6 +99,99 @@ export class AtatIamStack extends cdk.Stack {
...
+            "cognito-idp:*TagResource",
+            "cognito-idp:UpdateGroup",
+            "cognito-idp:UpdateIdentityProvider",
+            "cognito-idp:UpdateResourceServer",
+            "cognito-idp:UpdateUserAttributes",
+            "cognito-idp:UpdateUserPool",
+            "cognito-idp:UpdateUserPoolClient",
+            "cognito-idp:UpdateUserPoolDomain",
+            "ec2:*",
+            "iam:*",
```
juliusfitzhugh-ccpo, line 175: This is more of a sanity check and make sure I understand. But here all of the `iam` permissions (and other services done the same way) will be allowed excepted for the ones denied above.

kylelaker-ccpo: That's correct! Yep!

kylelaker-ccpo: Or like `"cognito-idp:Create*"` allows anything that starts with `Create` from https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazoncognitouserpools.html

PR #547, opened 2021-12-19
Implement getApplication function

zachclark-ccpo:
## Overview
Implemented GET `/portfolios/:portfolioId/applications/:applicationId`, which returns an Application based on the provided ApplicationId.

This request is interesting because `portfolioid` is part of the path, but it is **NOT** used in the actual retrieval or validation of the Application. Perhaps we should revised this API route in the future.

### Testing

Use this ID with **POST** CreateApplication, which will return an **id** in the response, this **id** can be used with
**GET** GetApplication as the **applicationId**
example request

Feature/at 6886

### DIFF
--- a/packages/api/portfolios/application/getApplication.ts
+++ b/packages/api/portfolios/application/getApplication.ts
@@ -17,7 +17,7 @@ import { ApplicationRepository } from "../../repository/ApplicationRepository";
 /**
  * Retrieve an Application based on applicationId
  *
- * @param event - The POST request from API Gateway
+ * @param event - The GET request from API Gateway
  */
 export async function baseHandler(
   event: ApiGatewayEventParsed<Application>,
--- a/packages/api/portfolios/application/getApplication.ts
+++ b/packages/api/portfolios/application/getApplication.ts
@@ -39,7 +39,7 @@ export async function baseHandler(
     connection.close();
   }
 
-  return new ApiSuccessResponse<IApplicationCreate>(response, SuccessStatusCode.CREATED);
+  return new ApiSuccessResponse<IApplicationCreate>(response, SuccessStatusCode.OK);
 }
 
 export const handler = middy(baseHandler)
--- a/packages/api/portfolios/application/getApplication.ts
+++ b/packages/api/portfolios/application/getApplication.ts
@@ -1,0 +1,50 @@
+import "reflect-metadata";
+import { APIGatewayProxyResult, Context } from "aws-lambda";
+import { ApiSuccessResponse, SuccessStatusCode } from "../../utils/response";
+import middy from "@middy/core";
+import xssSanitizer from "../../utils/xssSanitizer";
+import jsonBodyParser from "@middy/http-json-body-parser";
+import JSONErrorHandlerMiddleware from "middy-middleware-json-error-handler";
+import cors from "@middy/http-cors";
+import { ApiGatewayEventParsed } from "../../utils/eventHandlingTool";
+import { validateRequestShape } from "../../utils/shapeValidator";
+import { CORS_CONFIGURATION } from "../../utils/corsConfig";
+import { errorHandlingMiddleware } from "../../utils/errorHandlingMiddleware";
+import { Application, IApplicationCreate } from "../../../orm/entity/Application";
+import { createConnection } from "../../utils/database";
+import { ApplicationRepository } from "../../repository/ApplicationRepository";
+
+/**
+ * Retrieve an Application based on applicationId
+ *
+ * @param event - The POST request from API Gateway
+ */
+export async function baseHandler(
+  event: ApiGatewayEventParsed<Application>,
+  context?: Context
+): Promise<APIGatewayProxyResult> {
+  validateRequestShape<Application>(event);
+  // it seems that we don't even need the portfolioId for this operation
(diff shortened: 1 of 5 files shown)


packages/api/portfolios/application/getApplication.ts
```diff
@@ -0,0 +1,50 @@
...
+ * @param event - The POST request from API Gateway
+ */
+export async function baseHandler(
+  event: ApiGatewayEventParsed<Application>,
+  context?: Context
+): Promise<APIGatewayProxyResult> {
+  validateRequestShape<Application>(event);
+  // it seems that we don't even need the portfolioId for this operation
+  // we should bring this up to Jeff, and see what he thinks
+  const portfolioId = event.pathParameters?.portfolioId as string;
```
juliusfitzhugh-ccpo, line 29: Good point. This also happens for environment operations.

packages/api/portfolios/application/getApplication.ts
```diff
@@ -0,0 +1,50 @@
...
+  const connection = await createConnection();
+  try {
+    // Retrieve application based on applicationId
+    response = await connection.getCustomRepository(ApplicationRepository).getApplication(applicationId);
+    console.log("Response:" + JSON.stringify(response));
+  } finally {
+    connection.close();
+  }
+
+  return new ApiSuccessResponse<IApplicationCreate>(response, SuccessStatusCode.CREATED);
```
juliusfitzhugh-ccpo, line 42: The status code should be `.OK` according to the spec for returning a `200` response upon success.

packages/api/portfolios/application/getApplication.ts
```diff
@@ -0,0 +1,50 @@
...
+import { CORS_CONFIGURATION } from "../../utils/corsConfig";
+import { errorHandlingMiddleware } from "../../utils/errorHandlingMiddleware";
+import { Application, IApplicationCreate } from "../../../orm/entity/Application";
+import { createConnection } from "../../utils/database";
+import { ApplicationRepository } from "../../repository/ApplicationRepository";
+
+/**
+ * Retrieve an Application based on applicationId
+ *
+ * @param event - The POST request from API Gateway
```
juliusfitzhugh-ccpo, line 20: Minor change.
```suggestion
 * @param event - The GET request from API Gateway
```

packages/api/portfolios/application/getApplication.ts
```diff
@@ -0,0 +1,50 @@
...
+    console.log("Response:" + JSON.stringify(response));
+  } finally {
+    connection.close();
+  }
+
+  return new ApiSuccessResponse<IApplicationCreate>(response, SuccessStatusCode.CREATED);
+}
+
+export const handler = middy(baseHandler)
+  .use(xssSanitizer())
```
juliusfitzhugh-ccpo, line 46: It does not hurt to keep the `xssSanitizer` middleware, but from my understanding for requests that do not require a body we can probably remove `xssSanitizer` or is there another reason we should keep it?

packages/api/portfolios/application/getApplication.ts
```diff
@@ -0,0 +1,50 @@
...
+    response = await connection.getCustomRepository(ApplicationRepository).getApplication(applicationId);
+    console.log("Response:" + JSON.stringify(response));
+  } finally {
+    connection.close();
+  }
+
+  return new ApiSuccessResponse<IApplicationCreate>(response, SuccessStatusCode.CREATED);
+}
+
+export const handler = middy(baseHandler)
```
juliusfitzhugh-ccpo, line 45: I believe we want to add `IpCheckerMiddleware` on all of our routes.

packages/api/portfolios/application/getApplication.ts
jasonburkert-ccpo: Agreed.  It can be removed for operations with no request body.

packages/api/portfolios/application/getApplication.ts
jasonburkert-ccpo: Agreed.  Should include as first middleware for all operations.

packages/api/portfolios/application/getApplication.ts
jasonburkert-ccpo: The `portfolioId` is not technically required for the retrieval of the application since `applicationId` should be unique. However, the API design presents a logical and hierarchical set of endpoints that illustrates the relationships between the entities.  I don't think it would makes sense to change this for the simple GET operations simply because an implementation detail allows us to do so.

As Julius explained, we should check that the `id` combination is valid and we're not retrieving the requested application child if the wrong parent is specified in the path parameters.  In that case let's return a 400 status with generic _Invalid input_ result, while logging something more specific about the parent not containing the requested child or something similar.

jasonburkert-ccpo (changes requested):
See comments, mostly echoing Julius

packages/api/portfolios/application/getApplication.ts
zachclark-ccpo: Making additional SELECTs to the database will only slow down the application as the database gets populated.

zachclark-ccpo: We could accomplish the portfolioId validation locally without making another SELECT.

PR #548, opened 2021-12-19
Implement getApplications function

zachclark-ccpo:
## Overview
Implemented GET `/portfolios/:portfolioId/applications`, which returns an array of Applications based on the provided PortfolioId.

### Testing

Use this ID with **GET** get Applications in Postman:

getapps

This will return an array of Applications.

Ticket: AT-6884

### DIFF
--- a/packages/api/portfolios/application/getApplications.test.ts
+++ b/packages/api/portfolios/application/getApplications.test.ts
@@ -1,0 +1,38 @@
+import { handler } from "../application/getApplications";
+import { Context } from "aws-lambda";
+import { ApiGatewayEventParsed } from "../../utils/eventHandlingTool";
+import { ApiSuccessResponse, SuccessStatusCode } from "../../utils/response";
+import { ErrorStatusCode } from "../../../email/utils/statusCodesAndErrors";
+
+const validRequest: ApiGatewayEventParsed<any> = {
+  body: {},
+  pathParameters: {
+    portfolioId: "8e935e1c-cbc9-4db5-81dd-d811652d2b81",
+  },
+} as any;
+// 8e935e1c-cbc9-4db5-81dd-d811652d2b8f
+// These local tests only work the proper local db environment
+describe("Local getApplications tests", () => {
+  it.skip("should return valid Application", async () => {
+    const result = await handler(validRequest, {} as Context, () => null);
+    console.log(result);
+    expect(result).toBeInstanceOf(ApiSuccessResponse);
+    expect(result?.statusCode).toEqual(SuccessStatusCode.CREATED);
+  });
+});
+// Validation tests
+describe("Validation tests", () => {
+  it("should fail due to having an incorrect path parameter", async () => {
+    const validRequest: ApiGatewayEventParsed<any> = {
+      body: {},
+      pathParameters: {
+        portfolioId: "wrong", // invalid path parameter (isn't uuidv4)
+      },
+      requestContext: { identity: { sourceIp: "7.7.7.7" } },
+    } as any;
+
+    const result = await handler(validRequest, {} as Context, () => null);
+    console.log(result?.body);
+    expect(result?.statusCode).toBe(ErrorStatusCode.NOT_FOUND);
+  });
+});
--- a/packages/api/portfolios/application/getApplications.ts
+++ b/packages/api/portfolios/application/getApplications.ts
@@ -36,7 +36,7 @@ export async function baseHandler(
     connection.close();
   }
 
-  return new ApiSuccessResponse<Array<Application>>(response, SuccessStatusCode.CREATED);
+  return new ApiSuccessResponse<Array<Application>>(response, SuccessStatusCode.OK);
 }
 
 export const handler = middy(baseHandler)
--- a/packages/api/portfolios/application/getApplications.ts
+++ b/packages/api/portfolios/application/getApplications.ts
@@ -17,7 +17,7 @@ import { ApplicationRepository } from "../../repository/ApplicationRepository";
 /**
  * Retrieve all Applications associated with the provided portfolioId
  *
- * @param event - The POST request from API Gateway
+ * @param event - The GET request from API Gateway
  */
 export async function baseHandler(
   event: ApiGatewayEventParsed<Application>,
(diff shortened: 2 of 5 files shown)


packages/api/portfolios/application/getApplications.ts
```diff
@@ -0,0 +1,47 @@
...
+  const connection = await createConnection();
+  try {
+    // Retrieve application based on applicationId
+    response = await connection.getCustomRepository(ApplicationRepository).getApplicationsByPortfolioId(portfolioId);
+    console.log("Response:" + JSON.stringify(response));
+  } finally {
+    connection.close();
+  }
+
+  return new ApiSuccessResponse<Array<Application>>(response, SuccessStatusCode.CREATED);
```
juliusfitzhugh-ccpo, line 39: I think the returned status code should be `.OK` according to the spec for a `200` response.

packages/api/portfolios/application/getApplications.ts
```diff
@@ -0,0 +1,47 @@
...
+import { CORS_CONFIGURATION } from "../../utils/corsConfig";
+import { errorHandlingMiddleware } from "../../utils/errorHandlingMiddleware";
+import { Application } from "../../../orm/entity/Application";
+import { createConnection } from "../../utils/database";
+import { ApplicationRepository } from "../../repository/ApplicationRepository";
+
+/**
+ * Retrieve all Applications associated with the provided portfolioId
+ *
+ * @param event - The POST request from API Gateway
```
juliusfitzhugh-ccpo, line 20: Minor change.
```suggestion
 * @param event - The GET request from API Gateway
```

packages/api/repository/ApplicationRepository.ts
```diff
@@ -39,8 +39,9 @@ export class ApplicationRepository extends Repository<Application> {
 
   // DELETE application (hard delete)
   async deleteApplication(applicationId: string): Promise<Application> {
-    console.log("Deleting: " + applicationId);
+    // Check if the app exists, if it doesn't findOneOrFail will throw 404 to error handling middleware
     const appToDelete = await this.findOneOrFail({ id: applicationId });
+    console.log("Deleting: " + applicationId);
     const deleteResult = await this.delete(applicationId);
     console.log(deleteResult);
     return appToDelete;
```
juliusfitzhugh-ccpo, line 47: This could probably be updated to return `void` since no data is to be returned. This relates to a comment in  PR #549 ([comment link](https://github.com/dod-ccpo/atat-web-api/pull/549/files#r772585887)).
```suggestion
  async deleteApplication(applicationId: string): Promise<void> {
    // Check if the app exists, if it doesn't findOneOrFail will throw 404 to error handling middleware
    const appToDelete = await this.findOneOrFail({ id: applicationId });
    console.log("Deleting: " + applicationId);
    const deleteResult = await this.delete(applicationId);
    console.log(deleteResult);
    console.log("Deleted application: " + appToDelete);
```

packages/api/portfolios/application/getApplications.test.ts
```diff
@@ -0,0 +1,38 @@
+import { handler } from "../application/getApplications";
+import { Context } from "aws-lambda";
+import { ApiGatewayEventParsed } from "../../utils/eventHandlingTool";
+import { ApiSuccessResponse, SuccessStatusCode } from "../../utils/response";
+import { ErrorStatusCode } from "../../../email/utils/statusCodesAndErrors";
```
jasonburkert-ccpo, line 5: Looks like we got some duplicated enums in the email package.
```suggestion
import { ApiSuccessResponse, SuccessStatusCode, ErrorStatusCode } from "../../utils/response";
```

packages/api/portfolios/application/getApplications.test.ts
```diff
@@ -0,0 +1,38 @@
...
+  },
+} as any;
+// 8e935e1c-cbc9-4db5-81dd-d811652d2b8f
+// These local tests only work the proper local db environment
+describe("Local getApplications tests", () => {
+  it.skip("should return valid Application", async () => {
+    const result = await handler(validRequest, {} as Context, () => null);
+    console.log(result);
+    expect(result).toBeInstanceOf(ApiSuccessResponse);
+    expect(result?.statusCode).toEqual(SuccessStatusCode.CREATED);
```
jasonburkert-ccpo, line 20:
```suggestion
    expect(result?.statusCode).toEqual(SuccessStatusCode.OK);
```

packages/api/portfolios/application/getApplications.test.ts
```diff
@@ -0,0 +1,38 @@
...
+      pathParameters: {
+        portfolioId: "wrong", // invalid path parameter (isn't uuidv4)
+      },
+      requestContext: { identity: { sourceIp: "7.7.7.7" } },
+    } as any;
+
+    const result = await handler(validRequest, {} as Context, () => null);
+    console.log(result?.body);
+    expect(result?.statusCode).toBe(ErrorStatusCode.NOT_FOUND);
+  });
```
jasonburkert-ccpo, line 37: I think this could be split into two tests.

If the path parameter isn't of the expected format, won't that return 400 BAD_REQUEST?
If the path parameter is of the expected format, but not present, then I'd expect 404 NOT_FOUND.  This particular test would have to use local db and be skipped for now.

packages/api/portfolios/application/getApplications.ts
```diff
@@ -0,0 +1,47 @@
...
+  } finally {
+    connection.close();
+  }
+
+  return new ApiSuccessResponse<Array<Application>>(response, SuccessStatusCode.OK);
+}
+
+export const handler = middy(baseHandler)
+  .use(xssSanitizer())
+  .use(jsonBodyParser())
```
jasonburkert-ccpo, line 44: I think comments on #547 apply here as well.  Should have `IpCheckerMiddleware` first, `xssSanitizer` middleware can be removed because no request body.

packages/api/repository/ApplicationRepository.ts
jasonburkert-ccpo: In theory I agree, since API spec specifies to return 204 NO_CONTENT for operation `deleteApplication`.
However, confusingly, it also says the content should be schema object `Application`.

It seems this is true for all `DELETE` operations with the exception of the operation for deleting a task order file.  They all return their respective schema objects, so this appears to be intentional.

I think we do need to return the deleted Application unless we change the API spec.

>       responses:
>        '204':
>          description: Successful operation
>          content:
>            application/json:
>              schema:
>                $ref: '#/components/schemas/Application'

jasonburkert-ccpo (changes requested):
I left a couple of comments.

PR #549, opened 2021-12-19
Implement deleteApplication function

zachclark-ccpo:
## Overview
Implemented DELETE `/portfolios/:portfolioId/applications/:applicationId`, which deletes an Application based on the provided ApplicationId.

This request is interesting because `portfolioid` is part of the path, but it is **NOT** used in the actual deletion or validation of the Application. Perhaps we should revised this API route in the future.

### Testing

Use this ID with **POST** CreateApplication, which will return an **id** in the response, this **id** can be used with
**DELETE** DeleteApplication as the **applicationId**
example request

*This screenshot is used to represent the path parameters, but the actual request should be made with `DELETE` not `GET`.

Ticket: AT-6888

### DIFF
--- a/packages/api/portfolios/application/deleteApplication.ts
+++ b/packages/api/portfolios/application/deleteApplication.ts
@@ -1,0 +1,49 @@
+import "reflect-metadata";
+import { APIGatewayProxyResult, Context } from "aws-lambda";
+import { NoContentResponse } from "../../utils/response";
+import middy from "@middy/core";
+import xssSanitizer from "../../utils/xssSanitizer";
+import jsonBodyParser from "@middy/http-json-body-parser";
+import JSONErrorHandlerMiddleware from "middy-middleware-json-error-handler";
+import cors from "@middy/http-cors";
+import { ApiGatewayEventParsed } from "../../utils/eventHandlingTool";
+import { validateRequestShape } from "../../utils/shapeValidator";
+import { CORS_CONFIGURATION } from "../../utils/corsConfig";
+import { errorHandlingMiddleware } from "../../utils/errorHandlingMiddleware";
+import { Application } from "../../../orm/entity/Application";
+import { createConnection } from "../../utils/database";
+import { ApplicationRepository } from "../../repository/ApplicationRepository";
+
+/**
+ * Retrieve an Application based on applicationId
+ *
+ * @param event - The POST request from API Gateway
+ */
+export async function baseHandler(
+  event: ApiGatewayEventParsed<Application>,
+  context?: Context
+): Promise<APIGatewayProxyResult> {
+  validateRequestShape<Application>(event);
+  // it seems that we don't even need the portfolioId for this operation
+  // we should bring this up to Jeff, and see what he thinks
+  const portfolioId = event.pathParameters?.portfolioId as string;
+  const applicationId = event.pathParameters?.applicationId as string;
+  let response;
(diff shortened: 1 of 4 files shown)


packages/api/portfolios/application/deleteApplication.ts
```diff
@@ -0,0 +1,49 @@
...
+import { CORS_CONFIGURATION } from "../../utils/corsConfig";
+import { errorHandlingMiddleware } from "../../utils/errorHandlingMiddleware";
+import { Application } from "../../../orm/entity/Application";
+import { createConnection } from "../../utils/database";
+import { ApplicationRepository } from "../../repository/ApplicationRepository";
+
+/**
+ * Delete an Application based on applicationId
+ *
+ * @param event - The POST request from API Gateway
```
juliusfitzhugh-ccpo, line 20: Minor fix.
```suggestion
 * @param event - The DELETE request from API Gateway
```

packages/api/portfolios/application/deleteApplication.ts
```diff
@@ -0,0 +1,49 @@
...
+  // we should bring this up to Jeff, and see what he thinks
+  const portfolioId = event.pathParameters?.portfolioId as string;
+  const applicationId = event.pathParameters?.applicationId as string;
+  let response;
+  // Establish database connection
+  const connection = await createConnection();
+  try {
+    // Retrieve application based on applicationId
+    response = await connection.getCustomRepository(ApplicationRepository).deleteApplication(applicationId);
+    console.log("Response:" + JSON.stringify(response));
```
juliusfitzhugh-ccpo, line 37: The comment might have been copied over by accident, so it should probably be removed or updated. The `console.log()` could be moved to inside of `deleteApplication` method and there wouldn't be a need for the `console.log()`. The `response` variable is probably not needed either since there is nothing being returned in this operation (I think I forgot to remove it in my branch as well).

```suggestion
   await connection.getCustomRepository(ApplicationRepository).deleteApplication(applicationId);
```

packages/api/portfolios/application/deleteApplication.ts
```diff
@@ -0,0 +1,49 @@
...
+  } finally {
+    connection.close();
+  }
+  return new NoContentResponse();
+}
+
+export const handler = middy(baseHandler)
+  .use(xssSanitizer())
```
jasonburkert-ccpo, line 45: Should have `IpCheckerMiddleware` first, `xssSanitizer` middleware can be removed because no request body.

packages/api/portfolios/application/deleteApplication.ts
```diff
@@ -0,0 +1,49 @@
...
+ */
+export async function baseHandler(
+  event: ApiGatewayEventParsed<Application>,
+  context?: Context
+): Promise<APIGatewayProxyResult> {
+  validateRequestShape<Application>(event);
+  // it seems that we don't even need the portfolioId for this operation
+  // we should bring this up to Jeff, and see what he thinks
+  const portfolioId = event.pathParameters?.portfolioId as string;
+  const applicationId = event.pathParameters?.applicationId as string;
```
jasonburkert-ccpo, line 30: Relevant comment https://github.com/dod-ccpo/atat-web-api/pull/547#discussion_r773465257

jasonburkert-ccpo (changes requested):
See comments.

PR #557, opened 2021-12-22
Fix: name collision on operationId 'deleteTaskOrder'

jasonburkert-ccpo:
This fixes a naming collision in the new API spec.  Same operationId had been used on different endpoints.
```
:1001  /portfolios/{portfolioId}/taskOrders/{taskOrderId}:
:1087      operationId: deleteTaskOrder
...
:1537  /taskOrderFiles/{taskOrderId}:
:1575      operationId: deleteTaskOrder
```

Renamed task order file operation to disambiguate:

- `deleteTaskOrder` -> `deleteTaskOrderFile`

Renamed all task order file operations and existing implementations for consistency:

- `uploadTaskOrder` -> `uploadTaskOrderFile`
- `getTaskOrderMetadata` -> `getTaskOrderFileMetadata`
- `downloadTaskOrder` -> `downloadTaskOrderFile`

The [OpenAPI specification](https://swagger.io/specification/) states:

>Unique string used to identify the operation. The id MUST be unique among all operations described in the API.


juliusfitzhugh-ccpo (approved):
Looks good. We definitely want to avoid any kind of unexpected behavior. Did an error or something else happen?

jasonburkert-ccpo:
@juliusfitzhugh-ccpo No error occurred.  I just happened to notice it.  The old API spec only has a single such operation for the task order file operation.   When the new API spec was added we added the capability to delete entities such as Task Orders, and that introduced the conflict with the PDF file operation.

joshcotten-ccpo (approved):
Second set of eyes

PR #638, opened 2022-02-28
Implement startProvisionJob operation

juliusfitzhugh-ccpo:
After a reset of the repo, the previous AWS step functions (sfn) state machine is being reused for provisioning jobs from SNow.

The lambdas used in the sfn are contained in the `api/provision` folder.

The `start-provision-job.ts` lambda constructs a `CspInvocation` object from the `ProvisionRequest` and is the input into the initial state of the sfn which starts execution of the state machine. Middy is used for validation.
- `POST /provisioning-jobs is valid` and starts `ProvisioningStateMachine` execution
- `POST /provisioning-jobs` invalid inputs respond with validation errors

The `provisioning-sfn-workflow.ts` contains the majority of the components for setting up the sfn states and logic. This is to help reduce the logic in the `atat-web-api-stack`. The `sample-fn.ts` is merely a dummy function used as a temporary placeholder.

The `api-sfn-function` is used as a base function used in the sfn states, but [as pointed out](https://github.com/dod-ccpo/atat-web-api/pull/638#discussion_r816074122) this is not a path we want to continue down as we have previously ran into issues. 

Tests have been moved from `/test` and placed adjacent to the code being tested. The`jest.config.js` has been updated to capture test similar to before. 

Some `linting` and `tsc` errors handled in branches https://github.com/dod-ccpo/atat-web-api/pull/638/commits/677fed22a369039680706045f01c9f2a9919efd1 and https://github.com/dod-ccpo/atat-web-api/pull/638/commits/190b48f7d2b2c6b847a99c56e886de041d7b4088.

Ticket: AT-7050


jest.config.js
```diff
@@ -1,6 +1,6 @@
 module.exports = {
   testEnvironment: 'node',
-  roots: ['<rootDir>/test'],
+  roots: ['<rootDir>/test', '<rootDir>/api', '<rootDir>/utils'],
```
kylelaker-ccpo, line 3: Do we want to just put all tests in `/test`? Or should we move the CDK tests into their respective directories and make this `<rootDir>` or `<rootDir>/**` or something?

lib/constructs/ApiSfnFunction.ts
```diff
@@ -0,0 +1,80 @@
...
+      // vpc: props.lambdaVpc,
+      memorySize: 256,
+      timeout: Duration.seconds(5),
+      ...props.functionPropsOverride,
+    });
+
+    // APIGW service
+    this.fn.addPermission("AllowApiGatewayInvoke", { principal: APIGW_SERVICE_PRINCIPAL });
+    // editing name from API spec
+    (this.fn.node.defaultChild as lambda.CfnFunction).overrideLogicalId(id + "Function");
```
kylelaker-ccpo, line 73: Is this still necessary?

models/CloudServiceProviders.ts
```diff
@@ -0,0 +1,26 @@
...
+  NETWORK_2 = "NETWORK_2",
+  NETWORK_3 = "NETWORK_3",
+}
+
+export const CSP_URI = {
+  [CloudServiceProvider.CSP_A]: CloudServiceProviderUri.A,
+  [CloudServiceProvider.CSP_B]: CloudServiceProviderUri.B,
+  [CloudServiceProvider.CSP_C]: CloudServiceProviderUri.C,
+  [CloudServiceProvider.CSP_D]: CloudServiceProviderUri.D,
+};
```
kylelaker-ccpo, line 26: Would it be better to do:

```ts
export interface ICloudServiceProvider {
  name: string;
  uri: string;
  networks: Network[];
}

export class CloudServiceProvider implements ICloudServiceProvider {
  // IDK if these modifiers are valid together but it does need to be static and public; readonly may
  // be implied by that
  public static readonly CSP_A = new CloudServiceProvider("CSP_A", "www.cspA.com/v1", [Network.A, Network.B]);
  public static readonly CSP_B = new CloudServiceProvider("CSP_B", "www.cspB.com/v1", []);
  public static readonly CSP_C = new CloudServiceProvider("CSP_C", "www.cspC.com/v1", [Network.A]);
  public static readonly CSP_D = new CloudServiceProvider("CSP_D", "www.cspD.com/v1", [Network.C]);

  private constructor(name: string, uri: string, networks: Network[]) {
    this.name = name;
    this.uri = uri;
    this.networks = [...networks];
  }
}
```

This way you always have CSP A associated to its name, URI (which let's be careful because that's not a URI per spec), and networks and you can't ever accidentally do something like:

```
POST ${CloudServiceProviderUri.B}

{
  "CSP": "${CloudServiceProvider.A}"
}

File: lib/atat-web-api-stack.ts [hunk 4]
```diff
@@ -1,11 +1,85 @@
...
+      },
+      logGroup: stateMachineLogGroup,
+    }).stateMachine;
+
+    // the provisioning lambda that translate and invokes the state machine
+    const provisioningJobFn = new ApiSfnFunction(this, "ProvisioningJobRequest", {
+      method: HttpMethod.POST,
+      handlerPath: "api/provision/startProvisionJob.ts",
+      stateMachine: this.provisioningStateMachine,
+    }).fn;
```
kylelaker-ccpo, line 83: This all feels fairly complicated to be at the top-level of the stack -- can we abstract it into its own construct in a new file?

lib/constructs/ApiSfnFunction.ts
```diff
@@ -0,0 +1,80 @@
...
+  readonly functionPropsOverride?: lambdaNodeJs.NodejsFunctionProps;
+
+  /**
+   * The State Machine that will start execution when this function
+   * is called.
+   */
+  readonly stateMachine: sfn.IStateMachine;
+}
+
+export class ApiSfnFunction extends Construct {
```
kylelaker-ccpo, line 41: We ran into a lot of complexity with creating the `ApiXFunction`s before... should we talk about whether this is still the way we want to go?

kylelaker-ccpo (commented):
More questions than anything else; just want to make sure we talk about some of this before rolling forward too much

lib/constructs/ApiSfnFunction.ts
juliusfitzhugh-ccpo: Probably not...I think my understanding is a bit fuzzy for the new APIGW setup, so I included it just in case.

lib/atat-web-api-stack.ts
juliusfitzhugh-ccpo: I agree and tried to trim it down, but I'll try to abstract it await into its own construct.

models/CloudServiceProviders.ts
juliusfitzhugh-ccpo: That is definitely better and avoid some accidents. Thanks for the suggestion! @kylelaker-ccpo Can you expand a little on the URI portion not being a URI per the spec?

lib/constructs/ApiSfnFunction.ts
juliusfitzhugh-ccpo: To be honest, I was going down the path of trying to implement the function using the decorator pattern (as previously suggested), but it was an epic fail and I wasn't getting anywhere. Hence, I did this for the time being to ensure functionality first.

models/CloudServiceProviders.ts
jasonburkert-ccpo: @juliusfitzhugh-ccpo I think you've got it right.  My interpretation is that he's saying `www.cspD.com/v1` is not, [per spec](https://datatracker.ietf.org/doc/html/rfc3986), a URI, but your example `https://www.cspA.com/` would be.

kylelaker-ccpo: Correct! We need to specify a scheme and `https://` would correct. In general, we probably want to _assume_ and _require_ HTTPS. So everything you currently have but with `https://` prepended would be fine!

utils/errors.ts
```diff
@@ -0,0 +1,113 @@
...
+
+/**
+ * To be used when a function has not been implemented
+ */
+export const NOT_IMPLEMENTED = new OtherErrorResponse("Not implemented", ErrorStatusCode.NOT_FOUND);
+
+/**
+ * To be used when a function has not been implemented
+ */
+export const PORTFOLIO_ALREADY_SUBMITTED = new OtherErrorResponse(
```
kylelaker-ccpo, line 110: We probably don't want to add a bunch of errors that we aren't using yet. Thoughts?

lib/constructs/ApiSfnFunction.ts
kylelaker-ccpo: Let's yank it and see what breaks? We can always add it again if we need to

jest.config.js
kylelaker-ccpo: I think in the past @jeffsegal-ccpo has expressed liking having tests "next to" the code they test (iirc)? Maybe it'd be better to move the infra tests (esp since you've got the rest of these tests in place already)?

jeffsegal-ccpo: Thanks @kylelaker-ccpo, that has generally been my preference as modules get larger and you have to scroll three screens down to switch between the source and test files. Happy to defer to team consensus on this though. the previous config was 
```
    roots: ['<rootDir>'],
    testMatch: ['**/*.test.ts'],
```

models/CloudServiceProviders.ts
juliusfitzhugh-ccpo: I will update it with `https://`, thanks for clearing that up.

utils/errors.ts
juliusfitzhugh-ccpo: Yeah, I agree. There were other derivative errors I removed too, so I think this should be done for other errors not being used.

juliusfitzhugh-ccpo: Reduced to 2 errors and removed a decent amount in `error-handling-middleware`.

lib/atat-web-api-stack.ts
```diff
@@ -42,5 +56,25 @@ export class AtatWebApiStack extends cdk.Stack {
         })
       )
     );
+
+    // State Machine
```
jeffsegal-ccpo, line 60: @kylelaker-ccpo were you envisioning seeing all of this functionality wrapped into a separate Construct as a means to modularize the Stack as much as possible?

models/provisioning-jobs.ts
```diff
@@ -0,0 +1,99 @@
...
+  operators: Array<string>;
+}
+
+export interface ProvisionRequest {
+  jobId: string;
+  userId: string;
+  portfolioId: string;
+  operationType: ProvisionRequestType;
+  // could not use the CloudServiceProviders static properties so used strings
+  targetCsp: "CSP_A" | "CSP_B" | "CSP_C" | "CSP_D";
```
jeffsegal-ccpo, line 37: You could use `CSP_A.name` instead though right?

jeffsegal-ccpo (changes requested):
Looks good! Just a few small items to consider.

api/provision/start-provision-job.test.ts
```diff
@@ -0,0 +1,120 @@
...
+        payload: {
+          name: "Sample Portfolio",
+          fundingSources,
+          operators,
+        },
+      },
+      requestContext: { identity: { sourceIp: "9.9.9.9" } },
+    } as any;
+    const response = await handler(request, {} as Context, () => null);
+    console.log(response);
```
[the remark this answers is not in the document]
juliusfitzhugh-ccpo: Yeah, it does add a lot of noise and they aren't needed.

models/provisioning-jobs.ts
juliusfitzhugh-ccpo: No typescript would complain. When I try to use the `CSP_A.name` I get the following errors:
- `'CloudServiceProvider' only refers to a type, but is being used as a namespace here.`
- `Property 'targetCsp' of exported interface has or is using private name 'CloudServiceProvider'`

Then in the `start-provisioning-job.ts` I get another error when constructing the `endpoint` (i.g, `${CloudServiceProvider[targetCsp].uri}/portfolios`)
- `Element implicitly has an 'any' type because expression of type 'CloudServiceProvider.CSP_A.name' can't be used to index type 'typeof CloudServiceProvider'.`

This was the option that prevented errors. I'm open to other suggestions though if there is a better way.

jeffsegal-ccpo: Is that all because `name` isn't `public`? If it's any more complicated then that no worries.

api/provision/sample-fn.ts
```diff
@@ -0,0 +1,14 @@
+import { Context } from "aws-lambda";
+
+/**
+ * Sample lambda used as a placeholder prior to implementation of other lambdas
+ *
+ * @param stateInput - input to the state task that is processed
+ */
+export async function handler(stateInput: any, context?: Context): Promise<unknown> {
```
kylelaker-ccpo, line 8: _eventually_ (i dont think we want to do it now), we are going to want to make this more specific than an `any` not necessarily in the function definition but for sure towards the top of the body.

models/provisioning-jobs.ts
juliusfitzhugh-ccpo: I did try to make `name`  `public`, but that did not solve it either unfortunately.

api/provision/sample-fn.ts
jeffsegal-ccpo: Agreed - that gets us closer to a contract-based approach for making these functions more modular and tightly unit testable.

utils/middleware/error-handling-middleware.ts
```diff
@@ -0,0 +1,41 @@
...
+import { serializeError } from "serialize-error";
+import { ValidationErrorResponse } from "../response";
+import { INTERNAL_SERVER_ERROR, REQUEST_BODY_INVALID } from "../errors";
+
+export const errorHandlingMiddleware = (): middy.MiddlewareObj<APIGatewayProxyEvent, APIGatewayProxyResult> => {
+  const onError: middy.MiddlewareFn<APIGatewayProxyEvent, APIGatewayProxyResult> = async (
+    request
+  ): Promise<ValidationErrorResponse | void> => {
+    const error = serializeError(request.error)!;
+    const errorMessage = error?.message as string;
```
kylelaker-ccpo, line 12: `foo?.attr as` <something that doesn't include undefined> is inherently unsafe, right?

api/provision/start-provision-job.ts
```diff
@@ -0,0 +1,101 @@
...
+export async function baseHandler(event: APIGatewayProxyEvent, context: Context): Promise<APIGatewayProxyResult> {
+  try {
+    const cspInvocationJob = transformProvisionJob(event.body as any);
+    console.log("SentToSfn: " + JSON.stringify(cspInvocationJob));
+
+    // starting the execution
+    const result = await sfnClient.startExecution({
+      input: JSON.stringify(cspInvocationJob),
+      stateMachineArn: SFN_ARN,
+    });
```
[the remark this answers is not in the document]
juliusfitzhugh-ccpo: It did cross my mind, but are you able to add specific logic in API GW to transform the request? I know there is mapping capabilities, but I wasn't sure the extend of them. 

Since we were already using the lambda to transform the incoming request previously, I decide to keep using the lambda to invoke the step function rather than directly from API GW. I'm open to look into it more though as that would reduce the need to use a lambda.

api/provision/sample-fn.ts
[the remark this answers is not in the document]
juliusfitzhugh-ccpo: Yeah, it can. I'm not really sure why I put `unknown`. I just thought this is unknown and put `unknown`, haha.

utils/errors.ts
```diff
@@ -0,0 +1,18 @@
...
+ * To be used when a server side error occurs.  Hides error/implementation details.
+ */
+export const INTERNAL_SERVER_ERROR = new OtherErrorResponse("Server error", ErrorStatusCode.INTERNAL_SERVER_ERROR);
+
+/**
+ * To be used when a request body exists but is invalid
+ * Could be invalid because:
+ *  - request body is not of the expected Content-Type (for example, application/json, or application/pdf)
+ *  - is not valid JSON / PDF
+ *  - when JSON, is not of the expected type (for example, doesn't look like PortfolioStep/FundingStep/ApplicationStep)
```
jasonburkert-ccpo, line 13: Please update the potential reasons a request body could be invalid
- Remove *Step language from our prior approach
- Remove application/pdf as we're no longer expecting PDF content

api/provision/start-provision-job.ts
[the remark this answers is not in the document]
jeffsegal-ccpo: Almost definitely but unless there are any significant issues I'd ask that we punt this to a separate ticket so that we can get the rest of the functions pushed through.

utils/middleware/error-handling-middleware.ts
```diff
@@ -0,0 +1,41 @@
...
+
+    if (errorMessage === "CSP portfolio ID required.") {
```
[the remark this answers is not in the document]
jeffsegal-ccpo: Ah yes I was wondering the same thing when I was looking last week but forgot to add a comment.

lib/constructs/sfn-lambda-invoke-task.ts
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * Creates a mapping from an array of tasks, where each task is used for
+ * constructing the State Machine workflow. the id
+ *
+ * @param scope - the stack which the task belongs to (e.g., this)
+ * @param tasks - array of task objects to be used for creating SfnLambdaTask
+ * @returns - an sfnTasks object with all tasks mapped by the id provided
+ */
+export function mapTasks(scope: Construct, tasks: Array<TaskProps>): TasksMap {
```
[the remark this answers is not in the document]
juliusfitzhugh-ccpo: Yeah, I think in the past we decided on `Array<X>`, but with the reset I think its good to bring it up and standardize it. I kind of like the `Props[]` for simplicity, but I'm okay with either.

tsconfig.json
```diff
@@ -3,7 +3,8 @@
     "target": "ES2018",
     "module": "commonjs",
     "lib": [
-      "es2018"
+      "es2018",
+      "dom"
```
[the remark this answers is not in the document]
juliusfitzhugh-ccpo: This was added as a fix from a `tsc error` that was showing up (see https://github.com/dod-ccpo/atat-web-api/pull/638/commits/677fed22a369039680706045f01c9f2a9919efd1). I'll try removing `dom` and see if it comes back.

utils/middleware/error-handling-middleware.ts
juliusfitzhugh-ccpo: Yeah, that makes sense. Replaced with simply `error.message`.

models/cloud-service-providers.ts
```diff
@@ -0,0 +1,30 @@
...
+  NETWORK_1 = "NETWORK_1",
+  NETWORK_2 = "NETWORK_2",
+  NETWORK_3 = "NETWORK_3",
+}
+export interface ICloudServiceProvider {
+  name: string;
+  uri: string;
+  networks: Array<Network>;
+}
+export class CloudServiceProvider implements ICloudServiceProvider {
```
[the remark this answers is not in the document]
juliusfitzhugh-ccpo: I wondered if they were needed, since the `CloudServiceProvider` defines them as `readonly`. I do not see a reason _not_ to add them though, so adding them.

api/provision/start-provision-job.ts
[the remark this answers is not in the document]
jasonburkert-ccpo: Made note about future ticket

lib/atat-web-api-stack.ts
```diff
@@ -42,5 +56,25 @@ export class AtatWebApiStack extends cdk.Stack {
...
+    const provisioningSfn = new ProvisioningWorkflow(this, { environmentName });
+    this.provisioningStateMachine = new StateMachine(this, "ProvisioningStateMachine", {
+      stateMachineProps: {
+        definition: provisioningSfn.workflow,
+      },
+      logGroup: provisioningSfn.logGroup,
+    }).stateMachine;
+
+    // Provisioning lambda that translate and invokes the state machine
```
jasonburkert-ccpo, line 69:
```suggestion
    // Provisioning lambda that translates and invokes the state machine
```

lib/atat-web-api-stack.ts
```diff
@@ -42,5 +56,25 @@ export class AtatWebApiStack extends cdk.Stack {
...
+    const provisioningJob = new ApiSfnFunction(this, "ProvisioningJobRequest", {
+      method: HttpMethod.POST,
+      handlerPath: "api/provision/start-provision-job.ts",
+      stateMachine: this.provisioningStateMachine,
+    });
+
+    // APIGW Provisioning Job Resource
+    const provisioningJobsResource = apigw.restApi.root.addResource("provisioning-job");
+    provisioningJobsResource.addMethod(provisioningJob.method, new LambdaIntegration(provisioningJob.fn));
```
jasonburkert-ccpo, line 78: To match case of route
```suggestion
    const provisioningJobResource = apigw.restApi.root.addResource("provisioning-job");
    provisioningJobResource.addMethod(provisioningJob.method, new LambdaIntegration(provisioningJob.fn));
```

lib/constructs/provisioning-sfn-workflow.ts
```diff
@@ -0,0 +1,113 @@
...
+ */
+export const internalErrorResponse = sfn.Condition.numberGreaterThanEquals("$.cspResponse.statusCode", 500);
+
+/**
+ * Retry condition check when the status code is 500, with a max retry of 6
+ */
+export const maxRetries = sfn.Condition.and(
+  sfn.Condition.numberGreaterThan("$.cspResponsePass.retryCount", 6),
+  sfn.Condition.numberGreaterThanEquals("$.cspResponse.statusCode", 500)
+);
```
jasonburkert-ccpo, line 32: Do we want to avoid mention of the specific codes and values in the comments so that comments and implementation can't diverge in the future?

api/provision/start-provision-job.ts
```diff
@@ -0,0 +1,101 @@
...
+ *
+ * @param request - provisioning request from SNOW
+ * @returns - transformed request to send to the targeted CSP
+ */
+function transformProvisionJob(request: ProvisionRequest): CspInvocation {
+  const { operationType, portfolioId, payload, targetCsp } = request;
+
+  const headers = {
+    "Content-Type": "application/json",
+    ...CORS_CONFIGURATION,
```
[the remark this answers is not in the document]
jasonburkert-ccpo: The above comment should impact a number of things beyond this specific line. CORS config definitions, imports, packages

api/provision/start-provision-job.test.ts
```diff
@@ -0,0 +1,120 @@
...
+    const request = {
+      body: {
+        ...provisioningBodyNoPayload,
+        payload: {
+          name: "Sample Portfolio",
+          fundingSources,
+          operators,
+        },
+      },
+      requestContext: { identity: { sourceIp: "9.9.9.9" } },
```
jasonburkert-ccpo, line 43: Should we bother to extract this hypothetical value and re-use?

jasonburkert-ccpo (changes requested):
Nice work.  I left a few minor comments.

lib/constructs/provisioning-sfn-workflow.ts
juliusfitzhugh-ccpo: Good point, avoided mentioning the status codes.

api/provision/start-provision-job.test.ts
juliusfitzhugh-ccpo: Yeah, that works. It didn't cross my mind.

api/provision/start-provision-job.ts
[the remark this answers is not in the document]
juliusfitzhugh-ccpo: Removed things related to CORS.

api/provision/sample-fn.ts
juliusfitzhugh-ccpo: Yup, that sounds good to me. This was just a placeholder function for the time being.

utils/cors-config.ts
```diff
@@ -0,0 +1,26 @@
...
+  origin?: string;
+  origins?: string[];
+  exposeHeaders?: string;
+  maxAge?: number | string;
+  requestHeaders?: string;
+  requestMethods?: string;
+  cacheControl?: string;
+}
+
+// TODO(AT-6712): limit the origin(s) permitted by CORS policy
```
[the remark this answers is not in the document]
juliusfitzhugh-ccpo: Removed.

api/provision/start-provision-job.ts
```diff
@@ -0,0 +1,101 @@
...
+const SFN_ARN = process.env.SFN_ARN ?? "";
+
+/**
+ * Starts provisioning request from SNOW and starts the Step Function execution
+ *
+ * @param event - POST request from API Gateway with provisioning job properties
+ */
+export async function baseHandler(event: APIGatewayProxyEvent, context: Context): Promise<APIGatewayProxyResult> {
+  try {
+    const cspInvocationJob = transformProvisionJob(event.body as any);
```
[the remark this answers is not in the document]
juliusfitzhugh-ccpo: Fair. With using `middy` which wraps the `baseHandler` and uses that event as types for the different middlewares being used (from my understanding at least). I came up with the below solution/type, for the time being rather than using `any` in the middlewares. 

```typescript 
export interface ILambdaEvent {
  body: ProvisionRequest;
  requestContext: { identity: { sourceIp: string } };
}
```

tsconfig.json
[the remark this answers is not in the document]
juliusfitzhugh-ccpo: Tried without 'dom' and got the `tsc` error `Cannot find name 'ReadableStream'` from files with a path like `node_modules/@aws-sdk/client-s3/dist-types/models/models_0.d.ts`.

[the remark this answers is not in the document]
kylelaker-ccpo: Ugh yeah... looks like this is required due to https://github.com/aws/aws-sdk-js-v3/issues/2896. Alternatively, we'll need another workaround...

I _think_ `tsc` supports comments in the `tsconfig.json`. Can we add a comment to make a note of this. That _might_ end up breaking SonarCloud (which we'll catch on a PR run) or `esbuild` (which we'll catch as soon as we try to bundle a `NodeJsFunction`). 
```suggestion
      // Required to workaround aws/aws-sdk-js-v3#2896
      "dom"
```

api/provision/start-provision-job.test.ts
kylelaker-ccpo: Small nitpick here if you happen to be in the file again: If we don't care what the `sourceIp` really is, can we pick an address from [the RFC 5737 list](https://datatracker.ietf.org/doc/html/rfc5737#section-3)? `9.9.9.9` is a semi-common DNS service provider so picking something from one of the TEST-NET-N ranges might make it a little more clear that the value is meaningless.

tsconfig.json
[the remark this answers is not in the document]
juliusfitzhugh-ccpo: Added in the comment, but it appears the `eslint` caused it to fail. 

Thanks for referencing my PR to https://github.com/aws/aws-sdk-js-v3/issues/2896

api/provision/start-provision-job.ts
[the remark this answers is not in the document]
juliusfitzhugh-ccpo: Cool, thanks for the suggestion. I'll make a note of this.

tsconfig.json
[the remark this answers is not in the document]
juliusfitzhugh-ccpo: Removed the comment to prevent breaking the pipeline.

lib/constructs/provisioning-sfn-workflow.ts
```diff
@@ -0,0 +1,113 @@
...
+import { Construct } from "constructs";
+import { mapTasks, TasksMap } from "./sfn-lambda-invoke-task";
+
+/**
+ * Successful condition check
+ */
+export const successResponse = sfn.Condition.numberEquals("$.cspResponse.Payload.statusCodeFn", 200);
+
+/**
+ * Client error condition check∂
```
jasonburkert-ccpo, line 13: Strange char snuck in here at the end of the line

package.json
```diff
@@ -33,9 +37,19 @@
     "typescript": "~4.6.2"
   },
   "dependencies": {
+    "@aws-sdk/client-s3": "^3.54.0",
+    "@aws-sdk/client-sfn": "^3.52.0",
+    "@middy/core": "^2.5.7",
+    "@middy/http-cors": "^2.5.7",
```
jasonburkert-ccpo, line 43: Think we can remove this package if no longer in use

jasonburkert-ccpo (changes requested):
Three small things for follow-up

package.json
juliusfitzhugh-ccpo: Thanks, forgot that, removed.

api/provision/start-provision-job.test.ts
```diff
@@ -0,0 +1,120 @@
...
+const provisioningBodyNoPayload = {
+  jobId: "81b31a89-e3e5-46ee-acfe-75436bd14577",
+  userId: "21d18790-bf3e-4529-a361-460ee6d16e0b",
+  portfolioId: "b02e77d1-234d-4e3d-bc85-b57ca5a93952",
+  operationType: ProvisionRequestType.ADD_OPERATORS,
+  targetCsp: CloudServiceProvider.CSP_A.name,
+  targetNetwork: Network.NETWORK_1,
+};
+
+export const requestContext = { identity: { sourceIp: "9.9.9.9" } };
```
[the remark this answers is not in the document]
juliusfitzhugh-ccpo: Updated.

lib/constructs/provisioning-sfn-workflow.ts
juliusfitzhugh-ccpo: Finger slip maybe, removed.

jasonburkert-ccpo (approved):

api/provision/start-provision-job.ts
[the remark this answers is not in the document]
jasonburkert-ccpo: Ticket for the above is `AT-7177`

PR #641, opened 2022-02-28
Create additional CDK unit tests

zachclark-ccpo:
## Overview
This PR involves creating additional unit tests to verify that the CloudFormation templates generated by CDK meet our expectations.

This included writing fine-grained assertions to test specific stacks (`atat-iam-stack`, `atat-net-stack`, and `atat-web-api-stack`)

### Resources used
Official CDK v2 documentation: https://docs.aws.amazon.com/cdk/v2/guide/testing.html

### DIFF
--- a/lib/atat-iam-stack.ts
+++ b/lib/atat-iam-stack.ts
@@ -116,6 +116,7 @@ export class AtatIamStack extends cdk.Stack {
     });
 
     const baseDenies = new iam.ManagedPolicy(this, "AtatUserDenyPolicy", {
+      description: "Denies access to all resources not explicitly granted in other IAM policies",
       statements: [
         new iam.PolicyStatement({
           sid: "DenyAllOrganizations",
--- a/lib/atat-iam-stack.ts
+++ b/lib/atat-iam-stack.ts
@@ -116,7 +116,7 @@ export class AtatIamStack extends cdk.Stack {
     });
 
     const baseDenies = new iam.ManagedPolicy(this, "AtatUserDenyPolicy", {
-      description: "Denies access to all resources not explicitly granted in other IAM policies",
+      description: "Denies access to sensitive resources and actions",
       statements: [
         new iam.PolicyStatement({
           sid: "DenyAllOrganizations",
--- a/test/apigateway.test.ts
+++ b/test/apigateway.test.ts
@@ -28,7 +28,6 @@ describe("ATAT apigateway construct creation", () => {
     testApigw.grantOnRoute(testUser, "*", "/");
     // THEN
     const template = Template.fromStack(stack);
-    // template.hasResourceProperties("AWS::IAM::Policy", {});
     template.hasResourceProperties(
       "AWS::IAM::Policy",
       Match.objectLike({
--- a/test/apigateway.test.ts
+++ b/test/apigateway.test.ts
@@ -1,8 +1,8 @@
 import * as cdk from "aws-cdk-lib";
-import { Match, Template } from "aws-cdk-lib/assertions";
-import { User } from "aws-cdk-lib/aws-iam";
-import { AtatRestApi, RestApiVpcConfiguration } from "../lib/constructs/apigateway";
 import * as ec2 from "aws-cdk-lib/aws-ec2";
+import * as iam from "aws-cdk-lib/aws-iam";
+import { Match, Template } from "aws-cdk-lib/assertions";
+import { AtatRestApi, RestApiVpcConfiguration } from "../lib/constructs/apigateway";
 
 describe("ATAT apigateway construct creation", () => {
   test("Ensure AtatRestApi is created", async () => {
--- a/test/apigateway.test.ts
+++ b/test/apigateway.test.ts
@@ -1,0 +1,108 @@
+import * as cdk from "aws-cdk-lib";
+import { Match, Template } from "aws-cdk-lib/assertions";
+import { User } from "aws-cdk-lib/aws-iam";
+import { AtatRestApi, RestApiVpcConfiguration } from "../lib/constructs/apigateway";
+import * as ec2 from "aws-cdk-lib/aws-ec2";
+
+describe("ATAT apigateway construct creation", () => {
+  test("Ensure AtatRestApi is created", async () => {
+    // GIVEN
+    const app = new cdk.App();
+    const stack = new cdk.Stack(app, "TestStack");
+    // WHEN
+    const testApigw = new AtatRestApi(stack, "SampleApi");
+    testApigw.restApi.root.addMethod("ANY");
+    // THEN
+    const template = Template.fromStack(stack);
+    template.hasResourceProperties("AWS::ApiGateway::RestApi", {});
+  });
+
+  test("Ensure grantOnRoute method properly generates IAM policy with execute-api:Invoke action", async () => {
+    // GIVEN
+    const app = new cdk.App();
+    const stack = new cdk.Stack(app, "TestStack");
+    // WHEN
+    const testApigw = new AtatRestApi(stack, "SampleApi");
+    testApigw.restApi.root.addMethod("ANY");
+    const testUser = new User(stack, "testUser");
+    testApigw.grantOnRoute(testUser, "*", "/");
+    // THEN
+    const template = Template.fromStack(stack);
+    // template.hasResourceProperties("AWS::IAM::Policy", {});
+    template.hasResourceProperties(
+      "AWS::IAM::Policy",
+      Match.objectLike({
+        PolicyDocument: {
+          Statement: [
+            {
+              Action: "execute-api:Invoke",
+              Effect: "Allow",
+            },
+          ],
+          Version: "2012-10-17",
+        },
+      })
+    );
+  });
+
+  test("Ensures AtatRestApi with vpcConfig in props properly sets EndpointConfiguration to PRIVATE", async () => {
+    // GIVEN
+    const app = new cdk.App();
+    const stack = new cdk.Stack(app, "TestStack");
+    // WHEN
+    const vpc = new cdk.aws_ec2.Vpc(stack, "TestVpc", {});
+    const interfaceEndpoint = new cdk.aws_ec2.InterfaceVpcEndpoint(stack, "Test VPC Endpoint", {
+      vpc,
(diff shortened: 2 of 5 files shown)


test/atat-iam.test.ts
```diff
@@ -23,3 +23,212 @@ describe("ATAT network creation", () => {
...
+              Resource: "*",
+              Sid: "XRayReadAccess",
+            },
+          ],
+          Version: "2012-10-17",
+        },
+        Description: "Grants read access to specific resources not in ViewOnlyAccess",
+        Path: "/",
+      })
+    );
```
kylelaker-ccpo, line 115: I am about to open a PR that actually removes at the `dynamodb:` permissions. The others will probably stay. I am not sure that the individual permissions that we associate with each policy are necessarily worth testing. I think there _are_ something we want to test:

- Named exports (which are basically the stack's API) exist as expected and the ARNs parse to a resource of the correct type
- That specific _named_ IAM roles/policies exist (anything where the `ResourceName` attribute, such as `roleName` or `policyName` is provided)
- We don't create any resources of type `AWS::IAM::Policy` (this may fail right now...)

kylelaker-ccpo (commented):
Some initial thoughts on this specific test. Overall, looks good!

kylelaker-ccpo:
@zachclark-ccpo As part of this, can you look into whether it's possible to write reasonable tests for `bin/` or if, because it's all written at the top-level of the module, it won't be. If it is doable, please write those tests; otherwise, please add that directory to the coverage exclusions configuration for SonarCloud.

zachclark-ccpo:
Moving this out of draft to see the Sonarcloud coverage reports.

zachclark-ccpo:
expectedcoverage
Expected coverage after merge

test/atat-iam.test.ts
jeffsegal-ccpo: I agree re: testing the individual permissions but the APIGW one looks great. I think the place where we'll see the most value is in testing our own utility methods to ensure that they do what they intend to do.

zachclark-ccpo: The PR has been updated with the proper IAM policies in https://github.com/dod-ccpo/atat-web-api/pull/641/commits/5793aa47f8d88c36774ff4b9f2fc46b6d6e7bed8. The `AWS::IAM::Policy` tests are not failing, you can see in the checks for the PR that all the jest tests completed.

zachclark-ccpo: We are already testing the named exports for `roleName`, but I am unsure why we wouldn't want to test the actual policy statement to ensure it is what we expect?

jeffsegal-ccpo: My initial issue was that it wasn't clear that we were actually testing our own code with a `given -> when -> expect` structure but upon further review I think they are actually good insofar as we don't just slap whatever our templates generate into those tests but treat those statements as something that we maintain over time with an eye on what we really want to see there.

jeffsegal-ccpo (approved):
No other issues beyond what I've discussed already in the comments. Great stuff!

lib/atat-iam-stack.ts
```diff
@@ -116,6 +116,7 @@ export class AtatIamStack extends cdk.Stack {
     });
 
     const baseDenies = new iam.ManagedPolicy(this, "AtatUserDenyPolicy", {
+      description: "Denies access to all resources not explicitly granted in other IAM policies",
```
kylelaker-ccpo, line 119: That's not really what this does. This would actually be interesting to turn into a permissions boundary...

What this is actually doing is just denying access to Organizations _explicitly_ just in case we give someone `*` (which we do). We really don't want folks mucking with Organizations.

test/apigateway.test.ts
```diff
@@ -0,0 +1,107 @@
...
+              Effect: "Allow",
+            },
+          ],
+          Version: "2012-10-17",
+        },
+      })
+    );
+  });
+
+  test("Ensures AtatRestApi with vpcConfig in props properly sets EndpointConfiguration to PRIVATE", async () => {
```
kylelaker-ccpo, line 47:
```suggestion
  test("Ensures AtatRestApi with vpcConfig in props properly sets EndpointConfiguration to PRIVATE", async () => {
```

We're actually testing a little more than that. It may be more appropriate to break this into _two_ `test()`s. One that checks the `EndpointConfiguration` and one that checks the `Policy`.

test/atat-iam.test.ts
```diff
@@ -23,3 +23,233 @@ describe("ATAT network creation", () => {
...
+  let stack: cdk.Stack;
+  let template: Template;
+
+  beforeEach(() => {
+    app = new cdk.App();
+    stack = new AtatIam.AtatIamStack(app, "TestIamStack", {});
+    template = Template.fromStack(stack);
+  });
+
+  test("Fully assert generalReadAccess IAM managed policy with matchers", async () => {
```
kylelaker-ccpo, line 38: I really think this test is too specific. _any_ change we make to IAM permissions (adding or removing) is going to require a change to this test. That doesn't necessarily feel correct. We probably want to think more about what the API of the IAM stack is and test that, rather than just a raw CloudFormation template diff.

test/atat-iam.test.ts
```diff
@@ -23,3 +23,233 @@ describe("ATAT network creation", () => {
...
+            },
+          ],
+          Version: "2012-10-17",
+        },
+        Description: "Grants additional security auditor access beyond SecurityAudit",
+        Path: "/",
+      })
+    );
+  });
+  test("Fully assert developerRwAccess IAM managed policy with matchers", async () => {
```
kylelaker-ccpo, line 133: I do think this is a good test overall (mostly because it's something we've broken before). I think the test title isn't ideal though. Maybe:

```
Ensure developers can access CDK-related resources
```
?

kylelaker-ccpo (changes requested):
Really love seeing all the work on this but I think as far as testing hygiene, there are a few things we should talk about so we don't turn our tests into a maintenance issue.

test/atat-iam.test.ts
jeffsegal-ccpo: This is what I was equivocating about as well. I was trying to understand the best practices a little deeper, whether there are use cases where we would actually want to do a raw CFN template diff, etc.

test/apigateway.test.ts
zachclark-ccpo: Good idea!

test/atat-iam.test.ts
zachclark-ccpo: The last thing I want is it to become an albatross.

test/apigateway.test.ts
```diff
@@ -0,0 +1,118 @@
...
+      Users: ["testUser"],
+    });
+  });
+
+  test("Ensures AtatRestApi with vpcConfig in props properly sets EndpointConfiguration to PRIVATE", async () => {
+    // GIVEN
+    const app = new cdk.App();
+    const stack = new cdk.Stack(app, "TestStack");
+    // WHEN
+    const vpc = new cdk.aws_ec2.Vpc(stack, "TestVpc", {});
```
kylelaker-ccpo, line 102:
```suggestion
    const vpc = new ec2.Vpc(stack, "TestVpc", {});
```

test/apigateway.test.ts
```diff
@@ -0,0 +1,118 @@
...
+    const interfaceEndpoint = new cdk.aws_ec2.InterfaceVpcEndpoint(stack, "Test VPC Endpoint", {
```
kylelaker-ccpo, line 103:
```suggestion
    const interfaceEndpoint = new ec2.InterfaceVpcEndpoint(stack, "Test VPC Endpoint", {
```

test/apigateway.test.ts
zachclark-ccpo:
```
Property 'ec2' does not exist on type 'typeof import("/home/zclark/atat-02-2022/atat-web-api/node_modules/aws-cdk-lib/index")'.ts(2339)
```

test/apigateway.test.ts
```diff
@@ -0,0 +1,118 @@
+import * as cdk from "aws-cdk-lib";
+import { Match, Template } from "aws-cdk-lib/assertions";
+import { User } from "aws-cdk-lib/aws-iam";
+import { AtatRestApi, RestApiVpcConfiguration } from "../lib/constructs/apigateway";
+import * as ec2 from "aws-cdk-lib/aws-ec2";
```
kylelaker-ccpo, line 5:
```suggestion
import * as cdk from "aws-cdk-lib";
import * as ec2 from "aws-cdk-lib/aws-ec2";
import * as iam from "aws-cdk-lib/aws-iam";
import { Match, Template } from "aws-cdk-lib/assertions";
import { AtatRestApi, RestApiVpcConfiguration } from "../lib/constructs/apigateway";
```

Can we organize imports (here and elsewhere)? Note that this suggestion can't be directly applied. Usage of `User` needs to change to `iam.User`.

test/apigateway.test.ts
```diff
@@ -0,0 +1,118 @@
...
+        PolicyDocument: {
+          Statement: [
+            {
+              Action: "execute-api:Invoke",
+              Effect: "Allow",
+            },
+          ],
+          Version: "2012-10-17",
+        },
+      })
```
kylelaker-ccpo, line 43:
```suggestion
      {
        PolicyDocument: {
          Statement: [
            {
              Action: "execute-api:Invoke",
              Effect: "Allow",
            },
          ],
          Version: "2012-10-17",
        },
      }
```

test/apigateway.test.ts
```diff
@@ -0,0 +1,118 @@
...
+                  stack.resolve((testApigw.restApi.deploymentStage.node.defaultChild as cdk.CfnResource).ref),
+                  `/*${routeName}`,
+                ],
+              ],
+            },
+          },
+        ],
+        Version: "2012-10-17",
+      },
+      Users: ["testUser"],
```
kylelaker-ccpo, line 93:
```suggestion
      Users: [testUser.userName],
```

kylelaker-ccpo (changes requested):

kylelaker-ccpo (approved):
Really appreciate you taking the time to talk through this PR and the changes. This is a really good baseline for us to get started with. I think we'll learn more over time and make additional improvements but I am really happy to be starting here.

PR #655, opened 2022-03-03
Add basic policy for IAM users using the gateway

kylelaker-ccpo:
This creates a basic policy to allow users to access the gateway at
specified routes on specified methods (one could allow "*" on "/*"). It
also creates a basic permissions boundary to ensure that the created
users cannot do anything beyond executing the API.

Ticket: AT-7075

### DIFF
--- a/lib/http.ts
+++ b/lib/http.ts
@@ -1,0 +1,11 @@
+/**
+ * The HTTP methods
+ */
+export enum HttpMethod {
+  GET = "GET",
+  POST = "POST",
+  PUT = "PUT",
+  PATCH = "PATCH",
+  DELETE = "DELETE",
+  OPTIONS = "OPTIONS",
+}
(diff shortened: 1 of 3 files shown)


jasonburkert-ccpo:
This looks reasonable to me.  I appreciate the copious comments about the permission boundary.

lib/http.ts
```diff
@@ -0,0 +1,11 @@
...
+ * The HTTP methods
+ */
+export enum HttpMethod {
+  GET = "GET",
+  POST = "POST",
+  PUT = "PUT",
+  PATCH = "PATCH",
+  DELETE = "DELETE",
+  OPTIONS = "OPTIONS",
+}
```
juliusfitzhugh-ccpo, line 11: Do we want to have this in two locations? I added the same contents at `utils/http.ts`, so wondering if I should delete the one I have and use the one from this location (assuming this will be merged before my branch).

juliusfitzhugh-ccpo (approved):
Only a small question which is not blocking anything. Otherwise looks good.

lib/http.ts
kylelaker: Let's use this one?

juliusfitzhugh-ccpo: Fine by me.

PR #666, opened 2022-03-08
Apply permission boundary only to users

kylelaker-ccpo:


kylelaker-ccpo:
@zachclark-ccpo an interesting test to work on in #641 might be to assert that there isn't a Permission Boundary defined for some types of resources

lib/aspects/user-only-permission-boundary.ts
```diff
@@ -0,0 +1,17 @@
...
+  constructor(boundary: iam.ManagedPolicy) {
+    this.policy = boundary;
+  }
+
+  visit(node: IConstruct): void {
+    if (!(node instanceof iam.User)) {
+      return;
+    }
+    iam.PermissionsBoundary.of(node).apply(this.policy);
+  }
```
juliusfitzhugh-ccpo, line 16: When is this `visit` method actually being called, during `cdk deploy`? 

Also, I believe you mentioned it already, but want to make sure I understand, but here the users IAM gets an additional policy to invoke the `api-gw`, along with any other permissions (if any) that are present for that particular user?

juliusfitzhugh-ccpo (approved):
Added few questions to make sure I understand. 

The fix works and I am able to run the `lambdas` and `state machine` fine. I did notice the `api-gw` log did not report anything.

lib/aspects/user-only-permission-boundary.ts
kylelaker-ccpo: The CDK itself leverages the [Visitor pattern](https://refactoring.guru/design-patterns/visitor) through [Aspects](https://docs.aws.amazon.com/cdk/v2/guide/aspects.html). So we don't call `visit` but it gets applied by the Aspects framework.

jeffsegal-ccpo: I think Julius was asking whether it the aspect was actually applied during `synth` or `deploy`.

juliusfitzhugh-ccpo: Thanks for pointing out documentation that helps clear things up.

PR #678, opened 2022-03-15
Implement 'Result Lambda' and 'Result Queue'

jeffsegal-ccpo:
Adding ResultFunction to Provisioning Step Function workflow. Its job is simply to take the results of the CSP Invocation step and put them onto a SQS Queue. 

Infrastructure added:
* NodeJs Lambda Function (`ResultFunction`)
* SQS Queue (`ProvisioningQueue`)
* IAM Policies for controlling access to the queue

Includes one infrastructure test and one code test.

Ticket: AT-7056

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,5 +7,6 @@ node_modules
 .cdk.staging
 cdk.out
 cdk.context.json
+.idea
 
 coverage/
--- a/.gitignore
+++ b/.gitignore
@@ -7,7 +7,6 @@ node_modules
 .cdk.staging
 cdk.out
 cdk.context.json
-.idea
 
 coverage/
 
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,5 @@ cdk.context.json
 .idea
 
 coverage/
+
+.idea
--- a/api/provision/mock-invocation-fn.ts
+++ b/api/provision/mock-invocation-fn.ts
@@ -1,9 +1,8 @@
-import { CspInvocation, cspInvocationSchema, CspResponse } from "../../models/provisioning-jobs";
+import { provisionRequestSchema, CspResponse, ProvisionRequest } from "../../models/provisioning-jobs";
 import { errorHandlingMiddleware } from "../../utils/middleware/error-handling-middleware";
 import middy from "@middy/core";
 import validator from "@middy/validator";
 import JSONErrorHandlerMiddleware from "middy-middleware-json-error-handler";
-import { CloudServiceProvider } from "../../models/cloud-service-providers";
 import { REQUEST_BODY_INVALID } from "../../utils/errors";
 import { ValidationErrorResponse } from "../../utils/response";
 
@@ -16,9 +15,13 @@ import { ValidationErrorResponse } from "../../utils/response";
  * @param stateInput - Csp Invocation request
  * @return - CspResponse or throw error for 500 or above
  */
-export async function baseHandler(stateInput: CspInvocation): Promise<CspResponse | ValidationErrorResponse> {
+export async function baseHandler(stateInput: ProvisionRequest): Promise<CspResponse | ValidationErrorResponse> {
   console.log("STATE INPUT: ", JSON.stringify(stateInput));
-  if (!stateInput || !stateInput.endpoint) {
+  if (!stateInput) {
+    return REQUEST_BODY_INVALID;
+  }
+  const cspInvocation = stateInput.cspInvocation;
+  if (cspInvocation === undefined || !cspInvocation.endpoint) {
     return REQUEST_BODY_INVALID;
   }
 
@@ -37,11 +40,10 @@ export async function baseHandler(stateInput: CspInvocation): Promise<CspRespons
   return cspResponse;
 }
 
-export function createCspResponse(request: CspInvocation): CspResponse {
-  const { endpoint } = request;
+export function createCspResponse(request: ProvisionRequest): CspResponse {
   let response: CspResponse;
-  switch (endpoint.slice(0, 23)) {
-    case CloudServiceProvider.CSP_A.uri:
+  switch (request.targetCsp.name) {
+    case "CSP_A":
       response = {
         code: 200,
         content: {
@@ -50,7 +52,7 @@ export function createCspResponse(request: CspInvocation): CspResponse {
       };
       console.log("Success response : " + JSON.stringify(response));
       return response;
-    case CloudServiceProvider.CSP_B.uri:
+    case "CSP_B":
       response = {
         code: 400,
         content: {
@@ -72,6 +74,6 @@ export function createCspResponse(request: CspInvocation): CspResponse {
 }
 
 export const handler = middy(baseHandler)
-  .use(validator({ inputSchema: cspInvocationSchema }))
+  .use(validator({ inputSchema: provisionRequestSchema }))
   .use(errorHandlingMiddleware())
   .use(JSONErrorHandlerMiddleware());
--- a/api/provision/mock-invocation-fn.ts
+++ b/api/provision/mock-invocation-fn.ts
@@ -21,7 +21,7 @@ export async function baseHandler(stateInput: ProvisionRequest): Promise<CspResp
     return REQUEST_BODY_INVALID;
   }
   const cspInvocation = stateInput.cspInvocation;
-  if (cspInvocation === undefined || !cspInvocation.endpoint) {
+  if (!cspInvocation?.endpoint) {
     return REQUEST_BODY_INVALID;
   }
 
--- a/api/provision/result-fn.test.ts
+++ b/api/provision/result-fn.test.ts
@@ -11,7 +11,7 @@ const sqsMock = mockClient(sqsClient);
 const withResponse = {
   ...provisioningBodyWithPayload,
   cspResponse: {
-    code: "200",
+    code: 200,
     content: {},
   },
 };
@@ -22,6 +22,7 @@ beforeEach(() => {
 
 describe("Validate input", () => {
   it("should accept input with cspResponse", async () => {
+    console.log(JSON.stringify(withResponse));
     const response = await handler(withResponse, {} as Context);
     console.log(response);
     expect(response).not.toBeInstanceOf(ValidationErrorResponse);
--- a/api/provision/result-fn.test.ts
+++ b/api/provision/result-fn.test.ts
@@ -4,34 +4,14 @@ import { mockClient } from "aws-sdk-client-mock";
 import { ValidationErrorResponse } from "../../utils/response";
 import { SendMessageCommand } from "@aws-sdk/client-sqs";
 import { sqsClient } from "../../utils/aws-sdk/sqs";
+import { provisioningBodyWithPayload } from "./start-provision-job.test";
 
 const sqsMock = mockClient(sqsClient);
 
-const initialState = {
-  jobId: "81b31a89-e3e5-46ee-acfe-75436bd14577",
-  userId: "21d18790-bf3e-4529-a361-460ee6d16e0b",
-  portfolioId: "b02e77d1-234d-4e3d-bc85-b57ca5a93952",
-  operationType: "ADD_PORTFOLIO",
-  targetCsp: "CSP_A",
-  targetNetwork: "NETWORK_1",
-  payload: {
-    name: "About to change to full PR",
-    operators: ["admin1@mail.mil", "superAdmin@mail.mil"],
-    fundingSources: [
-      {
-        taskOrderNumber: "1234567890123",
-        clinNumber: "9999",
-        popStartDate: "2021-07-01",
-        popEndDate: "2022-07-01",
-      },
-    ],
-  },
-} as any;
-
 const withResponse = {
-  ...initialState,
+  ...provisioningBodyWithPayload,
   cspResponse: {
-    code: 200,
+    code: "200",
     content: {},
   },
 };
@@ -41,13 +21,9 @@ beforeEach(() => {
 });
 
 describe("Validate input", () => {
-  it("should reject input without cspResponse", async () => {
-    const response = await handler(initialState, {} as Context);
-    expect(response).toBeInstanceOf(ValidationErrorResponse);
-  });
-
   it("should accept input with cspResponse", async () => {
     const response = await handler(withResponse, {} as Context);
+    console.log(response);
     expect(response).not.toBeInstanceOf(ValidationErrorResponse);
   });
 });
--- a/api/provision/result-fn.test.ts
+++ b/api/provision/result-fn.test.ts
@@ -1,0 +1,77 @@
+import { Context } from "aws-lambda";
+import { handler } from "./result-fn";
+import { mockClient } from "aws-sdk-client-mock";
+import { sqsClient } from "../../utils/aws-sdk/sqs";
+import { ValidationErrorResponse } from "../../utils/response";
+import { match } from "sinon";
+import { SendMessageCommand } from "@aws-sdk/client-sqs";
+
+const sqsMock = mockClient(sqsClient);
+
+const initialState = {
+  jobId: "81b31a89-e3e5-46ee-acfe-75436bd14577",
+  userId: "21d18790-bf3e-4529-a361-460ee6d16e0b",
+  portfolioId: "b02e77d1-234d-4e3d-bc85-b57ca5a93952",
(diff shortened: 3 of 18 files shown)


atat_provisioning_api.yaml
```diff
@@ -259,6 +259,4 @@ components:
           type: integer
         content:
           type: object
-          additionalProperties: true
-        request:
-          $ref: '#/components/schemas/CspInvocation'
\ No newline at end of file
+          additionalProperties: true
```
zachclark-ccpo, line 262: I see that `CspResponse` is now:
```
CspResponse:
      required:
        - code
        - content
        - request
      properties:
        code:
          type: integer
        content:
          type: object
          additionalProperties: true
```
`request` is required, but it is not defined in properties?
I see the `$ref` was deleted for it, but shouldn't we at least define it as an object in the yaml?

.gitignore
```diff
@@ -7,5 +7,8 @@ node_modules
 .cdk.staging
 cdk.out
 cdk.context.json
+.idea
 
 coverage/
+
+.idea
```
juliusfitzhugh-ccpo, line 14: I believe this is a duplicate `.idea`.

api/provision/result-fn.test.ts
```diff
@@ -0,0 +1,77 @@
+import { Context } from "aws-lambda";
+import { handler } from "./result-fn";
+import { mockClient } from "aws-sdk-client-mock";
+import { sqsClient } from "../../utils/aws-sdk/sqs";
+import { ValidationErrorResponse } from "../../utils/response";
+import { match } from "sinon";
```
juliusfitzhugh-ccpo, line 6: Was this supposed to be used?

api/provision/result-fn.test.ts
```diff
@@ -0,0 +1,77 @@
...
+    expect(response).toBeInstanceOf(ValidationErrorResponse);
+    expect((response as ValidationErrorResponse).statusCode).toBe(400);
+  });
+
+  it("should accept input with cspResponse", async () => {
+    const response = await handler(withResponse, {} as Context);
+    expect(response).not.toBeInstanceOf(ValidationErrorResponse);
+    expect((response as ValidationErrorResponse).statusCode).toBe(200);
+  });
+});
```
juliusfitzhugh-ccpo, line 56: Nice tests, I like how compact they are.

api/provision/result-fn.test.ts
jeffsegal-ccpo: woops, fixed

api/provision/result-fn.ts
```diff
@@ -0,0 +1,30 @@
...
+  const QUEUE_URL = process.env.PROVISIONING_QUEUE_URL ?? "";
+  console.log("Sending result message to " + QUEUE_URL);
+  await sqsClient.send(
+    new SendMessageCommand({
+      QueueUrl: QUEUE_URL,
+      MessageBody: JSON.stringify(stateInput),
+    })
+  );
+
+  return { ...stateInput, statusCode: 200 };
```
juliusfitzhugh-ccpo, line 25: I used the `statusCode` earlier as a placeholder on the `sample-fn`, so I don't think it is needed here.

Also, since a message is being sent to a `SQS` queue do we need to return anything from this function?

api/provision/result-fn.test.ts
```diff
@@ -0,0 +1,77 @@
...
+
+describe("Validate behavior", () => {
+  it("should return the complete step fn state", async () => {
+    const response = await handler(withResponse, {} as Context);
+    expect(response).toEqual(
+      expect.objectContaining({
+        ...withResponse,
+        statusCode: 200,
```
juliusfitzhugh-ccpo, line 64: Also, made a [comment](https://github.com/dod-ccpo/atat-web-api/pull/678/files#r827266591) about the `statusCode` which I believe relates to this `statusCode` and can probably be removed here too.

lib/constructs/provisioning-sfn-workflow.test.ts
```diff
@@ -0,0 +1,88 @@
...
+      Statement: [
+        {
+          Action: ["sqs:SendMessage", "sqs:GetQueueAttributes", "sqs:GetQueueUrl"],
+          Effect: "Allow",
+          Resource: {
+            "Fn::GetAtt": [
+              stack.resolve((provisioningSfn.provisioningJobsQueue.node.defaultChild as cdk.CfnResource).logicalId),
+              "Arn",
+            ],
+          },
```
kylelaker-ccpo, line 33: Idk how much cleaner it'd be to do this since it'd also require an import of `import * as sqs from "aws-cdk-lib/aws-sqs"`

```suggestion
          Resource: stack.resolve((provisioningSfn.provisioningJobsQueue.defaultChild as sqs.CfnQueue).attrArn),
```

lib/constructs/provisioning-sfn-workflow.ts
```diff
@@ -45,18 +46,36 @@ export interface IProvisioningWorkflow {
...
+      queueName: "ProvisioningJobsQueue",
+    });
+
     // Provisioning State machine functions
     const sampleFn = new lambdaNodeJs.NodejsFunction(scope, "SampleFunction", {
       entry: "api/provision/sample-fn.ts",
       // vpc: props.vpc,
     });
+    this.resultFn = new lambdaNodeJs.NodejsFunction(scope, "ResultFunction", {
+      functionName: "ResultFunction",
```
kylelaker-ccpo, line 70: Do we _really_ want to name this? Won't this result in us only being able to build this once?

lib/constructs/provisioning-sfn-workflow.ts
```diff
@@ -45,18 +46,36 @@ export interface IProvisioningWorkflow {
...
+  readonly provisioningJobsQueue: sqs.IQueue;
+  readonly resultFn: lambdaNodeJs.NodejsFunction;
 
-  constructor(scope: Construct, props: ProvisioningWorkflowProps) {
+  constructor(scope: Construct, id: string, props: ProvisioningWorkflowProps) {
+    super(scope, id);
     const { environmentName } = props;
+
+    this.provisioningJobsQueue = new sqs.Queue(scope, "ProvisioningJobsQueue", {
+      queueName: "ProvisioningJobsQueue",
```
kylelaker-ccpo, line 61: Do we _really_ want to name this? Won't this result in us only being able to build this once?

lib/constructs/provisioning-sfn-workflow.ts
```diff
@@ -45,18 +46,36 @@ export interface IProvisioningWorkflow {
...
+
     // Provisioning State machine functions
     const sampleFn = new lambdaNodeJs.NodejsFunction(scope, "SampleFunction", {
       entry: "api/provision/sample-fn.ts",
       // vpc: props.vpc,
     });
+    this.resultFn = new lambdaNodeJs.NodejsFunction(scope, "ResultFunction", {
+      functionName: "ResultFunction",
+      environment: {
+        PROVISIONING_QUEUE_URL: this.provisioningJobsQueue.queueUrl,
```
kylelaker-ccpo, line 72: We shouldn't need to know the actual queue name because of this.

lib/constructs/provisioning-sfn-workflow.ts
```diff
@@ -45,18 +46,36 @@ export interface IProvisioningWorkflow {
...
+      },
+      entry: "api/provision/result-fn.ts",
+      // vpc: props.vpc,
+    });
+    this.provisioningJobsQueue.grantSendMessages(this.resultFn);
+    this.resultFn.addEnvironment("PROVISIONING_QUEUE_URL", this.provisioningJobsQueue.queueUrl);
```
kylelaker-ccpo, line 78: Didn't we just set this as part of the `NodeJsFunctionProps` above? (Line 72)? Or am I misreading?

lib/constructs/provisioning-sfn-workflow.ts
```diff
@@ -45,18 +46,36 @@ export interface IProvisioningWorkflow {
  * composed to form a workflow for provisioning jobs from SNOW to the CSPs.
  *
  */
-export class ProvisioningWorkflow implements IProvisioningWorkflow {
+export class ProvisioningWorkflow extends Construct implements IProvisioningWorkflow {
```
kylelaker-ccpo, line 49: Good catch

kylelaker-ccpo (changes requested):

lib/constructs/provisioning-sfn-workflow.ts
kylelaker-ccpo: So to expand on this a bit, `functionName` maps to the `FunctionName` CloudFormation property, which specifies the `FunctionName` as part of the `CreateFunction` API call. The name is the unique identifier of the function (as part of the `resourceName` field of its ARN). Because `Invoke` and other Lambda APIs accept the `FunctionName` parameter to uniquely identify a function, it must be unique within a region within an account. The same is true for nearly all resources that allow specifying the physical name (S3 Buckets, SQS Queues, etc) -- the name is a _unique_ identifier.

If this code is left as-is, we will only be able to build our stack once per region per account. This will break our Sandbox account. But even more significantly, it means that we can only create one `ProvisioningWorkflow` construct per stack.

In general, we should strongly avoid specifying physical names for resources as they don't really matter. We should pass references to functions/queues/buckets not their names (if we do need to pass a string, we use its name attribute). Exceptions are made in our IAM stack so that role names are predictable. But anything else will cause issues for devs working in parallel.

api/provision/result-fn.ts
```diff
@@ -0,0 +1,30 @@
...
+import { errorHandlingMiddleware } from "../../utils/middleware/error-handling-middleware";
+
+/**
+ * Result Lambda - Takes Step Fn input and sends to SQS Queue to be consumed
+ * by a scheduled job in Service Now.
+ *
+ * @param stateInput - input to the state task that is processed
+ */
+
+export async function baseHandler(stateInput: Record<string, unknown>): Promise<unknown> {
```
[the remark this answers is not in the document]
jasonburkert-ccpo: My understanding is that this Result Lambda will only dump to SQS.  Perhaps that integration is a good idea if we don't have to transform or manipulate the inputs.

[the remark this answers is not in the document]
jasonburkert-ccpo: We are leveraging middy handling.

api/provision/result-fn.ts
```diff
@@ -0,0 +1,30 @@
+import { sqsClient } from "../../utils/aws-sdk/sqs";
+import { SendMessageCommand } from "@aws-sdk/client-sqs";
+import middy from "@middy/core";
+import validator from "@middy/validator";
+import { provisioningResponseSchema } from "../../models/provisioning-jobs";
+import { errorHandlingMiddleware } from "../../utils/middleware/error-handling-middleware";
+
+/**
+ * Result Lambda - Takes Step Fn input and sends to SQS Queue to be consumed
+ * by a scheduled job in Service Now.
```
jasonburkert-ccpo, line 10: Suggest including a top-level comment that documents components in a standard way.  Perhaps using names from Provisioning Sequence Diagram, and referencing other components in the sequence.

Something like...
```suggestion
/**
 * Result Lambda
 * Input: output of Provisioning Step Function
 * Output: result added to Result Queue (to be consumed by a scheduled job in ServiceNow)
```

api/provision/result-fn.test.ts
jasonburkert-ccpo: perhaps a git push is needed?

.gitignore
jeffsegal-ccpo: fixed

api/provision/result-fn.test.ts
jeffsegal-ccpo: yeah sorry, i tend to say that a thing is fixed the moment i change it locally and in this case got mega sidetracked...  :)

api/provision/result-fn.ts
jeffsegal-ccpo: OK I figured as much but didn't want to assume for sure. Have since removed it. And no the return value is going to be ignored but I went ahead and let it return `$`, effectively.

api/provision/result-fn.test.ts
jeffsegal-ccpo: Done

lib/constructs/provisioning-sfn-workflow.test.ts
jeffsegal-ccpo: Definitely a bit cleaner - added the suggestion and the test still passes.

lib/constructs/provisioning-sfn-workflow.ts
jeffsegal-ccpo: Ahh thanks - I've been head-down strictly in the unit tests on this so didn't consider that. Really just wanted a clean way to identify that the particular function in question had been created. Ended up specifying the Role instead: 

```
        Role: stack.resolve((provisioningSfn.resultFn.role?.node.defaultChild as CfnRole).attrArn),
```

lib/constructs/provisioning-sfn-workflow.ts
jeffsegal-ccpo: Ah yes, classic. I first set it programmatically, then realized that I could set it via the parameter and commented out the programmatic line, then went back the next day when I had forgotten all about it and said "weird, I wonder why it still worked with this line commented out...". :)

Fixed now...

api/provision/result-fn.test.ts
```diff
@@ -0,0 +1,77 @@
...
+
+beforeEach(() => {
+  sqsMock.reset();
+});
+
+describe("Validate input", () => {
+  it("should reject input without cspResponse", async () => {
+    const response = await handler(initialState, {} as Context);
+    expect(response).toBeInstanceOf(ValidationErrorResponse);
+    expect((response as ValidationErrorResponse).statusCode).toBe(400);
```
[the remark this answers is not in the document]
jeffsegal-ccpo: I'm just removing the code entirely since it was really just a placeholder.

lib/constructs/provisioning-sfn-workflow.ts
kylelaker-ccpo: Might be able to use `description` too if we specify one. But yeah, that seems reasonable.

kylelaker-ccpo (dismissed):

juliusfitzhugh-ccpo (commented):
Regression testing performed on the deployed `AT-7056` stack. The sfn and lambdas are working as expected with the refactoring. Added a few additional comments.

api/provision/mock-invocation-fn.ts
```diff
@@ -16,9 +15,13 @@ import { ValidationErrorResponse } from "../../utils/response";
...
  */
-export async function baseHandler(stateInput: CspInvocation): Promise<CspResponse | ValidationErrorResponse> {
+export async function baseHandler(stateInput: ProvisionRequest): Promise<CspResponse | ValidationErrorResponse> {
   console.log("STATE INPUT: ", JSON.stringify(stateInput));
-  if (!stateInput || !stateInput.endpoint) {
+  if (!stateInput) {
+    return REQUEST_BODY_INVALID;
+  }
+  const cspInvocation = stateInput.cspInvocation;
+  if (cspInvocation === undefined || !cspInvocation.endpoint) {
```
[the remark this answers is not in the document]
jeffsegal-ccpo: Sure

lib/constructs/provisioning-sfn-workflow.ts
```diff
@@ -31,20 +32,36 @@ export interface IProvisioningWorkflow {
...
+  readonly provisioningJobsQueue: sqs.IQueue;
+  readonly resultFn: lambdaNodeJs.NodejsFunction;
 
-  constructor(scope: Construct, props: ProvisioningWorkflowProps) {
+  constructor(scope: Construct, id: string, props: ProvisioningWorkflowProps) {
+    super(scope, id);
     const { environmentName } = props;
+
+    this.provisioningJobsQueue = new sqs.Queue(scope, "ProvisioningJobsQueue", {
+      queueName: "ProvisioningJobsQueue",
```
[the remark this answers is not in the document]
jeffsegal-ccpo: Fixed

kylelaker-ccpo (approved):
Sounds good. We'll discuss my other comments about `CloudServiceProvider` again in the future

juliusfitzhugh-ccpo (approved):
Looks good!

PR #679, opened 2022-03-15
Implement 'Invocation Lambda' (mock)

juliusfitzhugh-ccpo:
Implement a lambda to mock invoking the CSP API and return different results 
based on the `CSP` provided in the request. 

- Implement `mock-lambda`
- Refactor how retries are handled in the `sfn-workflow`.
- Refactor types based on https://github.com/dod-ccpo/atat-web-api/pull/638#discussion_r826973193

Ticket: AT-7055

### DIFF
--- a/api/provision/mock-invocation-lambda.test.ts
+++ b/api/provision/mock-invocation-lambda.test.ts
@@ -1,0 +1,91 @@
+import { Context } from "aws-lambda";
+import { handler } from "./mock-invocation-lambda";
+import { CloudServiceProvider, Network } from "../../models/cloud-service-providers";
+import { ProvisionRequestType } from "../../models/provisioning-jobs";
+import { OtherErrorResponse, ValidationErrorResponse } from "../../utils/response";
+import { transformProvisionJob } from "./start-provision-job";
+
+const fundingSources = [
+  {
+    taskOrderNumber: "1234567890123",
+    clin: "9999",
+    popStartDate: "2021-07-01",
+    popEndDate: "2022-07-01",
+  },
+];
+const operators = ["admin1@mail.mil", "superAdmin@mail.mil"];
+const provisioningBodyNoPayload = {
+  jobId: "81b31a89-e3e5-46ee-acfe-75436bd14577",
+  userId: "21d18790-bf3e-4529-a361-460ee6d16e0b",
+  portfolioId: "b02e77d1-234d-4e3d-bc85-b57ca5a93952",
+  operationType: ProvisionRequestType.ADD_OPERATORS,
+  targetCsp: CloudServiceProvider.CSP_A.name,
+  targetNetwork: Network.NETWORK_1,
+};
+
+describe("Successful invocation of mock CSP function", () => {
+  it("should return 200 when CSP A provided in the request", async () => {
+    const stateInput = transformProvisionJob({
+      ...provisioningBodyNoPayload,
+      targetCsp: "CSP_A",
+      payload: {
+        name: "Mock A Invocation Portfolio",
+        fundingSources,
+        operators,
+      },
+    }) as any;
+    const response = await handler(stateInput, {} as Context);
+    console.log(response);
+    expect(response?.code).toBe(200);
+  });
+});
+
+describe("Failed invocation operations", () => {
+  it("should return a 400 when CSP_B is provided in the request", async () => {
+    const stateInput = transformProvisionJob({
+      ...provisioningBodyNoPayload,
+      targetCsp: "CSP_B",
+      payload: {
+        name: "Mock B Invocation Portfolio",
+        fundingSources,
+        operators,
+      },
+    }) as any;
+    const response = await handler(stateInput, {} as Context);
+    console.log(response);
+    expect(response?.code).toBe(400);
+  });
+  it("should return a 400 when additional payload property due to validation error", async () => {
+    const stateInput = transformProvisionJob({
+      ...provisioningBodyNoPayload,
+      targetCsp: "CSP_A",
+      payload: {
+        random: "property",
+        fundingSources,
+        operators,
+      },
+    } as any);
+    const response = await handler(stateInput, {} as Context);
+    console.log(response);
+    expect(response).toBeInstanceOf(ValidationErrorResponse);
+    expect(response.statusCode).toBe(400);
+  });
+  it("should throw a 500 error when a CSP internal error occurs", async () => {
+    const stateInput = transformProvisionJob({
+      ...provisioningBodyNoPayload,
+      targetCsp: "CSP_C",
+      payload: {
+        name: "Mock C Invocation Portfolio",
+        fundingSources,
+        operators,
+      },
+    }) as any;
+    expect(async () => await handler(stateInput, {} as Context)).rejects.toThrow(JSON.stringify({ code: 500, content: { some: "internal error"}, payload: stateInput}))
+  });
+  it("should return a 400 when null", async () => {
+    const response = await handler(undefined, {} as Context, () => null);
+    console.log(response);
+    expect(response).toBeInstanceOf(OtherErrorResponse);
+    expect(response.statusCode).toBe(400);
+  });
+});
--- a/api/provision/mock-invocation-lambda.test.ts
+++ b/api/provision/mock-invocation-lambda.test.ts
@@ -34,7 +34,7 @@ describe("Successful invocation of mock CSP function", () => {
         operators,
       },
     }) as any;
-    const response = await handler(stateInput, {} as Context);
+    const response: any = await handler(stateInput, {} as Context);
     console.log(response);
     expect(response?.code).toBe(200);
   });
@@ -51,7 +51,7 @@ describe("Failed invocation operations", () => {
         operators,
       },
     }) as any;
-    const response = await handler(stateInput, {} as Context);
+    const response: any = await handler(stateInput, {} as Context);
     console.log(response);
     expect(response?.code).toBe(400);
   });
@@ -65,7 +65,7 @@ describe("Failed invocation operations", () => {
         operators,
       },
     } as any);
-    const response = await handler(stateInput, {} as Context);
+    const response: any = await handler(stateInput, {} as Context);
     console.log(response);
     expect(response).toBeInstanceOf(ValidationErrorResponse);
     expect(response.statusCode).toBe(400);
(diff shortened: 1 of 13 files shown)


api/provision/mock-invocation-lambda.test.ts
```diff
@@ -0,0 +1,93 @@
...
+  it("should return 200 when CSP A provided in the request", async () => {
+    const stateInput = transformProvisionJob({
+      ...provisioningBodyNoPayload,
+      targetCsp: "CSP_A",
+      payload: {
+        name: "Mock A Invocation Portfolio",
+        fundingSources,
+        operators,
+      },
+    }) as any;
```
kylelaker-ccpo, line 36: We probably really do want to avoid all this `any`.

api/provision/mock-invocation-lambda.test.ts
```diff
@@ -0,0 +1,93 @@
...
+    const response: any = await handler(stateInput, {} as Context);
+    console.log(response);
+    expect(response?.code).toBe(200);
```
kylelaker-ccpo, line 39:
```suggestion
    const response: any = await handler(stateInput, {} as Context);
    expect(response?.code).toBe(200);
```

api/provision/mock-invocation-lambda.test.ts
```diff
@@ -0,0 +1,93 @@
...
+      },
+    }) as any;
+    expect(async () => await handler(stateInput, {} as Context)).rejects.toThrow(
+      JSON.stringify({ code: 500, content: { some: "internal error" }, payload: stateInput })
+    );
+  });
+  it("should return a 400 when null", async () => {
+    const response: any = await handler(undefined as any, {} as Context, () => null);
+    console.log(response);
+    expect(response).toBeInstanceOf(OtherErrorResponse);
```
kylelaker-ccpo, line 90:
```suggestion
    const response: any = await handler(undefined as any, {} as Context, () => null);
    expect(response).toBeInstanceOf(OtherErrorResponse);
```

utils/middleware/error-handling-middleware.ts
```diff
@@ -3,15 +3,23 @@ import { APIGatewayProxyResult } from "aws-lambda";
...
+import { StepFunctionRequestEvent, RequestBodyType, CspInvocation, CspResponse } from "../../models/provisioning-jobs";
 
-export const errorHandlingMiddleware = (): middy.MiddlewareObj<ILambdaEvent, APIGatewayProxyResult> => {
-  const onError: middy.MiddlewareFn<ILambdaEvent, APIGatewayProxyResult> = async (
-    request
-  ): Promise<ValidationErrorResponse | void> => {
-    const error = serializeError(request.error)!;
+export const errorHandlingMiddleware = (): middy.MiddlewareObj<
+  StepFunctionRequestEvent<RequestBodyType> | CspInvocation,
+  APIGatewayProxyResult | CspResponse | ValidationErrorResponse
```
kylelaker-ccpo, line 10: This feels like a lot of possible types. Can you help me understand what's going on here?

utils/middleware/error-handling-middleware.ts
```diff
@@ -3,15 +3,23 @@ import { APIGatewayProxyResult } from "aws-lambda";
...
+  const onError: middy.MiddlewareFn<
+    StepFunctionRequestEvent<RequestBodyType> | CspInvocation,
+    APIGatewayProxyResult | CspResponse | ValidationErrorResponse
+  > = async (request): Promise<ValidationErrorResponse | void> => {
+    const error = serializeError(request.error!);
     const errorMessage = error.message;
 
+    if (error.name === "MockCspApiError") {
+      throw request.error;
+    }
```
kylelaker-ccpo, line 21: Why is this special? Why does it get re-thrown?

jasonburkert-ccpo: Should something happen prior to the `throw`?

api/provision/mock-invocation-lambda.test.ts
juliusfitzhugh-ccpo: Yeah, I can't argue with that.

utils/middleware/error-handling-middleware.ts
kylelaker-ccpo: I guess my question is more -- why are we `throw`ing? Isn't the goal to _handle_ the errors? Why is this one special to be `throw`n rather than being converted to a returned error type?

utils/middleware/error-handling-middleware.ts
juliusfitzhugh-ccpo: @kylelaker-ccpo This comes from the different inputs and outputs being used for the `baseHandler`s. The `start-provision-job` lambda is initiated from a `POST` request, which output is sent to the step function `mock-invocation-lambda` and is of a different shape. The various types allows for the middlewares to handle the different inputs and outputs without throwing a typescript error since it is based on the inputs and outputs being used by the `baseHandler`.
- `StepFunctionRequestEvent<RequestBodyType> | CspInvocation` are the allowed inputs 
- `APIGatewayProxyResult | CspResponse | ValidationErrorResponse` are the allowed outputs

I guess a better way might be something like 
```typescript
type middlwareInputs = StepFunctionRequestEvent<RequestBodyType> | CspInvocation
type middlewareOutputs = APIGatewayProxyResult | CspResponse | ValidationErrorResponse

export const errorHandlingMiddleware = (): middy.MiddlewareObj<middlewareInputs, middlewareOutputs>
```

kylelaker-ccpo: Oh oops, I missed the comma; sorry! But yeah; that would be a lot cleaner! Maybe with semicolon?

juliusfitzhugh-ccpo: No problem. Oh yeah semicolon for sure

api/provision/mock-invocation-lambda.test.ts
juliusfitzhugh-ccpo: @kylelaker-ccpo I've removed the amount of `any` being used. The `any` being used by the responses was causing typescript to complain due to one of the types in the union not having a specific property even though one of the types have the property. Didn't think that was a problem, but maybe I am misunderstanding something. 
```typescript
Property 'code' does not exist on type 'CspResponse | ValidationErrorResponse'.
  Property 'code' does not exist on type 'ValidationErrorResponse'.
```

api/provision/start-provision-job.ts
```diff
@@ -54,7 +54,7 @@ export async function baseHandler(event: ILambdaEvent): Promise<APIGatewayProxyR
  * @param request - provisioning request from SNOW
  * @returns - transformed request to send to the targeted CSP
  */
-function transformProvisionJob(request: ProvisionRequest): CspInvocation {
+export function transformProvisionJob(request: ProvisionRequest): CspInvocation {
```
jasonburkert-ccpo, line 57: Suggest renaming this.  I think `Job` has connotations that are misleading, as though this involves the provisioning jobs queue which is further downstream in our sequence.

```suggestion
export function translateProvisionRequest(request: ProvisionRequest): CspInvocation {
```

utils/middleware/error-handling-middleware.ts
jasonburkert-ccpo: In that case...
```suggestion
    if (error.name === "MockCspApiError") {
      // force state machine task to retry
      throw request.error;
    }
```

jasonburkert-ccpo (changes requested):

api/provision/mock-invocation-lambda.test.ts
kylelaker-ccpo: Yeah, so the property has to exist on _all_ the possible types. `any` kinda bypasses those checks (basically, a way to say, "no `tsc`, I _do_ know what I'm doing). So of `CspResonse` and `ValidationErrorResponse`, that error implies that `ValidationErrorResponse` doesn't have a `code` property; so we'd get either `undefined` or an error if we tried to access it.

So we'd have to do at least one of the following:
 - ensure _all_ possible types have a `code` property (and any other properties we try to access)
 - narrow the type around attribute access to "prove" to typescript that it's safe (so an `instanceof` or `typeof` check, or maybe even checking `if (it.code) { ... }`)
 - create a `IStatusCodeable` (very bad name, plz dont use) type that requires a `code` and accept that -- this is the worst of the options in _this_ use case

api/provision/mock-invocation-lambda.test.ts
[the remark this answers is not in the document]
juliusfitzhugh-ccpo: Yeah, totally slipped my mind.

utils/middleware/check-csp-portfolio-id.ts
```diff
@@ -1,10 +1,22 @@
...
 import { APIGatewayProxyResult } from "aws-lambda";
-import { ILambdaEvent, ProvisionRequestType } from "../../models/provisioning-jobs";
+import {
+  StepFunctionRequestEvent,
+  ProvisionRequestType,
+  RequestBodyType,
+  CspResponse,
+} from "../../models/provisioning-jobs";
 import createError from "http-errors";
+import { ValidationErrorResponse } from "../response";
```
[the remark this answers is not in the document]
juliusfitzhugh-ccpo: Good point, added a comment. I agree, it helps give some context for each middleware.

api/provision/start-provision-job.ts
juliusfitzhugh-ccpo: Yeah, that makes sense, removing `Job` makes it clearer.

models/provisioning-jobs.ts
```diff
@@ -45,11 +46,17 @@ export interface CspInvocation {
...
+export interface StepFunctionRequestEvent<T> {
+  body: T;
+  requestContext: APIGatewayEventRequestContext;
 }
+export type RequestBodyType = ProvisionRequest;
 
+export interface CspResponse {
+  code: number;
+  content: unknown;
+  request: CspInvocation;
```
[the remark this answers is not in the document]
juliusfitzhugh-ccpo: Removed.

utils/middleware/check-csp-portfolio-id.ts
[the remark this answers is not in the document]
jasonburkert-ccpo: Nice change.

jasonburkert-ccpo (approved):

jasonburkert-ccpo:
Please be sure @kylelaker-ccpo comments are addressed prior to merging.

api/provision/mock-invocation-lambda.test.ts
juliusfitzhugh-ccpo: Went with option 2. Thanks for explaining. I have a better understanding, but realize I need to be more careful using `union` in typescript.

PR #680, opened 2022-03-15
Add esbuild package as dev dependency

jasonburkert-ccpo:
This will prevent the attempted use of docker for CDK bundling


jasonburkert-ccpo:
I ran into a problem where I had neither `docker` nor `esbuild` installed and `npm run test` would fail when CDK attempted to generate assets.

juliusfitzhugh-ccpo (approved):
We were using `esbuild` previously, so that seems to make sense.

kylelaker-ccpo:
I like this much better than requiring Docker (which hasn't gone well for us in the past :D)

PR #695, opened 2022-03-22
Use FIPS endpoints for all AWS APIs

kylelaker-ccpo:
The FIPS-compliant endpoints in the AWS SDK's `endpoints.json` file have
been updated to properly reflect the correct hostname.


kylelaker-ccpo:
@juliusfitzhugh-ccpo do you think we could work together on validating this some time? Basically, we'll want to make sure that Lambda can still invoke SQS (and SFN) API calls.

juliusfitzhugh-ccpo (approved):
Looks good! Things are working as expected. Sfn and lambdas are running, and messages are being sent to SQS.

PR #700, opened 2022-03-22
Fix IAM permissions

kylelaker-ccpo:
- Fix access to Step Functions
- Allow developers to read SQS Queues


jasonburkert-ccpo (approved):

juliusfitzhugh-ccpo (approved):
IAM permissions working as expected can see executions with Sfn and have access to SQS messages.

PR #705, opened 2022-03-24
Use the iam-floyd library for IAM policy creation

kylelaker-ccpo:
The [`iam-floyd`](https://github.com/udondan/iam-floyd) package provides
a fluent builder-like interface for create IAM policies. This gives them
a more human-readable format as well as nice abstractions for certain
things (like `ifEncrypted()`). This and the `ifX` methods make it a bit
easier to naturally attach conditions to IAM policies. In the end, we
still get an `iam.PolicyStatement` object (through the `cdk-iam-floyd`
package) which the CDK can naturally turn into JSON/CloudFormation.

### DIFF
--- a/lib/atat-iam-stack.ts
+++ b/lib/atat-iam-stack.ts
@@ -1,6 +1,7 @@
 import * as cdk from "aws-cdk-lib";
 import * as iam from "aws-cdk-lib/aws-iam";
 import * as custom from "aws-cdk-lib/custom-resources";
+import * as statement from "cdk-iam-floyd";
 
 import { Construct } from "constructs";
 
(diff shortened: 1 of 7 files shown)


kylelaker-ccpo:
oof our tests really dont like us changing the listing of IAM permissions at all...

lib/atat-iam-stack.ts
```diff
@@ -78,62 +54,38 @@ export class AtatIamStack extends cdk.Stack {
...
+            cdk.Aws.PARTITION
+          ),
+        // Additionally, values under the `cdk-bootstrap/` namespace for the qualifier may be necessary
+        // for the CDK to understand the environment and do a deployment
+        new statement.Ssm()
+          .toGetParameter()
+          .onParameter(
+            `cdk-bootstrap/${DEFAULT_BOOSTRAP_QUALIFIER}/*`,
+            cdk.Aws.ACCOUNT_ID,
+            undefined,
```
zachclark-ccpo, line 77: Do you need to pass `undefined` here because you can't set the region?

kylelaker-ccpo: Correct! We could probably also pass `"*"` but `undefined` is the default value. The methods take positional arguments, not a structure/object with named fields; therefore, because we want to later pass a value for `partition`, we must pass something for `region`. So passing `undefined` is basically the same as not passing anything at all.

zachclark-ccpo (approved):

lib/atat-iam-stack.ts
```diff
@@ -78,62 +54,38 @@ export class AtatIamStack extends cdk.Stack {
...
+            cdk.Aws.PARTITION
+          ),
+        // Within ATAT, developers may may need to read from various queues in order to debug or
```
juliusfitzhugh-ccpo, line 80:
```suggestion
        // Within ATAT, developers may need to read from various queues in order to debug or
```

juliusfitzhugh-ccpo (approved):
I like the refactoring with `cdk-iam-floyd`, I see why it is popular.

lib/atat-iam-stack.ts
```diff
@@ -31,24 +32,9 @@ export class AtatIamStack extends cdk.Stack {
...
-          resources: [`arn:${cdk.Aws.PARTITION}:states:*:${cdk.Aws.ACCOUNT_ID}:*:*`],
-        }),
-        new iam.PolicyStatement({
-          sid: "XRayReadAccess",
-          effect: iam.Effect.ALLOW,
-          actions: ["xray:BatchGetTraces", "xray:Get*", "xray:List*"],
-          resources: ["*"],
-        }),
+        new statement.Apigateway().toGET().onRestApis(undefined, cdk.Aws.PARTITION),
+        new statement.States().allReadActions(),
```
jasonburkert-ccpo, line 36: It's nice that you don't have to cobble together the read actions yourself

jasonburkert-ccpo (approved):
Really an remarkable change.  Looks nice.  Good refactoring.

As long as you are satisfied that the resulting `iam.PolicyStatement`s achieved through the use of `cdk-iam-floyd` are functionally equivalent, I think this a great change.  A nice forward looking use of a early version of a package.  It's been in development for a couple years.  Any concern about having just for contributors?

kylelaker-ccpo:
> It's been in development for a couple years. Any concern about having just four contributors?

I'm not super concerned about that. In a review, it seems like most of the updates to add new policy actions and conditions are automated, and we can always fall back to writing raw IAM templates in the future if the need arose. And at least at this stage (and for the semi-near future), it's not too heavy of a revert.

PR #721, opened 2022-03-29
Add initial pipeline

kylelaker-ccpo:
Uses `aws-cdk-lib/pipelines` to define `atat-pipeline-stack`

### DIFF
--- a/bin/atat-web-api.ts
+++ b/bin/atat-web-api.ts
@@ -3,6 +3,7 @@ import "source-map-support/register";
 import * as cdk from "aws-cdk-lib";
 // import { NIST80053R4Checks, NIST80053R5Checks } from "cdk-nag";
 import { AtatWebApiStack } from "../lib/atat-web-api-stack";
+import { AtatPipelineStack } from "../lib/atat-pipeline-stack";
 import { GovCloudCompatibilityAspect } from "../lib/aspects/govcloud-compatibility";
 import { RemovalPolicySetter } from "../lib/aspects/removal-policy";
 import * as utils from "../lib/util";
@@ -31,4 +32,8 @@ if (utils.isPossibleTemporaryEnvironment(environmentId)) {
 // cdk.Aspects.of(app).add(new NIST80053R5Checks({ verbose: true }));
 cdk.Aspects.of(app).add(new GovCloudCompatibilityAspect());
 
-const apiStack = new AtatWebApiStack(app, `${environmentName}AtatWebApiStack`, { environmentName });
+const apiStack = new AtatPipelineStack(app, `Pipeline`, {
+  environmentName,
+  branch: "feature/cdk-pipelines",
+  githubPatSecretPath: "auth/github/pat-7176", // pragma: allowlist secret
+});
--- a/lib/atat-pipeline-stack.ts
+++ b/lib/atat-pipeline-stack.ts
@@ -1,0 +1,41 @@
+import * as cdk from "aws-cdk-lib";
+import * as pipelines from "aws-cdk-lib/pipelines";
+import * as ssm from "aws-cdk-lib/aws-ssm";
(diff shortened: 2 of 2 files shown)


lib/atat-pipeline-stack.ts
```diff
@@ -0,0 +1,41 @@
...
+  }
+}
+
+export class AtatPipelineStack extends cdk.Stack {
+  constructor(scope: Construct, id: string, props: AtatPipelineStackProps) {
+    super(scope, id, props);
+
+    const pipeline = new pipelines.CodePipeline(this, "Pipeline", {
+      synth: new pipelines.ShellStep("Synth", {
+        input: pipelines.CodePipelineSource.gitHub("kylelaker-ccpo/atat-web-api", "feature/cdk-pipelines", {
```
kylelaker-ccpo, line 32: Use `props.branch` here. Consider making a `props.repo`. Or even:

```ts
export interface GitHubRepository {
  name: string;
  branch: string;
}
```

And add a field to `props` for that, or something like it.

bin/atat-web-api.ts
```diff
@@ -31,4 +32,8 @@ if (utils.isPossibleTemporaryEnvironment(environmentId)) {
 // cdk.Aspects.of(app).add(new NIST80053R5Checks({ verbose: true }));
 cdk.Aspects.of(app).add(new GovCloudCompatibilityAspect());
 
-const apiStack = new AtatWebApiStack(app, `${environmentName}AtatWebApiStack`, { environmentName });
+const apiStack = new AtatPipelineStack(app, `Pipeline`, {
+  environmentName,
+  branch: "feature/cdk-pipelines",
```
kylelaker-ccpo, line 37: This will need to be `develop` eventually.

bin/atat-web-api.ts
```diff
@@ -31,4 +32,8 @@ if (utils.isPossibleTemporaryEnvironment(environmentId)) {
...
+  githubPatSecretPath: "auth/github/pat-7176", // pragma: allowlist secret
```
kylelaker-ccpo, line 38: This name is temporary. Ideally, we'd use something that doesn't anger the secret scanner. I want to remove the `// pragma: allowlist secret` because it sets a bad precedent.

bin/atat-web-api.ts
```diff
@@ -31,4 +32,8 @@ if (utils.isPossibleTemporaryEnvironment(environmentId)) {
 // cdk.Aspects.of(app).add(new NIST80053R5Checks({ verbose: true }));
 cdk.Aspects.of(app).add(new GovCloudCompatibilityAspect());
 
-const apiStack = new AtatWebApiStack(app, `${environmentName}AtatWebApiStack`, { environmentName });
+const apiStack = new AtatPipelineStack(app, `Pipeline`, {
```
kylelaker-ccpo, line 35: Will we only ever have one stack named `Pipeline` within the account? Maybe?

kylelaker-ccpo:
Tests failed since this comes from a fork (which I needed to do for testing, though retrospectively, maybe I didnt)

lib/atat-pipeline-stack.ts
kylelaker-ccpo, line 32: CDK recommends using the `.connection` function but we can't do that because it leverages CodeStar and we're targeting GovCloud (where CodeStar is not available)

bin/atat-web-api.ts
jasonburkert-ccpo: What's the disadvantage of something like `AtatWebApiPipeline` instead?  Poor readability?

juliusfitzhugh-ccpo:
Looks good. I know this is a POC, so other than some IAM policy updates for others at some point to see the pipeline this looks really promising.

PR #739, opened 2022-04-06
Create deployment pipeline

kylelaker-ccpo:
This adopts a pipeline to deploy changes to non-sandbox environments and
solidifies the idea of what is in fact a "sandbox" environment by simply
leveraging a new context value. The pipeline is implemented using the
CDK's native CDK Pipelines construct library (`aws-cdk-lib.pipelines`).
This uses CodePipeline under the hood to perform the deployment.

Through this, IAM is significantly refactored to remove no-longer-needed
values and permissions (and also adds permissions to view the pipeline)
and moves the "base infrastructure" to be managed by the pipeline. This
sort of starts to treat sandbox environments as a special case rather
than a "real" environment; this more accurately represents how our
environments actually look (where sandbox is the only "weird" account).

We actually remove quite a bit of code here. And a lot of the "new" code
is actually just code that has been moved from other places. For the
most part, the Pipeline stack itself and the `bin/` folder are the only
places that had much net-new code besides the README, which has been
entirely rewritten from the ground up. The changes there should likely
be reviewed carefully. Additionally, the steps developers need to take
to test their environments locally have changed; the details for those
changes are in the new README.

A temporary workaround has been put in place to ensure that we don't
break access to API Gateway from ServiceNow or the internet. We have
more work to do for network transit before we can make that work
perfectly.

The changes here also mean we no longer have custom values in our
`.secrets.baseline` (for now).

Implementation details for how we create and manage the GitHub PAT will
not be covered in this repository and (for now) we will keep those
details internal.

Ticket: AT-7204


lib/atat-pipeline-stack.ts
```diff
@@ -0,0 +1,47 @@
...
+}
+
+export class AtatApplication extends cdk.Stage {
+  constructor(scope: Construct, id: string, props: cdk.StageProps & AtatProps) {
+    super(scope, id, props);
+    const iam = new AtatIamStack(this, `${props.environmentName}Iam`);
+    const net = new AtatNetStack(this, `${props.environmentName}Net`, { vpcCidr: props.vpcCidr });
+    const atat = new AtatWebApiStack(this, "Atat", {
+      environmentName: props.environmentName,
+      network: net,
```
juliusfitzhugh-ccpo, line 27: Does a `net` stack still get created if `props.vpcCidr` is `undefined`? It looks like a `vpcCidr` is not being supplied for the `app`, but I guess a default value might be provided by AWS. 

Also, do we still plan to have a shared vpc in sandbox, since we were having issues previously with running out of vpcs?

lib/atat-pipeline-stack.ts
```diff
@@ -0,0 +1,47 @@
...
+    });
+  }
+}
+
+export class AtatPipelineStack extends cdk.Stack {
```
juliusfitzhugh-ccpo, line 32: I'm still wrapping my head around the infrastructure tests, but do we need new tests to go along with the new pipeline infrastructure?

cdk.json
```diff
@@ -27,6 +27,9 @@
     "@aws-cdk/core:target-partitions": [
       "aws",
       "aws-us-gov"
-    ]
+    ],
```
juliusfitzhugh-ccpo, line 30: I don't think it makes a difference, but thought I'd ask. On line 14 is `yarn.lock` are we using `yarn` somewhere?  Also, on line 16 is `test`, we are no longer using a `test` folder anymore, so `test` is not needed is it?

juliusfitzhugh-ccpo (commented):
This looks good and I'm excited to see this in action. I asked some questions, but they are more geared towards making sure I understand things. 

Once the pipeline is deployed, then all the following merges into `develop` will keep the application updated without the need to manually deploy right? Or will we have another step of approval before that happens?

lib/atat-pipeline-stack.ts
kylelaker-ccpo: It _should_ it will just use the default CIDR range. But good catch! I'll make sure we have a way to propagate that down to the net stack. The App stack shouldn't need it.

lib/atat-pipeline-stack.ts
kylelaker-ccpo: Yeah, we do. That's why Sonar is grumpy right now :)

cdk.json
kylelaker-ccpo: Oh yeah, everything in that block above is part of the default template from `cdk init`. I think it's for the `cdk watch` command which we don't really use (or at least I don't _think_ any of us do but I'd love to see someone try it). So are the default feature flags in the context.

bin/atat-web-api.ts
```diff
@@ -18,17 +20,26 @@ if (!utils.isString(environmentParam)) {
...
 // These Aspects will not be enabled until more base infrastructure around KMS, etc
 // is in place.
 // cdk.Aspects.of(app).add(new NIST80053R4Checks({ verbose: true }));
 // cdk.Aspects.of(app).add(new NIST80053R5Checks({ verbose: true }));
 cdk.Aspects.of(app).add(new GovCloudCompatibilityAspect());
 
-const apiStack = new AtatWebApiStack(app, `${environmentName}AtatWebApiStack`, { environmentName });
+// Sandbox environments get just the API, all other environments get a pipeline to
+// actually handle the deployments.
+if (isSandbox) {
```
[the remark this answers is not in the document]
kylelaker-ccpo: Yep! That gets covered in the second commit message but I should have elevated that info into the PR body text as well. I'll work on updating the PR in a bit:

> This configures the application pipeline to deploy a "full" environment. It also allows for building "sandboxes" disconnected from a VPC and the rest of the environment. There will still be a "Sandbox" environment to build some guardrails for IAM (and the pipeline itself for that) but it's really no big deal (hopefully). Ideally, the only developer-facing change to deploy a sandbox environment would be the need to now add `-c atat:sandbox=true` to CLI invocations for diff/deploy.

lib/atat-pipeline-stack.ts
kylelaker-ccpo: Also... as far as VPC sharing, I was almost thinking that we just wouldn't care about the VPC in Sandbox environments? And make it a dev environment problem? And just rely on the "default" networking capabilities of the various services/resources in sandbox? Maybe that's a little too ambitious.

Because I _do_ still plan to deploy a `SandboxPipeline` (so that it can manage IAM), there actually _would_ be a Sandbox VPC (and IAM and API) that would track `develop` but could be configured to track a feature branch if that'd be useful for a ticket or for testing.

juliusfitzhugh-ccpo: Yeah, I don't think we care too much about the VPC in sandbox. Plus, we are doing less in AWS, so maybe things will be better. The VPCs were just resource intensive before

Cool, that makes sense (our IAM is every evolving). It might be nice to have that flexibility to track a feature branch for testing purposes, but I'll leave that up to you.

cdk.json
juliusfitzhugh-ccpo: Oh okay. Yeah, it does look like it's for the `cdk watch` which can also use `hot-swap` that you mentioned previously. This will help speed up deployment, so if I get a chance I'll try to use it.

bin/atat-web-api.ts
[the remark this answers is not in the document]
juliusfitzhugh-ccpo: Got it, I didn't go over each commit this time so I missed it.

lib/atat-pipeline-stack.ts
```diff
@@ -0,0 +1,47 @@
...
+  constructor(scope: Construct, id: string, props: AtatPipelineStackProps) {
+    super(scope, id, props);
+
+    const pipeline = new pipelines.CodePipeline(this, "Pipeline", {
```
zachclark-ccpo, line 36: Since we are using s3, should we include:
`selfMutation: true` ?
v2 docs: https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.pipelines-readme.html#s3-error-access-denied

ShellStep docs explicitly setting `selfMutation` to `true` or `false`: https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.pipelines.ShellStep.html#example

lib/atat-web-api-stack.ts
```diff
@@ -3,22 +3,32 @@ import * as apigw from "aws-cdk-lib/aws-apigateway";
...
+    const { environmentName, network } = props;
 
-    const api = new AtatRestApi(this, "SampleApi");
+    const apiProps: AtatRestApiProps = {};
+    if (network) {
+      apiProps.vpcConfig = {
+        vpc: network.vpc,
+        interfaceEndpoint: network.endpoints.apigateway,
+      };
+    }
```
zachclark-ccpo, line 29: Should we throw an error here if `network` isn't provided?

lib/atat-pipeline-stack.ts
kylelaker-ccpo: https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.pipelines.CodePipeline.html#selfmutation

> selfMutation?
> Type: boolean (optional, default: true)

We should be good to go here.

lib/atat-net-stack.ts
```diff
@@ -87,16 +87,17 @@ export class AtatNetStack extends cdk.Stack {
...
+    this.endpoints = {
+      ...this.endpoints,
+      s3: s3Endpoint,
+      apigateway: apiGatewayEndpoint,
+      lambda: lambdaEndpoint,
+      logs: logsEndpoint,
+      sqs: sqsEndpoint,
+      xray: xrayEndpoint,
+      ssecrets: secretsManagerEndpoint,
+      sfn: stepFunctionsEndpoint,
```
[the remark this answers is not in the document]
kylelaker-ccpo: I don't know if I love it... I am playing around with this. We'll see if I end up reverting it but I like that it means we can map back to services.

lib/atat-pipeline-stack.ts
```diff
@@ -0,0 +1,47 @@
...
+      synth: new pipelines.ShellStep("Synth", {
+        input: pipelines.CodePipelineSource.gitHub(props.repository, props.branch, {
+          authentication: cdk.SecretValue.secretsManager(props.githubPatName),
```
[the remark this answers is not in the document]
kylelaker-ccpo: So right now, as I am developing, it's me/my account. I am going try to see if we can just provide an empty/invalid PAT since the repo is public. But otherwise, this is a conversation we'll need to come back to.

lib/atat-web-api-stack.ts
kylelaker-ccpo: We shouldn't. The better solution if we wanted to require `network` would to be to just... require `network` :) See https://github.com/dod-ccpo/atat-web-api/pull/739#discussion_r844507277 for more information.

lib/atat-pipeline-stack.ts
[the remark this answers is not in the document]
kylelaker-ccpo: Okay so "use a fake token" didnt work... we will have to use a real one. Let's sync up on the best way to handle that.

zachclark-ccpo (changes requested):
Please update the README with deployment instructions

lib/atat-iam-stack.ts
```diff
@@ -88,17 +90,6 @@ export class AtatIamStack extends cdk.Stack {
       statements: [new statement.Organizations().deny().allActions().onAllResources()],
     });
 
-    // Eventually, this should be restricted to the specific services in use for the application
-    // rather than allowing all IAM actions. This still theoretically will give anyone who can pass
-    // the role permission to do almost anything in the account; however, it at least requires that
-    // it happen via CloudFormation.
-    const cloudFormationExecutionRole = new iam.Role(this, "CloudFormationExecutionRole", {
```
jeffsegal-ccpo, line 95: Assuming I'm tracking this correctly, it's really nice to let CodePipeline handle some of this complexity.

README.md
```diff
@@ -1,40 +1,78 @@
...
 
- * `npm run build`   compile typescript to js
- * `npm run watch`   watch for changes and compile
- * `npm run test`    perform the jest unit tests
- * `cdk deploy`      deploy this stack to your default AWS account/region
- * `cdk diff`        compare deployed stack with current state
- * `cdk synth`       emits the synthesized CloudFormation template
+Generally, the application is implemented to operate in a stateless fashion (with some transient state) leveraging
+SQS Queues, Step Functions State Machines, and a REST API. The private, internal-facing API is referred to as HOTH
+and serves a helpful API for other internal implementation details of the ATAT client application. Currently, HOTH
```
jeffsegal-ccpo, line 11: Shall we define the acronym here? :)

jeffsegal-ccpo (approved):
README looks great. Excited to be moving in a more formal direction for CI/CD while still maintaining the niceties of the sandbox environments for devs.

lib/atat-iam-stack.ts
kylelaker-ccpo: In this case, it's actually CDKv2. This has been dead code for a few weeks now. The specifics for this role are now defined as part of the CDK Bootstrap stack. That becomes more useful especially when we move towards considering a cross-account setup.

README.md
kylelaker-ccpo: Done! Had to reflow the text a bit so the commit isn't quite as clean as just adding
the word.

PR #774, opened 2022-04-14
Make pipeline track `develop` by default

kylelaker-ccpo:
This couldn't be included in the initial PR because the pipeline was
already deployed and watching the feature branch. Had it switched to
watching develop before the pipeline was merged all sorts of wonkiness
could have occurred and I didn't really want to clean that up. This just
switches the default.


juliusfitzhugh-ccpo (approved):
Yay for pipeline for `develop`!

PR #778, opened 2022-04-14
Update to ESLint v8

kylelaker-ccpo:
This closes out the remaining outdated packages that we continue to be
notified about. Currently, we keep seeing Dependabot trying to update
`eslint-plugin-promise` to v6 but that adds/requires ESLint 8 support.
Unfortunately, `eslint-config-standard` v16 does not support ESLint and
will not support it until v17. It seems that v17 is ready for release
but they're holding off until external packages that they test
compatibility with are "working" with v17. Frankly, I don't think any of
the things holding those packages back are holding us back. There is
a pre-release version of `eslint-config-standard` v17 that we can adopt
today and start using to unblock other dependency updates.

We run the lint as part of every PR, so if there's a significant issue
with linting or building, we'll catch it quickly.


kylelaker-ccpo:
See https://github.com/standard/eslint-config-standard/issues/208 for the tracking issue for `eslint-config-standard` v17.

kylelaker-ccpo:
...Fascinating because an `npm run lint` worked locally...

kylelaker-ccpo:
That's a different error at least

kylelaker-ccpo:
Ahh there we go! I should have [read the CHANGELOG](https://github.com/standard/standard/blob/master/CHANGELOG.md). `eslint-plugin-node` needs to be replaced with `eslint-plugin-n`.

kylelaker-ccpo:
Since starting this PR, a production release of Standard v17 was created.

zachclark-ccpo (approved):

PR #798, opened 2022-04-22
Update AWS CDK to v2.21.0

kylelaker-ccpo:
This updates the CDK and adopts a new feature flag for safety against accidental secret exposure. See https://github.com/aws/aws-cdk/releases/tag/v2.21.0 for release notes.


cdk.json
```diff
@@ -28,6 +28,7 @@
       "aws",
       "aws-us-gov"
     ],
+    "@aws-cdk/core:checkSecretUsage": true,
```
juliusfitzhugh-ccpo, line 31: Nice, it never hurts to have those extra safety checks!

juliusfitzhugh-ccpo (approved):

cdk.json
kylelaker-ccpo: It'll help in AT-7076 to make sure we're passing IAM User Access keys around safely too! So practical and timely!

PR #825, opened 2022-04-28
Add notional IdP configuration

kylelaker-ccpo:
This adds an initial implementation of an internal identity provider
that we can leverage and have trusted by the various CSPs who may
implement the ATAT API specification.

In this configuration, various client applications can be registered
as IdP clients. The initial CSP implementation is performed with Amazon
Cognito; however, a generic interface is also created to allow other
potential implementations. Additionally, the initial supported
application type is based on Lambda functions but theoretically any
wrapper could be created implementing the necessary interface (including
a State Machine, EC2 instance, or any other resource).

The GovCloudCompatibility Aspect is updated to handle various
differences in Cognito between the standard and GovCloud partitions.

The initial client implementation just fetches a token using the client
credentials flow and then does nothing with it. The token type and
expiration are logged (but not the token itself).


idp/client.ts
```diff
@@ -0,0 +1,45 @@
...
+export async function handler(event: APIGatewayProxyEvent, context: Context): Promise<APIGatewayProxyResult> {
+  logger.addContext(context);
+  logger.info("Retriving client secret from Secrets Manager", { secretId: IDP_CLIENT_SECRET_PATH });
+  const clientSecret = (await secrets.getSecretValue({ SecretId: IDP_CLIENT_SECRET_PATH })).SecretString!;
+  logger.info("Retrieving a token from IdP using Client ID", { clientId: IDP_CLIENT_ID });
+  const form = new FormData();
+  form.set("grant_type", "client_credentials");
+  form.set("client_id", IDP_CLIENT_ID);
+  const response = await axios.post(`https://${IDP_DOMAIN}/oauth2/token`, form, {
+    headers: {
```
zachclark-ccpo, line 22: Should there be a time out for the `await`?

kylelaker-ccpo: There should be a timeout, though to nitpick a bit, it's not the `await` but for the HTTP request as a whole.

idp/client.ts
```diff
@@ -0,0 +1,49 @@
...
+const logger = new Logger({ serviceName: "TestIdp" });
+
+const secrets = new SecretsManager({ useFipsEndpoint: true });
+
+const IDP_CLIENT_ID = process.env.IDP_CLIENT_ID!;
+const IDP_CLIENT_SECRET_NAME = process.env.IDP_CLIENT_SECRET_NAME!;
+const IDP_DOMAIN = process.env.IDP_DOMAIN!;
+
+// We use the ScheduledEvent here but in actuality, we don't care about the type
+// of the event. This function is only ever going to invoked via the CLI or console
```
jasonburkert-ccpo, line 14:
```suggestion
// of the event. This function is only ever going to be invoked via the CLI or console
```

lib/constructs/identity-provider.ts
```diff
@@ -0,0 +1,250 @@
...
+    }
+  }
+
+  public addClient(app: IIdentityProviderClientApplication, scopes: string[]): IIdentityProviderClient {
+    const client = new CognitoIdentityProviderClient(this, `${app.name}Client`, {
+      userPool: this.userPool,
+      clientApplication: app,
+      scopes,
+    });
+    // We can pretty safely perform this because custom domains are not support in GovCloud
```
jasonburkert-ccpo, line 204:
```suggestion
    // We can pretty safely perform this because custom domains are not supported in GovCloud
```

idp/client.ts
jasonburkert-ccpo, line 14: Is the focus on `logger.info` or `logger.error` output directly due to expectation of invocation only from CLI or console?

lib/constructs/identity-provider.ts
```diff
@@ -0,0 +1,250 @@
...
+  clientApplication: IIdentityProviderClientApplication;
+  scopes: string[];
+}
+
+/**
+ * Properties for creating a client on the identity provider.
+ */
+export interface CognitoIdentityProviderClientProps extends IdentityProviderClientOptions {
+  /**
+   * The identity provider to create a client for.
```
jasonburkert-ccpo, line 92:
```suggestion
   * The identity provider for which to create a client.
```

lib/constructs/identity-provider.ts
```diff
@@ -0,0 +1,250 @@
...
+  userPool: cognito.IUserPool;
+}
+
+/**
+ * A custom resource with additional details about an AWS::Cognito::UserPoolClient.
+ *
+ * This AWS::Cognito::UserPoolClient resource type exposes depressingly few attributes
+ * about the Client itself (like, not even the ID). This custom resource performs
+ * a cognito-idp:DescribeUserPoolClient API call to fetch all the attributes. This
+ * may not be a totally-ideal use of a custom resource but it means that we don't have
```
jasonburkert-ccpo, line 228: I challenge this hyphenation!
```suggestion
 * may not be a totally ideal use of a custom resource but it means that we don't have
```

jasonburkert-ccpo (approved):
Looks good.  I have no real critique, just English nitpicking (mostly for fun).
Sometimes I just read every word hoping that at least some of it sinks in via osmosis.
Nice work.

idp/client.ts
kylelaker-ccpo: Yeah, pretty much. We could probably type `event` as `any` or `{}` but originally I had thought I'd setup a trigger to just call it every so often for fun.

But yeah the data never gets _sent_ anywhere since there's not really any way to get the return value (it's not returned to another service like API Gateway or Step Functions). The real expectation is that this is being invoked just to "prove out" the process and maybe only once ever in it's life.

It is possible to get the payload from a `lambda:Invoke` via the CLI iirc but honestly, that feels hardly worth it.

So yeah, I'm making people go dig in the logs mostly because it's easier to implement and more consistent with what we'll need to do someday anyway. Because soon, this function won't exist. We'll turn it into a library that other functions can leverage as needed (probably very soon actually, in AT-7337). And no Function should ever _return_ any information about its token (and we assume in our accounts if you can directly Invoke a Lambda you probably can also read its logs).

kylelaker-ccpo:
Thanks! I'll patch up the wording and add a link to other relevant materials tomorrow.

lib/atat-web-api-stack.ts
```diff
@@ -85,5 +88,26 @@ export class AtatWebApiStack extends cdk.Stack {
...
+            { name: "read-portfolio", description: "Allow reading information about portfolios" },
+            { name: "write-portfolio", description: "Allow creating and updating portfolios" },
+          ],
+        },
+      ],
+    });
+    const demoApp = atatIdp.addClient(
+      new idp.IdentityProviderLambdaClient(
+        "DemoClient",
+        new nodejs.NodejsFunction(this, "SampleFn", { entry: "idp/client.ts" })
```
juliusfitzhugh-ccpo, line 108: Is there a way to invoke this demo client without admin that i'm missing?

idp/client.ts
```diff
@@ -0,0 +1,50 @@
...
+  const clientSecret = (await secrets.getSecretValue({ SecretId: IDP_CLIENT_SECRET_NAME })).SecretString!;
+  logger.info("Retrieving a token from IdP using Client ID", { clientId: IDP_CLIENT_ID });
+  const form = new URLSearchParams();
+  form.set("grant_type", "client_credentials");
+  form.set("client_id", IDP_CLIENT_ID);
+  let response;
+  try {
+    response = await axios.post(`https://${IDP_DOMAIN}/oauth2/token`, form, {
+      // Allow the request to take its time but give some time for cleanup
+      timeout: context.getRemainingTimeInMillis() - 500,
```
zachclark-ccpo, line 29: Good `timeout` here.

lib/constructs/identity-provider.ts
```diff
@@ -0,0 +1,247 @@
...
+  public readonly domainName: string;
+
+  public readonly userPool: cognito.IUserPool;
+  public readonly domain: cognito.IUserPoolDomain;
+
+  private readonly resourceServers: cognito.UserPoolResourceServer[] = [];
+  constructor(scope: Construct, id: string, props?: CognitoIdentityProviderProps) {
+    super(scope, id);
+    this.userPool = new cognito.UserPool(this, "IdP", {});
+    this.domain = this.userPool.addDomain("Domain", {
```
zachclark-ccpo, line 177: `domain` named Domain?
Should we name this atat-domain or something else?

lib/atat-web-api-stack.ts
```diff
@@ -85,5 +88,26 @@ export class AtatWebApiStack extends cdk.Stack {
...
+      domainPrefix: `atat${environmentName.toLowerCase()}`,
+      scopes: [
+        {
+          resourceServerName: "atat",
+          scopes: [
+            { name: "read-cost", description: "Allow reading cost information" },
+            { name: "read-portfolio", description: "Allow reading information about portfolios" },
+            { name: "write-portfolio", description: "Allow creating and updating portfolios" },
+          ],
+        },
```
zachclark-ccpo, line 102: Should these `scopes` live in another file?

.github/workflows/test.yml
```diff
@@ -103,5 +103,4 @@ jobs:
       - name: Scan for new (not-yet-allowed) secrets
         run: >-
           git ls-files -z |
-          xargs -0 detect-secrets-hook --baseline .secrets.baseline --exclude-files
-          'package-lock.json'
+          xargs -0 detect-secrets-hook --baseline .secrets.baseline
```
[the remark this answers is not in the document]
kylelaker-ccpo: We aren't -- we're just deferring to the `exclude` list in the baseline file (or at least, that's what this should be changing)

lib/atat-web-api-stack.ts
kylelaker-ccpo: There is not. And that's intentional

lib/constructs/identity-provider.ts
kylelaker-ccpo: Nope! In the construct tree this will end up having the names concatenated. We probably actually need to get _out_ of the habit of jamming `Atat` in front of everything and just defining it in the top level. You can see the emitted in Template to see that it still has a unique logical ID

lib/atat-web-api-stack.ts
kylelaker-ccpo: Maybe? We probably want to narrow down exactly what to call our Resource Server. These likely _should_ be context values actually.

lib/atat-web-api-stack.ts
juliusfitzhugh-ccpo: Fair, I was trying to make sure it did as we expected, but believe you mentioned you got the expected log.

juliusfitzhugh-ccpo (approved):
Looks good from everything I've seen. There are a few code smells by sonar, but nothing major.

lib/constructs/identity-provider.ts
```diff
@@ -0,0 +1,247 @@
...
+      cognitoDomain: {
+        domainPrefix: props?.domainPrefix ?? id.toLowerCase(),
+      },
+    });
+    this.domainName = this.domain.domainName;
+    if (props?.scopes) {
+      for (const scopeConfig of props.scopes) {
+        if (!scopeConfig.resourceServerName) throw new Error("Cognito requires a resource server");
+        this.resourceServers.push(
+          this.userPool.addResourceServer(scopeConfig.resourceServerName, {
```
jeffsegal-ccpo, line 187: I'm making sure I understand how Resource Servers work, but does this give every user in the User Pool all the scopes?

jeffsegal-ccpo: Digging into the docs a bit more, it looks like this is what we need: https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_cognito.OAuthSettings.html

kylelaker-ccpo: I'm not entirely sure that I understand? A Resource Server has a name and scopes configured. Then clients can have scopes they're allowed to use; however, those scopes must have first been configured on/with a resource server associated with the pool.

I do think it's important to emphasize that in our flow, we shouldn't have users at any point, right? It's all going to be `client_credentials` flows? So the scope(s) of a user should be irrelevant.

The client can request (and be configured for) zero or more scopes on zero or more resource servers.

That object (`OAuthSettings`) is what we're passing to the second parameter of `addResourceServer`. We get a little lazy in how we define things for clients (and use use a string formatted as `<resourceServerId>/<scopeName>`) rather than requiring them to pass the whole object in.

If we don't configure a Client to have a scope here, then Cognito will not allow us to request it when we get a token. The only way to configure a scope for a client is if a Resource Server with that scope is associated with the User Pool. The Resource Server name is explicitly required to be part of the scope name as well.

So we do a little bit of `map` magic to provide a slightly nicer and more generic interface to consumers of this module, but it's all designed with these constraints in mind.

jeffsegal-ccpo: Yes, you're right about users. I'm ultimately trying to make sure that we can configure a client with the appropriate scope, but it looks like we just need to add a `scope` param to our `/token` request to ensure that the token being used by a particular Lambda gets the least privileged scope (per standard OAuth).  Sorry for sowing some confusion here.

jeffsegal-ccpo: Great, thanks!

kylelaker-ccpo: Uhh oops:

> If the client doesn't request any scopes, the authentication server uses all custom scopes associated with the client.

https://docs.aws.amazon.com/cognito/latest/developerguide/token-endpoint.html#post-token-request-parameters-in-body

idp/client.ts
```diff
@@ -0,0 +1,135 @@
...
+ * @param clientId The OAuth Client ID
+ * @param scopes A list of scopes to request
+ */
+function buildTokenRequest(clientId: string, scopes?: string[]): URLSearchParams {
+  const form = new URLSearchParams();
+  form.set("grant_type", "client_credentials");
+  form.set("client_id", clientId);
+  if (scopes?.length) {
+    form.set("scope", scopes.join(" "));
+  }
```
zachclark-ccpo, line 38: I am curious why `scopes` are optional for the `buildTokenRequest` function?

kylelaker-ccpo: Great question!! So https://docs.aws.amazon.com/cognito/latest/developerguide/token-endpoint.html#post-token-request-parameters-in-body will be a good reference here. By default (at least with Cognito) the server will return all scopes the client is configured for. This allows us to limit the scopes if we want. 

From the Cognito docs:

> Can be a combination of any custom scopes associated with an app client. Any scope that you request must be activated for the app client, or Amazon Cognito will ignore it. If the client doesn't request any scopes, the authentication server uses all custom scopes associated with the client.

Even when using `client_credentials`, this is still an optional argument with a decent default behavior.

lib/constructs/identity-provider.test.ts
```diff
@@ -0,0 +1,144 @@
...
+  });
+});
+
+class NullClientApplication implements idp.IIdentityProviderClientApplication {
+  public name = "NULLCLIENT";
+  public isConfigured = false;
+  configure(): void {
+    this.isConfigured = true;
+  }
+}
```
zachclark-ccpo, line 97: Should this be moved to the top of the test file/ comment explanation?

zachclark-ccpo (approved):

lib/constructs/identity-provider.test.ts
kylelaker-ccpo: Probably not? It is very tightly integrated to the surrounding code. But I can do that if you'd like

lib/constructs/identity-provider.ts
kylelaker-ccpo: My general idea is that no client will ever be configured to be allowed more than one scope anyway? So in practice, it _shouldn't matter_. For completeness, yes, a client should still request the specific scope they need for a particular action.

So the User Pool will have all the scopes defined. Each client will have the maximal scope(s) it needs, keeping that maximum set as minimal as possible. Each client will also, at runtime, only requests the scopes it needs.

jeffsegal-ccpo (approved):
Great stuff! Really excited for this to become a reality so quickly.

PR #831, opened 2022-05-02
Make using the console easier

kylelaker-ccpo:
This gives developers the ability to read and poke in the Lambda and API
Gateway consoles, permissions that may have been missing before. This
doesn't grant everything (and definitely not permission to invoke
anything) but hopefully it makes life a little easier.


juliusfitzhugh-ccpo (approved):
Thank you, this will surely help!

PR #832, opened 2022-05-02
Lint on commit

kylelaker-ccpo:
This automatically runs a `npm run lint:fix` when TS, JS, or JSON files
change. This will increase the amount of time it takes for the
pre-commit checks to run but at least your code will automatically be
linted!

This may not fit well into folks workflows and if it doesn't then let's
not merge it.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,9 +1,19 @@
 # See https://pre-commit.com for more information
 # See https://pre-commit.com/hooks.html for more hooks
 repos:
+  - repo: local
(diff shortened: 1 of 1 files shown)


juliusfitzhugh-ccpo (approved):
Works for me.

zachclark-ccpo (approved):

zachclark-ccpo:
Works locally for me as well.

jeffsegal-ccpo (approved):
Looks good and should cut down on possible linting churn.

PR #833, opened 2022-05-03
Update dependencies

kylelaker-ccpo:


kylelaker-ccpo:
Looks like `ts-jest@28` is out so I just went ahead and updates `jest@28` as well. Had been holding the latter at 27 until the former updated to resolve the dependency conflict.

juliusfitzhugh-ccpo (approved):

PR #838, opened 2022-05-04
Define generateDocument Request for DOW

zachclark-ccpo:
## Overview
Added the `/generate-document` POST request path with supporting schema objects to the newly renamed `hoth_api.yaml` (previously known as `atat_provisioning_api.yaml`).

Ticket: [AT-7343](https://ccpo.atlassian.net/browse/AT-7343)


hoth_api.yaml
```diff
@@ -1,62 +1,90 @@
 openapi: 3.0.2
 info:
   description: >-
-    This is the ATAT Provisioning API, an internal API solely meant for use by the ServiceNow component of ATAT in
-    order to kick off provisioning jobs.
-  version: '0.1'
-  title: ATAT Internal API - Provisioning
+    This is the HOTH (Hyperscaler Orchestration and Tracking Helper) API, a private internal-facing API that is intended
```
jeffsegal-ccpo, line 4:
```suggestion
    This is the HOTH (Hyperscale Orchestration and Tracking Helper) API, a private internal-facing API that is intended
```

hoth_api.yaml
```diff
@@ -265,4 +293,14 @@ components:
...
\ No newline at end of file
+          additionalProperties: true
+    GenerateDocumentRequest:
+      oneOf:
+        - $ref: "#/components/schemas/DescriptionOfWork"
+        - $ref: "#/components/schemas/SomeFutureDocument"
+    DescriptionOfWork:
+      description: "TODO: AT-7341, define DescriptionOfWork schema"
+      type: object
+    SomeFutureDocument:
```
jeffsegal-ccpo, line 304: We can just leave this off for now; I included it in the ticket just to make my intentions clear.

hoth_api.yaml
```diff
@@ -265,4 +293,14 @@ components:
...
           type: object
-          additionalProperties: true
\ No newline at end of file
+          additionalProperties: true
+    GenerateDocumentRequest:
+      oneOf:
+        - $ref: "#/components/schemas/DescriptionOfWork"
+        - $ref: "#/components/schemas/SomeFutureDocument"
+    DescriptionOfWork:
+      description: "TODO: AT-7341, define DescriptionOfWork schema"
```
jeffsegal-ccpo, line 302: Thanks for including the breadcrumb here

jeffsegal-ccpo (changes requested):

hoth_api.yaml
```diff
@@ -1,62 +1,90 @@
...
+            application/pdf:
+              schema:
+                type: string
+                format: binary
+        "400":
+          description: Invalid input
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ValidationError"
```
kylelaker-ccpo, line 37: We probably need a separate error type here too. Something like:

```json
{
  "errorCode": "ResponseTooLarge",
  "details": {
    "responseSize": 10202323,
    "maxAllowed": 6291456,
  }
}
```

We _know_ we have a 6MB response limit. And we don't currently know what our solution will be. Encoding a response for this would give us two benefits:
1. It's a code path we have to consider out of the gate in the first place
2. The client already has to anticipate a separate code path for this scenario (so that some day we can send a 201/303 response)

Though, this may be better as a `5xx` family error since there's nothing the client really did wrong (besides sending us too much text).

Should the DoW ever be >6MB? Probably not. But we need to anticipate hard limits of they're not part of our API.

kylelaker-ccpo (commented):
I would request changes but I feel like @jeffsegal-ccpo should :+1: or :-1: this suggestion before we move forward.

hoth_api.yaml
kylelaker-ccpo: Not according to the README :) We're orchestrating and tracking the hyperscalers. We're not hyperscale

kylelaker-ccpo: We can, of course, change the README. But regardless, gotta be consistent.

hoth_api.yaml
zachclark-ccpo: Should we return **413 Payload Too Large** as the error code?
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/413

hoth_api.yaml
jeffsegal-ccpo: Haha fair enough, I just assumed it was a typo. You win :)

hoth_api.yaml
jeffsegal-ccpo: Good call, and we should also remember to keep this in mind as we consider any potential solutions in the future. That is, whether one library or another, or one style or another (procedural vs template-driven) yields significantly smaller files.

kylelaker-ccpo: Looking into this more, Zach and I found that we probably want `422 Unprocessable Entity` if the request is otherwise correct

kylelaker-ccpo (approved):

jeffsegal-ccpo:
> I would request changes but I feel like @jeffsegal-ccpo should 👍 or 👎 this suggestion before we move forward.

Looks good to me, thanks guys.

PR #846, opened 2022-05-06
Cache IdP Client Secret

kylelaker-ccpo:
We should eagerly fetch the client secret and then cache it. Having to
look it up for every request is slow. The string will be stored
in-memory within the execution environment but not permanently as part
of the Environment configuration.

This follows the [recommendations to cache secrets][1].

[1]: https://docs.aws.amazon.com/secretsmanager/latest/userguide/retrieving-secrets.html


zachclark-ccpo (approved):

kylelaker-ccpo:
Quick non-scientific look at "Execution time" metrics suggests that this can reduce subsequent runs by 100-300ms.

PR #849, opened 2022-05-10
Changes for AWS CDK v2.24.0

kylelaker-ccpo:
The next version of the CDK will bring some nice features. First, it
will have a fix for the `UserPoolDomain.baseUrl` method, which allows
building URLs that work in GovCloud. Additionally, it adds support for
the NodeJS 16 runtime.

This PR _must not_ be merged now because it is not compatible with the
existing version of the CDK; however, I wanted to stage the changes so
that we mostly have them ready (almost as a to-do list) for when the
version is released (hopefully soon).


lib/constructs/identity-provider.ts
```diff
@@ -236,12 +236,8 @@ export class CognitoIdentityProvider extends Construct implements IIdentityProvi
...
       scopes,
     });
-    // The UserPoolDomain object provides a `baseUrl()` function; however, the URL it creates
-    // does not work in us-gov-west-1. Instead, we build the URL manually. We can assume that
-    // `domainName` is not a full domain as they are not supported in `us-gov-west-1`.
-    // See: aws/aws-cdk#20182
-    const fullDomain = `${this.domain.domainName}.auth-fips.${cdk.Aws.REGION}.amazoncognito.com`;
-    app.configure(client.clientId, client.secret, fullDomain);
+    // Amazon Cognito requires the usage of the FIPS endpoints in us-gov-west-1
+    app.configure(client.clientId, client.secret, this.domain.baseUrl({ fips: true }));
```
juliusfitzhugh-ccpo, line 240: This is much cleaner, nice work on the fix.

jeffsegal-ccpo (approved):

jasonburkert-ccpo (approved):
Very nice.

PR #856, opened 2022-05-11
Refactor IdP Client into a library

kylelaker-ccpo:
This sorta turns our IdP client into more of a private/internal library
that meets our needs. It doesn't rely on anything too fancy. It lets us
pull configuration for the IdP from the environment and turn it into
something useful. Ideally, this means a client could almost just import
`getToken` and call it with no arguments and get something workable in
the majority of use cases. It's not perfect and won't extend too far
beyong client credentials flow but honestly we shouldn't really need
that.

Ticket: AT-7337
Ticket: AT-7338


lib/constructs/identity-provider.ts
```diff
@@ -107,7 +107,7 @@ export class CognitoIdentityProviderClient extends Construct {
 
   constructor(scope: Construct, id: string, props: CognitoIdentityProviderClientProps) {
     super(scope, id);
-    const client = props.userPool.addClient("Client", {
+    const client = props.userPool.addClient(id, {
```
kylelaker-ccpo, line 110: This change is to make it so an IdP can actually have more than one client. I think @zachclark-ccpo mentioned before that something like this may be necessary and I said that was incorrect. _I_ was actually the one who was incorrect.

api/provision/mock-invocation-fn.ts
```diff
@@ -61,7 +75,8 @@ export function createCspResponse(request: ProvisionRequest): CspResponse {
       };
       console.log("Failed response : " + JSON.stringify(response));
```
juliusfitzhugh-ccpo: I noticed you were replacing `console.log` with `logger.info` was there a reason some of the `console.log` couldn't also be changed?

utils/logging.ts
```diff
@@ -0,0 +1,37 @@
...
+    };
+  }
+}
+
+const baseLogger = new Logger({
+  serviceName: process.env.POWERTOOLS_SERVICE_NAME ?? "ATAT",
+  logFormatter: new AtatLogFormatter(),
+});
+
+export const logger = baseLogger;
```
juliusfitzhugh-ccpo, line 37: You've taken the logger to the next level, nice!

lib/constructs/provisioning-sfn-workflow.ts
```diff
@@ -55,9 +60,20 @@ export class ProvisioningWorkflow extends Construct implements IProvisioningWork
...
     // Provisioning State machine functions
+    const cspConfig = secrets.Secret.fromSecretNameV2(
+      this,
+      "CspConfiguration",
+      this.node.tryGetContext("atat:CspConfigurationPath")
+    );
     const mockInvocationFn = new lambdaNodeJs.NodejsFunction(scope, "MockInvocationFunction", {
       entry: "api/provision/mock-invocation-fn.ts",
+      environment: { CSP_DATA_SECRET: cspConfig.secretArn },
+      timeout: cdk.Duration.seconds(5),
```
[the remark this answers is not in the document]
kylelaker-ccpo: We can go all the way to Lambda's max of 15min!

cdk.json
```diff
@@ -29,6 +29,7 @@
       "aws-us-gov"
     ],
     "@aws-cdk/core:checkSecretUsage": true,
+    "atat:CspConfigurationPath": "config/csp/integration",
```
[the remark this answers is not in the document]
kylelaker-ccpo: Correct. We can rename this parameter perhaps.

idp/client.ts
```diff
@@ -40,49 +85,53 @@ function buildTokenRequest(clientId: string, scopes?: string[]): URLSearchParams
...
-  clientId: string,
-  clientSecret: string,
-  baseUrl: string,
-  timeout?: number,
-  scopes?: string[]
-): Promise<TokenResponse> {
+export async function getToken(input?: GetTokenInput): Promise<TokenResponse> {
+  const { clientId, clientSecret, idpBaseUrl } = await (
+    input?.clientConfigurationProvider ?? getConfigurationFromEnvironment
+  )();
```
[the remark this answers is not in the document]
kylelaker-ccpo: Basically, I want to live in a world where even if a function were to get "all" its possible scopes, that list is still the exact same as the list it'd have asked for specifically.

utils/logging.ts
kylelaker-ccpo: I am really happy with this formatting!! I think eventually we'd be able to pass through the caller identity to the logs too. I've shared a screenshot of what this looks like in a Slack thread. And I think the only real remaining thing to have logging go the extra mile would be `middy` integration and some pre-built CloudWatch queries.

idp/client.ts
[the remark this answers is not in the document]
jeffsegal-ccpo: OK that makes perfect sense. I just didn't have a good feel for the boundary between what will remain and what will change once we add the cost polling Lambda.

api/provision/mock-invocation-fn.ts
```diff
@@ -40,7 +51,66 @@ export async function baseHandler(stateInput: ProvisionRequest): Promise<CspResp
...
+      content: {
+        details: "Invalid CSP provided",
+      },
+    };
+  }
+
+  const response = await axios.post(url, request.payload, {
+    headers: {
+      Authorization: `Bearer ${(await getToken()).access_token}`,
+      "User-Agent": "ATAT v0.2.0 client",
```
jeffsegal-ccpo, line 89: We'll eventually want to parameterize the version so we don't lose track and end up confusing ATAT servers but I'm good with a TODO for now

kylelaker-ccpo: Yeah somewhere on my TODO list for awhile has been "Dynamically find version in node project"

kylelaker-ccpo: Ehh but I guess that doesn't help here -- we want the version of the _spec_... hmm

kylelaker-ccpo: Actually... what we probably want to do is write an ATAT client library more generally... And then dynamically get the version of that. Something to think about. Maybe tomorrow when we're talking about tickets we need to write.

lib/constructs/identity-provider.ts
```diff
@@ -240,6 +240,14 @@ export class CognitoIdentityProvider extends Construct implements IIdentityProvi
     app.configure(client.clientId, client.secret, this.domain.baseUrl({ fips: true }));
     return client;
   }
+
+  public discoveryUrl(): string {
+    // Breaking up the variables is the most readble to make this work without the string being too long
+    const region = cdk.Aws.REGION;
+    const suffix = cdk.Aws.URL_SUFFIX;
+    const userPoolId = this.userPool.userPoolId;
+    return `https://cognito-idp-fips.${region}.${suffix}/${userPoolId}/.well-known/openid-configuration`;
```
zachclark-ccpo, line 249: Why do you need `${suffix}` isn't it always going to be `amazonaws.com`?

kylelaker-ccpo: Not entirely sure what it _would_ be in other partitions https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/pseudo-parameter-reference.html#cfn-pseudo-param-urlsuffix

kylelaker-ccpo: Cognito doesn't currently exist in any other partitions that do/would have a different URL suffix but it may some day and a deployment of our application should work in such a partition.

api/provision/csp-write-portfolio.ts
```diff
@@ -0,0 +1,142 @@
...
+  // we no longer need the mock implementations of CSP_<A|B|C|D> and can fully use the
+  // mock (or real) integration endpoints.
+  const cspResponse = await createCspResponse(stateInput);
+
+  // Throws a custom error identified by the state machine and the function retries 2
+  // times before failing continuing through the remaining states
+  if (cspResponse.code >= 500) {
+    const error = new Error(JSON.stringify(cspResponse));
+    Object.defineProperty(error, "name", {
+      value: "MockCspApiError",
```
jeffsegal-ccpo, line 40:
```suggestion
      value: "CspApiError",
```

api/provision/csp-write-portfolio.ts
```diff
@@ -0,0 +1,142 @@
...
+      response = {
+        code: 500,
+        content: {
+          some: "internal error",
+        },
+      };
+      logger.error("Internal error response", { response: response as any });
+      return response;
+    default:
+      response = await makeARealRequest(request);
```
jeffsegal-ccpo, line 133: Just making sure I am following - if we want to be able to test against our actual mock server, we can add it to our SM config and pass in `foo` as the CSP name, ensuring that we arrive at this clause, right?

jeffsegal-ccpo (changes requested):
Love how the tests turned out. Just one tiny suggestion that not necessarily be valid (depending on whether we consider ourselves in "mock" or "real" invocation land).

api/provision/csp-write-portfolio.ts
kylelaker-ccpo: Exactly! The CSP name that I've configured so far is `CSP_Mock` but we can choose basically whatever we want. Once we iron out some things with the existing Mock server implementation (around our Auth issues), then I'll update this to just always `makeARealRequest`.

So for today:
 - `"CSP_A" | "CSP_B" | "CSP_C" | "CSP_D"` -> The same mock 200/400/500 responses as before
 - Any other value -> Lookup in config
   - If present (for now, just `"CSP_Mock"`), use the configured URI and send a request
   - Else, return a 400

api/provision/csp-write-portfolio.ts
kylelaker-ccpo: Renamed! Good catch!

jeffsegal-ccpo (approved):

PR #867, opened 2022-05-16
Configure logs for indefinite retention

kylelaker-ccpo:
Right now most of our logs are AWS Lambda function logs, which have
their logs groups created automatically. Those log groups have
indefinite retention specified. Our other log groups use different
configurations and we should just make everything consistent until we
really solidify how long we plan on retaining for.


juliusfitzhugh-ccpo (approved):
Yeah, I agree we should stay consistent. I'm guessing how long depends on the auditing policy for the project?

kylelaker-ccpo:
> Yeah, I agree we should stay consistent. I'm guessing how long depends on the auditing policy for the project?

Yep 100%. We're definitely talking on the time scale of years (so weeks is definitely too small), and it's likely either 7 or 10.

PR #877, opened 2022-05-23
Define DescriptionOfWork schema (and supporting schema) in HOTH API

zachclark-ccpo:
## Overview
This is the schema used for the `generateDocument` POST request.

The `DescriptionOfWork` object includes all properties needed to complete the DOW document, and separate schemas had to be
created to improve readability.

These schemas were created referencing the specific SNOW Tables needed for the DOW (See https://github.com/dod-ccpo/atat-snow/pull/48). 

[Acquisition Package Field Mapping](https://ccpo.atlassian.net/wiki/spaces/AT/pages/2282160166/Acquisition+Package+Field+Mapping) Confluence Page was used along with the actual **Draft JWCC Description of Work Template** (See slack comment for link).

In total this PR adds the following schema:

1. `DescriptionOfWork`
2. `CurrentEnvironment`
3. `EnvironmentInstance`
4. `ClassificationLevel`
5. `StorageUnit`
6. `PerformanceRequirements`
7. `RequiredService`
8. `Period`
9. `ServiceOffering`
10. `PeriodOfPerformance`
11. `GFEOverview`
12. `ContractConsiderations`
13. `TrainingCourses`
14. `SensitiveInformation`

[Ticket: AT-7341](https://ccpo.atlassian.net/browse/AT-7341)

### DIFF
--- a/hoth_api.yaml
+++ b/hoth_api.yaml
@@ -330,7 +330,7 @@ components:
           maxLength: 300
         scope_surge:
           type: integer
-          maxLength: 50
+          maximum: 50
         current_environment: # DOW section 3
           $ref: "#/components/schemas/CurrentEnvironment"
         performance_requirements: # DOW section 4
@@ -381,22 +381,22 @@ components:
           maxLength: 1000
         number_of_vCPUs:
           type: integer
-          maxLength: 40
+          maximum: 40
         storage_type:
           type: string # Empty choice on SNOW side, set as empty String until we get details
         storage_amount:
           type: integer
-          maxLength: 40
+          maximum: 40
         storage_unit:
           $ref: "#/components/schemas/StorageUnit"
         memory_amount:
           type: integer
-          maxLength: 40
+          maximum: 40
         memory_unit:
           $ref: "#/components/schemas/StorageUnit"
         data_egress_monthly_amount:
           type: integer
-          maxLength: 40
+          maximum: 40
         data_egress_monthly_unit:
           $ref: "#/components/schemas/StorageUnit"
     ClassificationLevel:
--- a/hoth_api.yaml
+++ b/hoth_api.yaml
@@ -342,8 +342,10 @@ components:
         current_environment:
           $ref: "#/components/schemas/CurrentEnvironment"
         # DOW section 4 (and its subtasks)
-        performance_requirements:
-          $ref: "#/components/schemas/PerformanceRequirements"
+        required_services:
+          type: array
+          items:
+            $ref: "#/components/schemas/RequiredService"
         # DOW section 5 TBD
         # DOW section 6 TBD
         # DOW section 7
@@ -474,17 +476,6 @@ components:
       enum:
         - GB
         - TB
-    PerformanceRequirements:
-      description: >
-        Part of DOW, encapsulates all data for 4. Performance Requirements and its subtasks. 
-        On the API side, the subtasks will filter through the provided required_services.service_offerings
-        to create the subtasks fields.
-      type: object
-      properties:
-        required_services:
-          type: array
-          items:
-            $ref: "#/components/schemas/RequiredService"
     RequiredService:
       description: "Part of DOW, represents a single required service"
       type: object
--- a/hoth_api.yaml
+++ b/hoth_api.yaml
@@ -504,11 +504,7 @@ components:
           items:
             $ref: "#/components/schemas/Period"
         pop_start_request:
-          type: string
-          enum:
-            - Unselected
-            - Yes
-            - No
+          type: boolean
         requested_pop_start_date:
           type: string
           format: date
@@ -519,11 +515,7 @@ components:
             - No later than
             - No sooner than
         recurring_requirement:
-          type: string
-          enum:
-            - Unselected
-            - Yes
-            - No
+          type: boolean
     GFEOverview:
       description: "Part of DOW, represents GFE Overview fields"
       properties:
@@ -536,17 +528,9 @@ components:
           type: string
           maxLength: 6
         property_accountable:
-          type: string
-          enum:
-            - Unselected
-            - Yes
-            - No
+          type: boolean
         gfe_or_gfp_furnished:
-          type: string
-          enum:
-            - Unselected
-            - Yes
-            - No
+          type: boolean
     ContractConsiderations:
       description: "Part of DOW, represents Contract Considerations Table"
       properties:
@@ -557,21 +541,13 @@ components:
         packaging_shipping_other_explanation:
           type: string
         potential_conflict_of_interest:
-          type: string
-          enum:
-            - Unselected
-            - Yes
-            - No
+          type: boolean
         conflict_of_interest_explanation:
           type: string
         contractor_provided_transfer:
           type: boolean
         contractor_required_training:
-          type: string
-          enum:
-            - Unselected
-            - Yes
-            - No
+          type: boolean
         required_training_services: # This field looks incomplete on the SNOW side
           $ref: "#/components/schemas/TrainingCourses"
     TrainingCourses:
@@ -586,11 +562,7 @@ components:
       description: "Part of DOW, represents Sensitive Information Table fields"
       properties:
         pii_present:
-          type: string
-          enum:
-            - Unselected
-            - Yes
-            - No
+          type: boolean
         work_to_be_performed:
           type: string
         system_of_record_name:
@@ -619,22 +591,10 @@ components:
         FOIA_zip_postal_code:
           type: string
         BAA_required:
-          type: string
-          enum:
-            - Unselected
-            - Yes
-            - No
+          type: boolean
         potential_to_be_harmful:
-          type: string
-          enum:
-            - Unselected
-            - Yes
-            - No
+          type: boolean
         section_508_sufficient:
-          type: string
-          enum:
-            - Unselected
-            - Yes
-            - No
+          type: boolean
         accessibility_reqs_508:
           type: string
--- a/hoth_api.yaml
+++ b/hoth_api.yaml
@@ -356,6 +356,7 @@ components:
           $ref: "#/components/schemas/SensitiveInformation"
     CurrentEnvironment:
       description: "Part of DOW, encapsulates all data for 3. Background/Current Environment section."
+      type: object
       properties:
         current_environment_exists:
           type: boolean
@@ -367,6 +368,7 @@ components:
           type: string
     EnvironmentInstance:
       description: "Part of DOW, represents a single Environment Instance"
+      type: object
       properties:
         instance_name:
           type: string
@@ -415,6 +417,7 @@ components:
           $ref: "#/components/schemas/StorageUnit"
     ClassificationLevel:
       description: "Part of DOW, Classification Level table fields"
+      type: object
       properties:
         classification:
           type: string
@@ -440,6 +443,7 @@ components:
         Part of DOW, encapsulates all data for 4. Performance Requirements and its subtasks. 
         On the API side, the subtasks will filter through the provided required_services.service_offerings
         to create the subtasks fields.
+      type: object
       properties:
         required_services:
           type: array
@@ -447,6 +451,7 @@ components:
             $ref: "#/components/schemas/RequiredService"
     RequiredService:
       description: "Part of DOW, represents a single required service"
+      type: object
       properties:
         applicable_classification_levels:
           type: array
@@ -468,6 +473,7 @@ components:
           type: string
     Period:
       description: "Part of DOW, represents Period Table fields"
+      type: object
       properties:
         period_type:
           type: string
@@ -487,6 +493,7 @@ components:
           type: integer
     ServiceOffering:
       description: "Part of DOW, represents a single Service Offering (Service Offering table)"
+      type: object
       properties:
         service_offering_group:
           type: string
@@ -507,6 +514,7 @@ components:
           type: string
     PeriodOfPerformance:
       description: "Part of DOW, represents a single Period of Performance"
+      type: object
       properties:
         base_period:
           $ref: "#/components/schemas/Period"
@@ -529,6 +537,7 @@ components:
           type: boolean
     GFEOverview:
       description: "Part of DOW, represents GFE Overview fields"
+      type: object
       properties:
         dpas_unit_id:
           type: string
@@ -544,6 +553,7 @@ components:
           type: boolean
     ContractConsiderations:
       description: "Part of DOW, represents Contract Considerations Table"
+      type: object
       properties:
         packaging_shipping_none_apply:
           type: boolean
@@ -571,6 +581,7 @@ components:
           type: string
     SensitiveInformation:
       description: "Part of DOW, represents Sensitive Information Table fields"
+      type: object
       properties:
         pii_present:
           type: boolean
--- a/hoth_api.yaml
+++ b/hoth_api.yaml
@@ -331,6 +331,10 @@ components:
         scope_surge:
           type: integer
           maxLength: 50
+        current_environment: # DOW section 3
+          $ref: "#/components/schemas/CurrentEnvironment"
+        performance_requirements: # DOW section 4
+          $ref: "#/components/schemas/PerformanceRequirements"
     CurrentEnvironment:
       description: "Part of DOW, encapsulates all data for 3. Background/Current Environment section."
       properties:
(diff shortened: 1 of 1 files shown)


zachclark-ccpo:
Several notes for reviewers:

- Recently we changed a few tables to use a **Choice** data type (with `Yes, No, Unselected`). These fields used to use **Boolean**. In this API spec I included any of these **Choice** fields with an `enum` for the options, however I believe that all of these should just be **Boolean** and nullable instead of `Yes, No, Unselected)`. I didn't want to make the decision without feedback so please give me permission to make these fields `true/false` again.  Here is the PR reference for the `Yes, No, Unselected` change: https://github.com/dod-ccpo/atat-snow/pull/55
- Many columns on the SNOW side still sport the default `maxLength=40`, this is problematic for names longer than 40, I have included them in the schema and called out several with inline spec comments. We should create a PR to adjust these maxLengths, please respond to my inline comments on the spec when reviewing.
- You may notice there is no `AcquisitionPackage` schema. I was torn whether or not to create an `AcquisitionPackage` to hold everything however in practice it just served as a wrapper for everything contained in the `DescriptionOfWork` object, and following the guidance of the ticket I decided to not include it.

hoth_api.yaml
```diff
@@ -319,5 +319,351 @@ components:
...
+        required_training_services: # This field looks incomplete on the SNOW side
+          $ref: "#/components/schemas/TrainingCourses"
+    TrainingCourses:
+      description: "Part of DOW, part of schema for ContractConsiderations, includes training courses name and value"
+      type: object
+      properties:
+        name:
+          type: string
+        value:
+          type: string
```
juliusfitzhugh-ccpo, line 603: I think this object will have a slightly different structure which was decided by the UI for keeping order of the trainings required. Something like the below from my understanding. @jasonburkert is this how you understood it as well?
```javascript
const training_courses = {
   "1": "Intro training",
   "2": "Basic training",
   "3": "Intermediate training"
}
```

Might be better for us to deal with this as an `array` (converting before the request) with the `name` being the `index`, since there could be any number of trainings provided or none at all. Or leave it as a `string` and handle it how we see fit in AWS. Not sure which is the right approach, just suggestion alternatives.

juliusfitzhugh-ccpo (changes requested):
Nice work and some good questions! I addressed your questions best I could in the `hoth_api.yaml` and added some additional comments.

hoth_api.yaml
```diff
@@ -319,5 +319,351 @@ components:
...
+        # DOW section 5 TBD
+        # DOW section 6 TBD
+        period_of_performance: # DOW section 7
+          $ref: "#/components/schemas/PeriodOfPerformance"
+        # DOW section 8 TBD
+        gfe_overview: # DOW section 9
+          $ref: "#/components/schemas/GFEOverview"
+        contract_considerations: # DOW section 10 (and its subtasks)
+          $ref: "#/components/schemas/ContractConsiderations"
+        section_508_accessibility_standards_for_cloud_computing: # DOW section 11
```
juliusfitzhugh-ccpo, line 347: This is a pretty long name, should we try to shorten it? I would probably suggest `section_508_accessibility_standards` unless there is another name that conflicts. It doesn't need to be changed though, I just noticed it was long.

juliusfitzhugh-ccpo, line 347: This is a pretty long name, should we try to shorten it? I would probably suggest `section_508_accessibility_standards` unless there is another name that conflicts. It doesn't need to be changed though, I just noticed it was long.

hoth_api.yaml
```diff
@@ -319,5 +319,351 @@ components:
...
+      description: >
+        Part of DOW, encapsulates all data for 4. Performance Requirements and its subtasks. 
+        On the API side, the subtasks will filter through the provided required_services.service_offerings
+        to create the subtasks fields.
+      properties:
+        required_services:
+          type: array
+          items:
+            $ref: "#/components/schemas/RequiredService"
+          maxItems: 4000
```
juliusfitzhugh-ccpo, line 447: Does there need to be a `type` added here or above where the `$ref` is called? If so I think there are some `ref`s that need a `type` too.

hoth_api.yaml
```diff
@@ -319,5 +319,351 @@ components:
...
+          enum:
+            - Unselected
+            - Yes
+            - No
+    ContractConsiderations:
+      description: "Part of DOW, represents Contract Considerations Table"
+      properties:
+        packaging_shipping_none_apply:
+          type: boolean
+        packaging_shippping_other:
```
juliusfitzhugh-ccpo, line 572:
```suggestion
        packaging_shipping_other:
```

hoth_api.yaml
```diff
@@ -319,5 +319,351 @@ components:
...
+      properties:
+        to_title:
+          type: string
+          maxLength: 60
+        scope:
+          type: string
+          maxLength: 300
+        scope_surge:
+          type: integer
+          maximum: 50
```
juliusfitzhugh-ccpo, line 333: I think data related to the contract award (`Award/Mod Effective` and `Version Date` columns) is missing, which is used for the first table in the DoW [see attachment in epic](https://ccpo.atlassian.net/browse/AT-7232).

hoth_api.yaml
```diff
@@ -319,5 +319,351 @@ components:
...
+        instance_location:
+          type: string
+          enum:
+            - CSP
+            - Hybrid
+            - Not in the cloud / On-premise
+        csp_region:
+          $ref: "#/components/schemas/TargetCloudServiceProviderName" #On the SNOW side, this is CSP_REGION a choice field with no options, so using TargetCSPName instead
+        performance_tier: # This is also an empty choice on the SNOW side, set as empty String until we get details
+          type: string
```
kylelaker-ccpo, line 379: Nitpick but end-of-line comments are kinda unpleasant to read (and might produce unexpected things in certain situations). Throughout, please modify comments to just be at the right level of indentation on the line before.

```suggestion
          # On the SNOW side, this is CSP_REGION a choice field with no options, so using TargetCSPName instead
          $ref: "#/components/schemas/TargetCloudServiceProviderName" 
        performance_tier:
           # This is also an empty choice on the SNOW side, set as empty String until we get details
          type: string
```

Also, much like in the rest of our code, let's use a space after the marker to start the comment (here, `#`).

hoth_api.yaml
```diff
@@ -319,5 +319,351 @@ components:
...
+            - Secret
+        impact_level:
+          type: string
+          nullable: true
+          enum:
+            - IL2
+            - IL4
+            - IL5
+            - IL6
+            - null
```
kylelaker-ccpo, line 430: What if we just matched https://github.com/dod-ccpo/atat-csp-orchestration/blob/main/provisioning/atat_provisioning.yaml#L459 and made this not required (rather than nullable)?

hoth_api.yaml
```diff
@@ -319,5 +319,351 @@ components:
...
+          type: array
+          items:
+            $ref: "#/components/schemas/Period"
+          maxItems: 4000
+        pop_start_request:
+          type: string
+          enum:
+            - Unselected
+            - Yes
+            - No
```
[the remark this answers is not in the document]
kylelaker-ccpo: Yeah let's not reinvent `enum Bool { True, False, FileNotFound }`  here [https://thedailywtf.com/articles/what_is_truth_0x3f_](https://thedailywtf.com/articles/what_is_truth_0x3f_)

hoth_api.yaml
```diff
@@ -319,5 +319,351 @@ components:
...
+    GFEOverview:
+      description: "Part of DOW, represents GFE Overview fields"
+      properties:
+        dpas_unit_id:
+          type: string
+          maxLength: 6
+        property_custodian_name:
+          type: string
+          maxLength: 40 # This is another instance of maxLength in SNOW being weird, the person might have a name longer than 40 characters
```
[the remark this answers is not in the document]
kylelaker-ccpo: So for _all_ of these `maxLength` things, does the Document Generation API _care_ at all? I mean, let's not use this to encode limitations of downstream systems. If we can support a system with 41 characters or 1000 characters then let's stop putting `maxLength: N` just because "that's what SNOW does here". Every `maxLength` should encode _validation_ data or a true limitation of our API. So for things like ID Numbers etc that we know have a specific length, then `maxLength` (and potentially a `minLength`) are very valuable. If this is a person's name, we absolutely must not limit it unless for some reason we'd fail to generate a document (or if we'd generate something invalid) with more characters or if >40char is an indication of an error.

hoth_api.yaml
```diff
@@ -319,5 +319,351 @@ components:
...
+          maxLength: 50
+        FOIA_country:
+          type: string
+          maxLength: 50
+        FOIA_street_address_1:
+          type: string
+          maxLength: 40
+        FOIA_street_address_2:
+          type: string
+          maxLength: 40
```
kylelaker-ccpo, line 630: Again, justify all the `maxLength` or remove them.

hoth_api.yaml
```diff
@@ -319,5 +319,351 @@ components:
...
+          type: string
+          maxLength: 1000
+    EnvironmentInstance:
+      description: "Part of DOW, represents a single Environment Instance"
+      properties:
+        instance_name:
+          type: string
+          maxLength: 100
+        classification_level:
+          $ref: "#/components/schemas/ClassificationLevel"
```
kylelaker-ccpo, line 369: This doesn't necessarily need to match other APIs' naming; however, we've been sorta all over the place with `ClassificationLevel` or `SecurityDomain` or other terminology. @jeffsegal-ccpo thoughts on how/whether this can/should match the CSP API?

hoth_api.yaml
```diff
@@ -319,5 +319,351 @@ components:
...
+        pricing_model_expiration:
+          type: string
+          format: date
+          example: "2022-07-01"
+        operating_system_licensing:
+          type: string
+          maxLength: 1000
+        number_of_vCPUs:
+          type: integer
+          maximum: 40
```
kylelaker-ccpo, line 395: This just feels like an odd limit. Is this a business rule? 64 vCPU (or more) is very common. Why would we limit this to 40?

hoth_api.yaml
```diff
@@ -319,5 +319,351 @@ components:
...
+        storage_type:
+          type: string # Empty choice on SNOW side, set as empty String until we get details
+        storage_amount:
+          type: integer
+          maximum: 40
```
kylelaker-ccpo, line 400: Paired with `storage_unit`, this means that I'd be be able to order 1 GB of storage, 40 GB of storage, 1 TB of storage or 40 TB of storage. But I can't order 41 GB? Hell of jump from 40GB to 1TB!

kylelaker-ccpo (changes requested):
I think we need to talk as a team about what exactly we want to encode in these API specs. I don't love the `maxLength`s here resulting in arbitrary limits on string and I don't love the `maximum`s that create nonsensical gaps in allowed memory/storage/egress amounts.

HOTH is intended to be an API that can be consumed by one or more downstream systems; today, that is defined as the application defined in `atat-snow`; however, it may not always be. This specification is meant to specify the capabilities of the HOTH application, **not** `atat-snow`. Just because `atat-snow` may only ever send a 40 char string doesn't mean that we should only ever support a 40 char string. Because some day, when Yet Another Downstream System wants to leverage HOTH, it may have a user with an address that is 51 characters and what are we going to do then? Dump a 4xx back? That's just weird.

We should be specifying what data we can accept and process and return valid results for. I am absolutely not convinced we'll implode with 41 character name. We should be paying careful attention to the data formats we have/need and not just building something that only ever works with the `atat-snow` app because well, then why would a specification even be necessary in the first place?

Let's talk on/after our standup today or another nearby timeframe.

hoth_api.yaml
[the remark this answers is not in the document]
jasonburkert-ccpo: @juliusfitzhugh-ccpo That is accurate to my knowledge.  I'd only add that they did this because False was the default value and they wanted to avoid that in some cases.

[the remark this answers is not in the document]
jeffsegal-ccpo: And the guiding principle for these objects that we're specifying here is that they will be populated _from_ SNOW data and then passed on to HOTH. In other words, it should be aimed at consumption by HOTH (and adapted if necessary, e.g. `"Unselected"` becomes `null`) rather than catering towards SNOW.

hoth_api.yaml
zachclark-ccpo: @juliusfitzhugh-ccpo Good call, although OpenAPI doesn't require a type, it is not inferred from the `properties` keyword. I remember this being an issue in the past when converting to JSON. I will update the spec to be explicit about `type: object`

zachclark-ccpo:
When doing a final check of the schema, I noticed that `PerformanceRequirements` is just a wrapper array that could be moved to the top level.
```
    PerformanceRequirements:
      description: >
        Part of DOW, encapsulates all data for 4. Performance Requirements and its subtasks. 
        On the API side, the subtasks will filter through the provided required_services.service_offerings
        to create the subtasks fields.
      type: object
      properties:
        required_services:
          type: array
          items:
            $ref: "#/components/schemas/RequiredService"
```
Was removed, replaced with:
```
        # DOW section 4 (and its subtasks)
        required_services:
          type: array
          items:
            $ref: "#/components/schemas/RequiredService"
```
Which lives in the `DescriptionOfWork` schema

hoth_api.yaml
```diff
@@ -319,5 +319,351 @@ components:
...
+          $ref: "#/components/schemas/StorageUnit"
+    ClassificationLevel:
+      description: "Part of DOW, Classification Level table fields"
+      properties:
+        classification:
+          type: string
+          enum:
+            - Unclassified
+            - Top Secret
+            - Secret
```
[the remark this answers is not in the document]
zachclark-ccpo: Need to capitalize the `enum`s anyways, going to implement this change.

zachclark-ccpo:
Example schema:
```
{
  "award_history": [
    {
      "contract_award_type": "INITIAL_AWARD",
      "modification_order": 0,
      "effective_date": "2022-05-25"
    }
  ],
  "contract_information": {
    "contract_number": "string",
    "current_contract_exists": true,
    "contract_expiration_date": "2022-05-25",
    "incumbent_contractor_name": "string",
    "task_order_number": "string",
    "previous_task_order_number": "string"
  },
  "to_title": "string",
  "scope": "string",
  "scope_surge": 0,
  "current_environment": {
    "current_environment_exists": true,
    "environment_instances": [
      {
        "instance_name": "string",
        "classification_level": {
          "classification": "U",
          "impact_level": "IL2"
        },
        "instance_location": "CSP",
        "csp_region": "CSP_A",
        "performance_tier": "string",
        "pricing_model": "ON_DEMAND",
        "pricing_model_expiration": "2022-07-01",
        "operating_system_licensing": "string",
        "number_of_vCPUs": 0,
        "storage_type": "string",
        "storage_amount": 0,
        "storage_unit": "GB",
        "memory_amount": 0,
        "memory_unit": "GB",
        "data_egress_monthly_amount": 0,
        "data_egress_monthly_unit": "GB"
      }
    ],
    "additional_info": "string"
  },
  "required_services": [
    {
      "applicable_classification_levels": [
        {
          "classification": "U",
          "impact_level": "IL2"
        }
      ],
      "usage_description": "string",
      "need_for_entire_to_duration": true,
      "applicable_periods": [
        {
          "period_type": "BASE",
          "period_unit_count": 0,
          "period_unit": "DAY",
          "option_order": 0
        }
      ],
      "service_offerings": [
        {
          "service_offering_group": "ADVISORY",
          "name": "string"
        }
      ],
      "other_service_offering": "string"
    }
  ],
  "period_of_performance": {
    "base_period": {
      "period_type": "BASE",
      "period_unit_count": 0,
      "period_unit": "DAY",
      "option_order": 0
    },
    "option_periods": [
      {
        "period_type": "BASE",
        "period_unit_count": 0,
        "period_unit": "DAY",
        "option_order": 0
      }
    ],
    "pop_start_request": true,
    "requested_pop_start_date": "2021-07-01",
    "time_frame": "NO_LATER_THAN",
    "recurring_requirement": true
  },
  "gfe_overview": {
    "dpas_unit_id": "string",
    "property_custodian_name": "string",
    "dpas_custodian_number": "string",
    "property_accountable": true,
    "gfe_or_gfp_furnished": true
  },
  "contract_considerations": {
    "packaging_shipping_none_apply": true,
    "packaging_shipping_other": true,
    "packaging_shipping_other_explanation": "string",
    "potential_conflict_of_interest": true,
    "conflict_of_interest_explanation": "string",
    "contractor_provided_transfer": true,
    "contractor_required_training": true,
    "required_training_services": {
      "name": "string",
      "value": "string"
    }
  },
  "section_508_accessibility_standards": {
    "pii_present": true,
    "work_to_be_performed": "string",
    "system_of_record_name": "string",
    "FOIA_city_apo_fpo": "string",
    "FOIA_country": "string",
    "FOIA_street_address_1": "string",
    "FOIA_street_address_2": "string",
    "FOIA_address_type": "FOREGIN",
    "FOIA_state_province_code": "string",
    "FOIA_full_name": "string",
    "FOIA_email": "user@example.com",
    "FOIA_zip_postal_code": "string",
    "BAA_required": true,
    "potential_to_be_harmful": true,
    "section_508_sufficient": true,
    "accessibility_reqs_508": "string"
  }
}
```

jasonburkert-ccpo (approved):
Award history and contract information included. Enum values are improved.  Good changes.

hoth_api.yaml
zachclark-ccpo: Slack overflow reference to what I was talking about: https://stackoverflow.com/questions/47374980/schema-object-without-a-type-attribute-in-swagger-2-0/47390723#47390723

kylelaker-ccpo (approved):

PR #879, opened 2022-05-23
Support configurable CIDR blocks

jeffsegal-ccpo:
This PR satisfies AT-7448 by adding a `VpcCidr` context parameter to our CDK config. This parameter has the following rules:

1. If the given Environment is a Sandbox, `VpcCidr` must NOT be defined.
2.If the given Environment is NOT a Sandbox, `VpcCidr` must be a valid CIDR block such as `10.100.0.0/20`.

I also moved the majority of the logic out of `bin/atat-web-api.ts` to `bin/app.ts` so that we could unit test it without invoking `app.synth()`. In doing so I changed the `process.exit()` calls to throw Errors instead. I added a number of test cases for input validation but did not include any CDK level validations (e.g. of generated CloudFormation templates).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -46,7 +46,7 @@ jobs:
         run: |
           npm ci
       - name: Synthesize CDK template
-        run: cdk synth -c "atat:EnvironmentId=CiTest"
+        run: cdk synth -c "atat:EnvironmentId=CiTest" -c "atat:Sandbox=true"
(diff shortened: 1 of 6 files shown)


kylelaker-ccpo (approved):
Thanks for this! It looks awesome. So let's flip this back to `develop` for the `atat:VersionControlBranch` context, merge it, and then manually deploy once to Sandbox and Dev.

PR #906, opened 2022-06-01
Require MFA when assuming all roles

kylelaker-ccpo:
Using this will require that everyone add `mfa_serial = arn:aws-us-gov:iam::<ACCOUNT_ID>:mfa/<USERNAME>` to _every_ block in their `~/.aws/config` file and to enter an MFA code each time they assume a role (well, unless their CLI was kind enough to cache the credentials and they're still within the `RoleSessionDuration`).


kylelaker-ccpo:
The parts of this that we actually need long term have been integrated into #929

PR #927, opened 2022-06-07
Add two cost queues to infrastructure

juliusfitzhugh-ccpo:
Create new `Queue` construct.
Add two queues to the atat-web-api-stack
- `costRequestQueue`
- `costResponseQueue`
- add unit test for queue

Update the provisioning queue to use new `Queue` construct.

Ticket: AT-7453

### DIFF
--- a/lib/constructs/sqs.ts
+++ b/lib/constructs/sqs.ts
@@ -1,0 +1,34 @@
+import * as sqs from "aws-cdk-lib/aws-sqs";
+import { QueueProps } from "aws-cdk-lib/aws-sqs";
+import { Construct } from "constructs";
+
+export interface SQSProps extends QueueProps {
+  /**
+   * Props to override the default props of the created SQS
+   */
+  overrideSqsProps?: QueueProps;
+  /**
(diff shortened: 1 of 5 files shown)


lib/constructs/sqs.ts
```diff
@@ -0,0 +1,34 @@
...
+   * Environment name to help differentiate the queues made
+   */
+  environmentName: string;
+}
+
+/**
+ * A basic SQS queue
+ * - defaults to an FIFO queue
+ */
+export class Queue extends Construct {
```
jeffsegal-ccpo, line 20: What do we think about naming our wrapper constructs with an `Atat` prefix? E.g. `AtatQueue` just to be clear that it's including some ATAT-standard configs and behavior?

juliusfitzhugh-ccpo: I like it that works for me. I originally had `SQS` and thought that wasn't really good, and `Queue` seemed a little better. But I like `AtatQueue` it seems just right, haha.

kylelaker-ccpo (approved):

PR #929, opened 2022-06-07
Add tooling to support STIG validation and remediation

kylelaker-ccpo:
This adds a template based on the [Operational Best Practices for NIST
800-53 rev4][1]. Some rules have been disabled because they are not
supported in us-gov-west-1. Eventually, rather than commenting them out,
we may want to consider using a `Condition` based on the region.

[1]: https://github.com/awslabs/aws-config-rules/blob/master/aws-config-conformance-packs/Operational-Best-Practices-for-NIST-800-53-rev-4.yaml.


manual-steps/stig-remediation.ts
```diff
@@ -0,0 +1,43 @@
...
+    // V-222536
+    MinimumPasswordLength: 15,
+    PasswordReusePrevention: 24,
+    RequireLowercaseCharacters: true,
+    RequireNumbers: true,
+    RequireSymbols: true,
+    RequireUppercaseCharacters: true,
+  });
+}
+
```
jeffsegal-ccpo, line 41: Are we assuming that this script will be invoked by someone who assumes the `CompliantFrameworkAccountAccessRole` role?

kylelaker-ccpo: Yep -- that will be included in the docs that I plan to finish writing tomorrow.

cfn-templates/management-account-iam.yaml
```diff
@@ -0,0 +1,181 @@
...
+      Description: Allow users to modify their own IAM configuration
+      PolicyDocument:
+        Version: "2012-10-17"
+        Statement:
+          - Sid: AllowViewAccountInfo
+            Effect: Allow
+            Action:
+              - iam:GetAccountPasswordPolicy
+              - iam:ListVirtualMFADevices
+            Resource: "*"
```
kylelaker-ccpo, line 35: Sonar was grumpy about this. I marked this as safe with the following justification:

> Both of the actions in this statement are required in order for users to be able modify their own IAM permissions. Neither of these actions support filtering based on resources (the account password policy doesn't have an ARN and the other call is a List action) according the Service Authorization Reference (https://docs.aws.amazon.com/service-authorization/latest/reference/list_identityandaccessmanagement.html).

kylelaker-ccpo:
There isn't a meaningful way to write a test for the added script (which is what Sonar is upset about for coverage). It'd be a test that just did a mock to make sure that the exact same static values get set the ones in the root. And we'd have some weird complexity because S3 Control `PutPublicAccessBlock` is a wonky API call that requires passing the Account ID (even though it's really unclear when, if ever, you can pass a _different_ Account ID so that's a thing I'd love to learn at some point) and so we've gotta do an STS `GetCallerIdentity` call. And anyway it just doesn't feel meaningful to write out the password policy twice.

manual-steps/stig-remediation.ts
kylelaker-ccpo: Eventually, by the way (now that I think about it more), I'd like to deploy this as a custom resource via a CloudFormation StackSet from the central account of the AWS Organization.

PR #935, opened 2022-06-09
AT-7454: Added CostRequest & CostResponse schema objects

jeffsegal-ccpo:
This PR simply defines the schema to be used in other tickets. Ended up re-using `CspResponse` rather than recreating the wheel for `CostResponse`.

### DIFF
--- a/hoth_api.yaml
+++ b/hoth_api.yaml
@@ -315,6 +315,35 @@ components:
         content:
           type: object
           additionalProperties: true
+    CostRequest:
+      description: >-
+        Represents a request for Cost data to be fulfilled by the ATAT Orchestrator by invoking the appropriate /costs
+        API for a given Cloud Service Provider
+      properties:
+        request_id:
+          description: ID to enable correlating Cost reqeusts with responses
+          type: string
+          format: uuid
+        portfolio_id:
+          type: string
+        target_csp:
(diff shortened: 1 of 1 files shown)


hoth_api.yaml
```diff
@@ -315,6 +315,35 @@ components:
         content:
           type: object
           additionalProperties: true
+    CostRequest:
```
zachclark-ccpo, line 318: I would add
```
type: object
```
under the `description`

hoth_api.yaml
```diff
@@ -315,6 +315,35 @@ components:
...
+          description: Start date of Cost query
+          type: string
+          format: date
+          example: "2021-07-01"
+        end_date:
+          description: End date of Cost query
+          type: string
+          format: date
+          example: "2022-07-01"
+    CostResponse:
```
zachclark-ccpo, line 341: I would also add `type: object` to the **CostResponse**

zachclark-ccpo (commented):
Looks good, just be sure to explicitly declare`type: object` , this will help if we ever need to use this schema in JSON format and its OpenAPI 3.0 best practice.

hoth_api.yaml
jeffsegal-ccpo: done

zachclark-ccpo (approved):

PR #941, opened 2022-06-14
Update dependencies

kylelaker-ccpo:


juliusfitzhugh-ccpo:
@kylelaker-ccpo good call. The `puppeteer-core` and `@sparticuz/chrome-aws-lambda` dependences need to have the same version since I believe the chromium version required could be different for `puppeteer-core`. I'll keep an eye out and open up a PR once `@sparticuz/chrome-aws-lambda` is upgraded to `v14.4.0` (which has been kept update recently https://github.com/Sparticuz/chrome-aws-lambda/commits/master).

juliusfitzhugh-ccpo (approved):

PR #944, opened 2022-06-14
Bundle puppeteer-core with the generate-document fn

juliusfitzhugh-ccpo:
Add `puppeteer-core` as dependency in the bundling of the generate
document lambda, due to the below error. The error causes the
`ATAT Generate Description of Work` subflow to error.
```
Cannot find module 'puppeteer-core/lib/cjs/puppeteer/common/Browser.js'
```

**Note:** This was not included previously, because it did not appear to be required and worked in sandbox without the addition of `puppeteer-core` being bundled.


kylelaker-ccpo (approved):

juliusfitzhugh-ccpo:
Deployed a fresh stack (without auth) in sandbox, and it works as expected in the `ATAT Generate Description of Work` subflow.

PR #977, opened 2022-06-23
Update dependencies

kylelaker-ccpo:
This:
 - Updates dependencies to the latest version
 - Removes `cdk-iam-floyd`
 - Uses new features of the logger
 - Applies user identity persistent attributes to _all_ loggers when possible
 - Removes all `console.{log,error}` statements


kylelaker-ccpo:
Tried to adopt v0.11.0 of `@aws-lambda-powertools/logger` but now I am concerned that may result in duplicate event logs...

juliusfitzhugh-ccpo:
@kylelaker-ccpo I see your concern, I'm guessing you are thinking the new `logEvent` will log in addition to the `inputOuputLogger` middleware? Did you test it out already, if not I can try to run it before it on a deployed stack.

kylelaker-ccpo:
@juliusfitzhugh-ccpo since we have the logger to capture all events/results, i removed `logEvent: true` but kept `clearState` so that we don't keep persistent attributes between invocations (so that we can add user identity information)

juliusfitzhugh-ccpo (approved):
Looks good. Its a bummer that `cdk-iam-floyd` had to go, but the cdk `iam` library isn't so bad either. Also, thanks for cleaning up those `console.log/error` (I think they were mine).

Ran some tests on the `provisioning-jobs`, `cost-jobs` and `generate-document` endpoints. Looks like things are working as expected and logging with the new logger updates.

utils/middleware/logging-context-middleware.ts
```diff
@@ -0,0 +1,22 @@
...
+import { APIGatewayProxyResult } from "aws-lambda";
+import { CommonMiddlewareInputs } from "./common";
+import { logger } from "../logging";
+
+// Create the middleware and export it
+export const LoggingContextMiddleware = (): middy.MiddlewareObj<CommonMiddlewareInputs, APIGatewayProxyResult> => {
+  // Set up a before check, this will run before the handler
+  const before: middy.MiddlewareFn<CommonMiddlewareInputs, APIGatewayProxyResult> = async (request): Promise<void> => {
+    // log the sourceIp
+    logger.setPersistentLogAttributes({
```
jeffsegal-ccpo, line 11: This is pretty neat feature.

jeffsegal-ccpo (approved):
Looks good. My main concern is obviously regression testing in terms of all the IAM changes.

utils/middleware/logging-context-middleware.ts
```diff
@@ -0,0 +1,22 @@
+import middy from "@middy/core";
+import { APIGatewayProxyResult } from "aws-lambda";
+import { CommonMiddlewareInputs } from "./common";
+import { logger } from "../logging";
+
+// Create the middleware and export it
+export const LoggingContextMiddleware = (): middy.MiddlewareObj<CommonMiddlewareInputs, APIGatewayProxyResult> => {
```
jasonburkert-ccpo, line 7: Nice improvement.  Appreciate the more general nature of this logging middleware vice the IP logging middleware's console output.

lib/atat-iam-stack.ts
```diff
@@ -1,7 +1,6 @@
 import * as cdk from "aws-cdk-lib";
 import * as iam from "aws-cdk-lib/aws-iam";
 import * as custom from "aws-cdk-lib/custom-resources";
-import * as statement from "cdk-iam-floyd";
```
jasonburkert-ccpo, line 4: Unfortunate we had to un-floyd this.  So many additional lines.  But at least we can be more confident this won't break on the next CDK release.

document-generation/chromium.ts
```diff
@@ -20,7 +20,6 @@ export async function generateDocument(document: string): Promise<Buffer | undef
     await page.setContent(document);
     await page.emulateMediaType("screen");
     generatedDocument = await page.pdf(options);
-    console.log("DOCU: ", "generateDocument");
```
jasonburkert-ccpo, line 23: Thanks for the cleanup on things like this

jasonburkert-ccpo (approved):
Thank you for the effort

jasonburkert-ccpo:
> Looks good. Its a bummer that `cdk-iam-floyd` had to go, but the cdk `iam` library isn't so bad either. Also, thanks for cleaning up those `console.log/error` (I think they were mine).
> 
> Ran some tests on the `provisioning-jobs`, `cost-jobs` and `generate-document` endpoints. Looks like things are working as expected and logging with the new logger updates.

@juliusfitzhugh-ccpo Thank you for the thorough testing

kylelaker-ccpo:
> Looks good. My main concern is obviously regression testing in terms of all the IAM changes.

I agree, fortunately, our existing tests actually caught the mistakes that I made! And the cleanup to the `artifact` service permissions is basically equivalent and the only real change in permissions that I am aware of. Hopefully we will identify any regressions quickly if there are any in real-world testing.

PR #1007, opened 2022-07-05
Refactor CDK constructs

kylelaker-ccpo:
This refactors a pretty significant chunk of constructs within the
application to just directly extend other components rather than build
compositions based on them. For a few things, we're basically just
creating pre-configured variants of another construct type (SQS Queue,
State Machine, etc) and we can handle that pretty trivially and remove
some need to make wonky `.foo` attribute accesses. This also makes them
easier to just pass around within the CDK app.

Notably, the API Gateway REST API is not refactored. That adds a few
methods and my worry there is that we actually use some names that may
eventually conflict with the underlying resource (not likely but
possible) and so I kinda tried to avoid that.

Because the `SfnLambdaInvokeTask` was only ever created in one place, we
can also just get rid of it entirely. It wasn't doing a whole lot for
us. I think it made sense in a time where its use case was uncertain and
where we had more constructs with that pattern; today, I think we can
clean it up a bit.

This **will** result in resource replacement in already-deployed
environments as constuct paths (and therefore CloudFormation resource
logical IDs) will change. I feel that because these are mostly Queues
and connecting resources this is safe but there's a chance this will gum
up the pipeline and we'll either need to make some additional fixes or
revert.

### DIFF
--- a/lib/constructs/provisioning-sfn-workflow.ts
+++ b/lib/constructs/provisioning-sfn-workflow.ts
@@ -10,8 +10,8 @@ import { HttpMethod } from "../http";
 import { ApiSfnFunction } from "./api-sfn-function";
 import { IdentityProviderLambdaClient, IIdentityProvider } from "./identity-provider";
 import { mapTasks, TasksMap } from "./sfn-lambda-invoke-task";
-import { AtatQueue } from "./sqs";
-import { StateMachine } from "./state-machine";
+import { FifoQueue } from "./sqs";
+import { LoggingStandardStateMachine } from "./state-machine";
 
 /**
  * Successful condition check
@@ -56,7 +56,7 @@ export class ProvisioningWorkflow extends Construct implements IProvisioningWork
     super(scope, id);
     const { environmentName } = props;
 
-    this.provisioningJobsQueue = new AtatQueue(scope, "ProvisioningJobsQueue", { environmentName }).sqs;
+    this.provisioningJobsQueue = new FifoQueue(scope, "ProvisioningJobsQueue");
 
     // Provisioning State machine functions
     const cspConfig = secrets.Secret.fromSecretNameV2(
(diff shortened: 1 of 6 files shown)


kylelaker-ccpo:
This will also enable KMS encryption for our SQS queues. Should be fine

jeffsegal-ccpo:
> 

Ah thanks for catching that.

jeffsegal-ccpo (approved):
Looks good. Thanks for the cleanup.

lib/constructs/provisioning-sfn-workflow.ts
```diff
@@ -56,7 +56,7 @@ export class ProvisioningWorkflow extends Construct implements IProvisioningWork
     super(scope, id);
     const { environmentName } = props;
 
-    this.provisioningJobsQueue = new AtatQueue(scope, "ProvisioningJobsQueue", { environmentName }).sqs;
+    this.provisioningJobsQueue = new FifoQueue(scope, "ProvisioningJobsQueue");
```
juliusfitzhugh-ccpo, line 59: Is it okay removing the `environmentName`? I know there have been some name collisions in the past with similar stack resources not being able to be deployed if one stack had a similar name.

lib/constructs/sfn-lambda-invoke-task.ts
```diff
@@ -45,9 +17,14 @@ export type TasksMap = { [name: string]: SfnLambdaInvokeTask };
...
+  return tasks.reduce(
+    (mappedTasks, task) => ({
+      ...mappedTasks,
+      [task.id]: new sfnTasks.LambdaInvoke(scope, task.id, {
+        timeout: Duration.seconds(20),
+        ...task.props,
+      }),
+    }),
+    {}
+  );
```
juliusfitzhugh-ccpo, line 29: Nice, I like this. It's a much cleaner approach using `reduce`.

juliusfitzhugh-ccpo (approved):
Looks good, thanks for refactoring. I like the reduction in the attribute access, since I got caught missing one of them when testing 😆. Only one question to make sure I understand.

PR #1011, opened 2022-07-07
Add logging to CostRequest fn

juliusfitzhugh-ccpo:
- add logs
- enable DEBUG level on logger
- fix typo

Ticket: AT-7457


kylelaker-ccpo (approved):
We'll need to eventually revert some of this.

PR #1015, opened 2022-07-07
Fix networking issues with AWS Secrets Manager

kylelaker-ccpo:
The bulk of the issue is described in the code comment. Putting this
explanation inline makes it more immediately clear to future
maintainers, reviewers, or auditors why we've omitted the flag in this
particular case.

All other services we invoke from within our VPC use the same hostname
for FIPS and non-FIPS so keeping `useFipsEndpoint` is fine for those
(SQS and Step Functions).


jeffsegal-ccpo (approved):
Appreciate the explanation.

PR #1016, opened 2022-07-07
Make modifications based on recent Lambda error

kylelaker-ccpo:
See the individual commits within the PR for more details on each change.

- Allow developers to purge SQS queue
- Log all DNS queries to CloudWatch Logs

### DIFF
--- a/lib/atat-iam-stack.ts
+++ b/lib/atat-iam-stack.ts
@@ -117,7 +117,7 @@ export class AtatIamStack extends cdk.Stack {
         // troubleshoot
         new iam.PolicyStatement({
           effect: iam.Effect.ALLOW,
-          actions: ["sqs:Get*", "sqs:List*", "sqs:ReceiveMessage"],
+          actions: ["sqs:Get*", "sqs:List*", "sqs:ReceiveMessage", "sqs:PurgeQueue"],
           resources: ["*"],
         }),
         // Grant access to read the pipeline state
--- a/lib/atat-net-stack.ts
+++ b/lib/atat-net-stack.ts
@@ -2,6 +2,7 @@ import * as cdk from "aws-cdk-lib";
 import * as ec2 from "aws-cdk-lib/aws-ec2";
 import * as logs from "aws-cdk-lib/aws-logs";
 import * as custom from "aws-cdk-lib/custom-resources";
+import * as route53Resolver from "aws-cdk-lib/aws-route53resolver";
 
 import { Construct } from "constructs";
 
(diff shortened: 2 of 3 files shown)


lib/atat-iam-stack.ts
```diff
@@ -117,7 +117,7 @@ export class AtatIamStack extends cdk.Stack {
         // troubleshoot
         new iam.PolicyStatement({
           effect: iam.Effect.ALLOW,
-          actions: ["sqs:Get*", "sqs:List*", "sqs:ReceiveMessage"],
+          actions: ["sqs:Get*", "sqs:List*", "sqs:ReceiveMessage", "sqs:PurgeQueue"],
```
juliusfitzhugh-ccpo, line 120: Thank you, I will use this with care 🙂. The `sqs:PurgeQueue` appears to wipe out the whole queue, which is probably fine, but I was wondering if `sqs:DeleteMessage` might be a less destructive permission since you can select individual messages (assuming other ones shouldn't be touched)?

lib/atat-net-stack.ts
```diff
@@ -55,12 +62,17 @@ export class AtatNetStack extends cdk.Stack {
...
+    const dnsLogsGroup = new logs.LogGroup(this, "VpcDnsQueryLogs", {
+      retention: logs.RetentionDays.INFINITE,
+    });
+    // Capture all DNS queries made by all hosts in the VPC
+    const dnsLoggingConfig = new route53Resolver.CfnResolverQueryLoggingConfig(this, "DnsLogging", {
+      destinationArn: dnsLogsGroup.logGroupArn,
+    });
+    const vpcDnsLogging = new route53Resolver.CfnResolverQueryLoggingConfigAssociation(this, "VpcDnsLogging", {
+      resolverQueryLogConfigId: dnsLoggingConfig.attrId,
+      resourceId: vpc.vpcId,
```
juliusfitzhugh-ccpo, line 74: Cool! So this is lower level logging? Are the logs from the resolver similar (or the same) logs that you were showing us in `Logs Insights`?

juliusfitzhugh-ccpo (approved):
Thanks for adding these. I left a few questions, but otherwise I don't see any issues.

lib/atat-iam-stack.ts
kylelaker-ccpo: `PurgeQueue` is a button in the Console which helps :) sometimes the messages get stuck "in-flight" between the failures and `PurgeQueue` seems to be a little better at dealing with that.

PR #1018, opened 2022-07-08
AT-7458: Granting Developers the sqs:SendMessage action

jeffsegal-ccpo:
This will enable easier testing of functionality that relies on sending messages to a particular Queue.


kylelaker-ccpo (approved):
We need to roll this back at the close of this sprint or next.

PR #1019, opened 2022-07-08
Add monitoring for nearly all resources

kylelaker-ccpo:


kylelaker-ccpo:
This now uses some of the same configuration that was created in #1031. We still use a separate SNS Topic because it serves a different purpose. Some day, these may need to go to different email addresses but today, that doesn't feel like a pressing need and would add additional complexity in configuration and code logic.

I like that we can monitor the notification stack created in #1031 too! Kinda cool.

jasonburkert-ccpo (approved):
Thanks for these improvements!

PR #1028, opened 2022-07-12
Make a more general way to give developers permission

kylelaker-ccpo:


jeffsegal-ccpo (approved):
This should be very useful for dev, thanks.

PR #1042, opened 2022-07-15
Update Security Group rules for API Gateway

kylelaker-ccpo:
This also disables VPC DNS Query Logs for now until we work out some IAM
and resource policy issues.


zachclark-ccpo (approved):

kylelaker-ccpo:
I am commenting out that block of code because I plan to either fix it or remove it later today.

PR #1043, opened 2022-07-15
Support custom domains for the API Gateway

kylelaker-ccpo:
This adds support for custom domains to the API Gateway. The way that
this is handled requires changes to a few core constructs within the
environment as well as a little bit of gluing AWS resources together.
First, API Gateway does not directly support custom domains on private
endpoints. This requires us to put _something_ in front of it that can
speak HTTPS with a valid certificate for a custom name. This uses a load
balancer. Because of issues experienced during troubleshooting, it was
easier to just add support for _both_ types of Load Balancers and we can
toggle between them in the future as needed.

Now, as for how we implement the Load Balancer's Target Group... that
requires a little more involvement. To start with, ELBv2 only supports a
few types of targets by default: EC2 Instances, IP Addresses, Lambda
Functions, and ALBs. Unfortunately, a VPC Endpoint (which is what we
have with our API Gateway Private Endpoints) isn't any of those things
directly. But it does have an IP address (well... it actually has
several). Unfortunately, no tooling directly exposes the IP addresses
based on VPC Endpoint ID; especially not CloudFormation/CDK.
Fortunately, with Custom Resources, we can glue together several AWS API
calls to describe the endpoint and get its associated Elastic Network
Interfaces (ENIs) and then describe those to get their IP addresses
(because despite ENIs being a lovely abstraction in EC2/AWS, an ELBv2
cannot target them).

Unfortunately, the CDK is very helpful. And by being so helpful, it
provides a lot of abstractions around ELBv2 Target Groups. But this
breaks the ability to just yeet arbitrary JSON into the `Targets`
property of the `AWS::ElasticLoadBalancingV2::TargetGroup` resource. No
matter what you try to do, the CDK either wraps it in a list or very
(not-so) helpfully tries to validate it (which won't work because with
our custom resource, it ends up being a `Fn::GetAtt`). And because it's
an intrinsic function and not actual raw data, we can't treat it like an
object and split it, iterate over it, or any other fun magic (in the
CDK, it's a `Token`). So we have to just treat it as an opaque object.
The only way we can effectively provide that as an opaque object is to
use escape hatches in the CDK to pass it directly to the underlying
CloudFormation resource property. That means that our Custom Resource
has to return a valid JSON object that can be passed to the `Targets`
object.

The configured Load Balancer will then balance across all the IP address
(and hopefully does so somewhat intelligently based on availability zone
since we do provide that data). The HTTPS/TLS certificate from ACM is
attached to the load balancer and that results in the certificate
matching the DNS name.

This does require two changes for consumers of the API:

 1. Consumers must now use the DNS name. The regional endpoint for the
    API is no longer available
 2. Consumers must pass the `x-apigw-api-id` header, as described in the
    [API Gateway docs for invoking via the DNS name][1]

Finally, this should not cause any issues for Developers' sandbox
environments. Those should be able to trivially be built without DNS
names and without VPC connectivity.

This implementation does have a few trade-offs as it closely ties the
usage of a custom domain to the usage of a VPC. For our use case, that
is likely pretty acceptable since any environments where we'd specify a
custom name, we'd also have a VPC. Second, we may need to kick off
something to re-trigger the custom resource at some point in the future
if we for any reason see the underlying IPs change without the VPCE ID
changing (but that is extremely unlikely to happen).

One option that was available that was rejected for now was having the
ALB target a Lambda to handle balancing between the VPC endpoints. This
would have resulted in additional code needing to be executed on _every_
request (so in many cases, doubling our number of Lambda function
invocations) and would have just added another link in the chain of
request/response size limitations that would become somewhat harder to
debug. Not using the Lambda option also allows us to maintain the
ability to use either an NLB or an ALB, whichever is most appropriate.

[1]: https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-private-api-test-invoke-url.html#apigateway-private-api-public-dns


bin/app.ts
```diff
@@ -11,13 +11,28 @@ export function createApp(props?: cdk.AppProps): cdk.App {
...
 
   if (!utils.isString(environmentParam)) {
     const err = "An EnvironmentId must be provided (use the atat:EnvironmentId context key)";
     console.error(err);
     throw new Error(err);
   }
 
+  let apiCertOptions: ApiCertificateOptions | undefined;
+
+  if (utils.isString(apiDomainParam) !== utils.isString(apiCertParam)) {
```
zachclark-ccpo, line 25: Good usage of `!==! here.

zachclark-ccpo:
> Finally, this should not cause any issues for Developers' sandbox
environments. Those should be able to trivially be built without DNS
names and without VPC connectivity.

When doing sandbox development we now need to pass the `x-apigw-api-id` header, and will need to update existing SNOW functions/actions/flows to include this header (we should store it in the ATAT Configuration Table).

lib/atat-web-api-stack.ts
```diff
@@ -62,11 +72,67 @@ export class AtatWebApiStack extends cdk.Stack {
...
     }
 
     const api = new AtatRestApi(this, "HothApi", apiProps);
+    if (props.apiDomain && network) {
+      const certificate = acm.Certificate.fromCertificateArn(this, "ApiCertificate", props.apiDomain.acmCertificateArn);
+      /* eslint-disable spaced-comment */
+      // Toggle between the ALB or the NLB by adding (ALB) or removing (NLB) the leading `/` on
+      // the following line. This is a _temporary_ functionality; only one will remain in the
+      // code base after this PR is merged.
+      //*
```
juliusfitzhugh-ccpo, line 85: This is a really neat toggle by removing the `/`. I had to actually do it to fully understand 😆

kylelaker-ccpo:
> When doing sandbox development we now need to pass the `x-apigw-api-id` header

I don't know that's correct -- Sandbox environments won't have a VPC or a Domain configuration which means that they won't use private endpoints or a custom domain. You _would_ need to pass that header for Dev (and I mean, if you tested against `sandbox.atat.dev` but more for the most part, we'll be using `atat.dev`).

> Should we store the API ID (in ATAT Configuration) or add a step to retrieve the API ID?

I don't want to start teaching SNOW to invoke arbitrary AWS APIs. I don't want to give the Users we create permission to do that either. I'd rather that we configure it statically. I don't think that the API Gateway ID should change very often (because this is the ID for the API GW API resource itself, not for the deployment or stage). I don't know how often this changes. I would be interested to find out (but it's probably too late to gather past data).

My preference would be to add this to the set of data we have to fetch/gather from the AWS console/API after we've created an environment. Perhaps we can find a way to notify ourselves if this chagnes.

kylelaker-ccpo:
One thing I want to note: getting the API GW API ID isn't really that much more of a lift that what we had to do before -- it's the first part of the `<api-id>.execute-api.<region>.amazonaws.com` URL. So we already were doing this; this just isolates it to us needing the first bit of that URL

lib/custom-resources/endpoint-ips.ts
```diff
@@ -0,0 +1,63 @@
...
+  const interfaceIds = (
+    await ec2.describeVpcEndpoints({
+      Filters: [
+        {
+          Name: "vpc-endpoint-id",
+          Values: [vpcEndpointId],
+        },
+      ],
+    })
+  )?.VpcEndpoints?.flatMap((endpoint) => endpoint.NetworkInterfaceIds).filter((eni): eni is string => !!eni);
```
juliusfitzhugh-ccpo, line 52: That's a cool use of `filter` with the `eni is string => !!eni`. Never seen it done like that before.

juliusfitzhugh-ccpo (commented):
Thanks for the explanation in the PR and the comments, it definitely helped. I now understand the complexity that was involved. From what I wrapped my head around, I didn't see anything that stood out. The custom resource to get the vpc endpoints was a creative work around. Nice work!

lib/constructs/vpc-endpoint-lb-target.ts
```diff
@@ -0,0 +1,170 @@
...
+
+  addTarget(...targets: elbv2.IApplicationLoadBalancerTarget[]): void {
+    throw new Error("Additional targets are not supported for VpcEndpointApplicationTargetGroup");
+  }
+}
+
+/**
+ * Allows using a VPC Endpoint as the target for a Network Load Balancer.
+ */
+export class VpcEndpointNetworkTargetGroup
```
jeffsegal-ccpo, line 151: Really like the use of OO design to cleanly support both types of ELBs (perhaps showing my bias as a Java guy at heart)

lib/custom-resources/endpoint-ips.ts
```diff
@@ -0,0 +1,69 @@
...
```
jeffsegal-ccpo, line 58: Is this a little trick to name anonymous functions?

jeffsegal-ccpo (changes requested):
Really great stuff, particularly in simplifying what is a pretty complex process to the greatest extent you could. I've got one small TODO request and one question - what's the mechanism for actually updating the target group of the ELB? I see that `endpoint-ips.ts` returns the IP, but I wanted to confirm my suspicion that this is just standard Custom Resource behavior for a create or update - that is, returning a modified object in response to an event actually changes the underlying resource. True?

lib/custom-resources/endpoint-ips.ts
kylelaker-ccpo: Not sure which part of this you're talking about but to expand this line out:

```ts
VpcEndpoints // Get the VPC endpoints
  // If VpcEndpoints exists, flatMap it to a list of NetworkInterfaceIds.
  // Because we are going to make a call for all of these, it's easier to have
  // a string[] than a string[][] at the end of this
  ?.flatMap((endpoint) => endpoint.NetworkInterfaceIds)
  // Only keep things that are truthy -- in some cases, `NetworkInterfaceIds` may
  // be undefined or something may happen that something else gets in there. But
  // prior to this filter, TSC is worried that our type overall is
  // `(string | undefined)[]`; so we want to filter out any undefined items.
  // Typically, we could just do `filter((eni) => eni)` which would only keep
  // truthy things but TypeScript isn't "smart" enough to understand that everything
  // left would be a string. So we have to instead specifically use a TypeScript
  // User-Defined Type Guard type for the function
  // (https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards)
  // This also means that we have to use `!!eni` instead of just `eni` because we
  // have to return a boolean to appease tsc (where as eni isn't a boolean, it's
  // string | undefined).
  .filter((eni): eni is string => !!eni);
```

kylelaker-ccpo:
> what's the mechanism for actually updating the target group of the ELB

This is the custom resource. So if I understand things correctly, the IPs shouldn't change unless the ENI does which should change for the life of the VPC Endpoint. If those assumptions are wrong then we have a few things to fix here.

But for right now, if the VPCE is replaced, it's ID/ARN changes. That will result in a
"update" operation on the custom resource with the new ID. Because we return the VPCE ID as the physical ID of the custom resource, CloudFormation will then consider it to have been "replaced". That will guarantee that the output will be re-evaluated. So then the Load Balancer Target Group resource created within the `VpcEndpointApplicationTargetGroup` or `VpcEndpointNetworkTargetGroup` will be updated with the new `Targets` block.

> returning a modified object in response to an event actually changes the underlying resource. True?

So in summary, yes.

lib/custom-resources/endpoint-ips.ts
jeffsegal-ccpo: That's not actually my question but I appreciate the answer as I continue to learn at least one new thing about TypeScript in your PRs. I was referring to the `eni is string` portion instead.

lib/atat-web-api-stack.ts
```diff
@@ -62,11 +72,67 @@ export class AtatWebApiStack extends cdk.Stack {
...
+        sslPolicy: elbv2.SslPolicy.FORWARD_SECRECY_TLS12_RES_GCM,
+        certificates: [certificate],
+        defaultTargetGroups: [
+          new VpcEndpointApplicationTargetGroup(this, "VpcEndpointTarget", {
+            vpcEndpointId: network.endpoints.apigateway.vpcEndpointId,
+            port: 443,
+            vpc: network.vpc,
+            protocol: elbv2.ApplicationProtocol.HTTPS,
+            healthCheck: {
+              healthyHttpCodes: "200-299,400-499",
```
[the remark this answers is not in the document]
kylelaker-ccpo: I didn't do that but I added another comment explaining why with a different TODO.

lib/atat-web-api-stack.ts
```diff
@@ -62,11 +72,51 @@ export class AtatWebApiStack extends cdk.Stack {
...
+              // on the load balancer because we would need to send some kind of header in
+              // order to successfully hit our API via the endpoint. We would hve to send
+              // either the Host or x-apigw-api-id headers and we can only specify paths.
+              // If we at least get a response and that response is either a 200 or it's a
+              // 403 (which is what API Gateway will return when we've failed to provide the
+              // header), then we should be all good.
+              // TODO: Perhaps mitigate this by closely monitoring the number of 4xx or 5xx
+              // returned from the ALB or if the ALB receives a large number of requests that
+              // never make it to the API Gateway? Other solutions may be available.
+              healthyHttpCodes: "200,403",
```
jeffsegal-ccpo, line 107: This seems like the best compromise for now.

jeffsegal-ccpo (approved):

lib/custom-resources/endpoint-ips.ts
kylelaker-ccpo: Oh yeah, So `eni is string` is a "return type" for a type predicate. It says, "this function returns a boolean and if true, this is a string; if false, it's something else". https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

Which is also why we need the `!!eni` coercion -- to force this to a `boolean`. Because TypeScript gets mad if your type predicate relies on truthiness instead of just being `true`/`false`.

juliusfitzhugh-ccpo (approved):
Nice work with adding support for custom domains! Once the `.secrets.baseline` is fixed, I think it should be good to go.

kylelaker-ccpo:
> I think it should be good to go

I might keep that broken for a bit until we fix SNOW to be able to pass the new header(s).

PR #1049, opened 2022-07-15
Mock a cost response

juliusfitzhugh-ccpo:
The plan is to revert.

Ticket: AT-7457


kylelaker-ccpo (approved):

kylelaker-ccpo:
> Not a long term solution and will be reverted

What's the blocker to reverting? I assume having the upstream mock provider service?

juliusfitzhugh-ccpo:
> What's the blocker to reverting? I assume having the external mock provider service?

@kylelaker-ccpo yeah, the external mock provider service was being worked on, so this is a backup for demoing.

zachclark-ccpo (approved):

PR #1050, opened 2022-07-18
Update dependencies

kylelaker-ccpo:
This performs a pretty typical update of our dependencies. Additionally, with AWS CDK v2.32.0, we are now able store objects in Secrets Manager as JSON using the new API. This allows us to store a much nicer structure in the Secret.

Note that due to the fact this changes the physical name of the resource, all existing Secrets Manager resources will be deleted during the next deployment; however, the underlying Access Keys will not be changed.


lib/constructs/api-user.ts
```diff
@@ -30,21 +31,16 @@ export class ApiUser extends Construct {
...
-      secretStringValue: accessKey.secretAccessKey,
+      secretName: `${props.secretPrefix}/${user.userName}`,
+      secretObjectValue: {
+        // This usage of 'unsafePlainText' is safe for two reasons. First, the underlying
+        // value is not truly a secret. Second, the Access Key ID will not actually appear
+        // within the template regardless (it will be a CloudFormation Fn::GetAtt
+        // intrinsic function).
+        accessKeyId: cdk.SecretValue.unsafePlainText(accessKey.accessKeyId),
+        secretAccessKey: accessKey.secretAccessKey,
+      },
```
juliusfitzhugh-ccpo, line 43: Nice! This definitely gives us more flexibility, thanks for implementing.

juliusfitzhugh-ccpo (approved):
@kylelaker-ccpo approving, but had a small question. Does something special need to happen with the `@aws-cdk/aws-*-alpha`? I noticed there was a peer dependency issue when trying to `npm i` before you mentioned you'd open a PR.

kylelaker-ccpo:
> Does something special need to happen with the `@aws-cdk/aws-*-alpha`?

If we're using `ncu`, then we need to basically `--reject "@aws-cdk/aws-*-alpha"`. The `cdk-monitoring-constructs` library that we're using (unfortunately) specifies a peer dependency on the `2.18` version (and `ncu` will helpfully bump that to `2.32`). It's kina annoying and I am wondering if `cdk-monitoring-constructs` is really worth it.

juliusfitzhugh-ccpo:
☝️ thanks for the explanation. It does put a bit of a bump in updating dependencies. We can give `cdk-monitoring-constructs` a bit more time and see how it goes.

PR #1059, opened 2022-07-20
Update HOTH specification with IGCE schema

juliusfitzhugh-ccpo:
Add the IGCE schema to support document generation of the IGCE document.

Summary of the information that needs to be captured is below. See [ticket comment](https://ccpo.atlassian.net/browse/AT-7757?focusedCommentId=337288) for a more in depth explanation of the IGCE document and an example with filled in data. 
### Period Sheets 
There are multiple period sheets that contain information for each period. The information required: 
- a `period of performance` - an amount of time (e.g., 6 months)
- the `funding document number` - relates to the funding plan
  - is either the GTC and order number when using G-Invoicing and the 7600 A/B documents, OR the MIPR number when using a MIPR document. 
- one or more `CLIN line items` (various properties)

### Summary Sheet
Most of the summary sheet will be filled in based upon formulas connected to the period sheets. However, there are two fields that need to be filled in: 
- the `surge capabilities` (e.g., 1-50%)
- the `funding document number` - same as the period sheet

### Supporting Documentation
On the `Instructions` sheet towards the bottom it refers to supporting documentation used for estimates that can come from CSP calculators. It is not clear which format the supporting document will be in currently (e.g., PDF, CSV, link), and that this may help else where causing the removal of the `Supporting Information for IGCE` sheet in the IGCE document.

Ticket: AT-7757

### DIFF
--- a/hoth_api.yaml
+++ b/hoth_api.yaml
@@ -465,7 +465,7 @@ components:
           - "3000_OTHER"
         dow_task_number:
           type: string
-        service:
+        service_offering:
           description: "The service offering for the particular line item."  
           type: string 
         item_description_or_config_summary:
--- a/hoth_api.yaml
+++ b/hoth_api.yaml
@@ -407,10 +407,77 @@ components:
           type: string
           enum:
             - DESCRIPTION_OF_WORK
+            - INDEPENDENT_GOVERNMENT_COST_ESTIMATE
         templatePayload:
           type: object
           oneOf:
             - $ref: "#/components/schemas/DescriptionOfWork"
+            - $ref: "#/components/schemas/IndependentGovernmentCostEstimate"
+    IndependentGovernmentCostEstimate:
+      description: "Data used for generating the Independence Government Cost Estimate (IGCE) document"
+      type: object
+      properties:
+        funding_document:
+          description: >-
+            Information related to a funding plan. Either both the GTC number and Order 
+            number or only the MIPR number. 
+          type: object
+          properties:
+            funding_type:
+              type: string
+              enum: 
+                - FS_FORM
+                - MIPR
+            gtc_number: 
+              type: string
+            order_number:
+              type: string
+            mipr_number: 
+              type: string
+        surge_capabilities:
+          type: string
+        periods_estimate:
+          type: array
+          items:
+            $ref: "#/components/schemas/PeriodEstimate"
+    PeriodEstimate:
+      description: "Information specific to a base or option period sheet in the IGCE document"
+      type: object
+      properties:
+        period:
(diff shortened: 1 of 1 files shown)


hoth_api.yaml
```diff
@@ -407,10 +407,77 @@ components:
...
+          type: array
+          items:
+            $ref: "#/components/schemas/PeriodEstimate"
+    PeriodEstimate:
+      description: "Information specific to a base or option period sheet in the IGCE document"
+      type: object
+      properties:
+        period:
+          description: "The period of performance for a specific period sheet."
+          $ref: "#/components/schemas/Period"
```
zachclark-ccpo, line 449: We should use:
```
    PeriodOfPerformance:
      description: "Part of DOW, represents a single Period of Performance"
      type: object
      properties:
        base_period:
          $ref: "#/components/schemas/Period"
        option_periods:
          type: array
          items:
            $ref: "#/components/schemas/Period"
        pop_start_request:
          type: boolean
        requested_pop_start_date:
          type: string
          format: date
          example: "2021-07-01"
        time_frame:
          type: string
          enum:
            - NO_LATER_THAN
            - NO_SOONER_THAN
        recurring_requirement:
          type: boolean
```

zachclark-ccpo: Specifically for 
```
 base_period:
          $ref: "#/components/schemas/Period"
        option_periods:
          type: array
          items:
            $ref: "#/components/schemas/Period"
```

zachclark-ccpo: My concern is ordering the Periods on the backend

hoth_api.yaml
```diff
@@ -407,10 +407,77 @@ components:
...
+        idiq_clin:
+          type: string
+          enum: 
+          - "1000_CLOUD"
+          - "2000_CLOUD_SUPPORT"
+          - "3000_OTHER"
+        dow_task_number:
+          type: string
+        service:
+          description: "The service offering for the particular line item."
```
zachclark-ccpo, line 469:
```suggestion
        service_offering:
          description: "The service offering for the particular line item."  
```

zachclark-ccpo (changes requested):
Small change requested 
```
service -> service_offering
```

hoth_api.yaml
juliusfitzhugh-ccpo: We would be able to identify the base period using `period_type` and then for all the option periods we can use the `option_order` to order on the backend. I'd be open to using the PoP schema, but I'd be concerned the extra data might cause validation issues.

zachclark-ccpo (approved):

PR #1074, opened 2022-07-27
Update dependencies

kylelaker-ccpo:
This updates many dependencies (including, as @juliusfitzhugh-ccpo will likely be excited about, `puppeteer`).


juliusfitzhugh-ccpo (approved):
Yay, puppeteer is finally being upgrade! 😎  A few packages are behind slightly behind, but I will capture them in the next batch shortly after this is merged.

PR #1075, opened 2022-07-28
Resolve or supress all 800-53r4 checks in ATAT stack

kylelaker-ccpo:
We still need to do some work to solve IAM and Network; however, this
fixes the issues in the main app stack. All remaining issues are
resolved or are suppressed and the check is added to the deployment
pipeline to ensure that we always remain compliant. This is far easier
to do now that we have networking solved (except that tricky
`AwsCustomResource`).

This also adds a synthesis of a full environment to our CI workflow.

### DIFF
--- a/lib/atat-web-api-stack.ts
+++ b/lib/atat-web-api-stack.ts
@@ -4,6 +4,7 @@ import * as acm from "aws-cdk-lib/aws-certificatemanager";
 import * as ec2 from "aws-cdk-lib/aws-ec2";
 import * as elbv2 from "aws-cdk-lib/aws-elasticloadbalancingv2";
 import * as iam from "aws-cdk-lib/aws-iam";
+import * as kms from "aws-cdk-lib/aws-kms";
 import * as lambda from "aws-cdk-lib/aws-lambda";
 import * as nodejs from "aws-cdk-lib/aws-lambda-nodejs";
 import { Construct } from "constructs";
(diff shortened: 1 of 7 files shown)


lib/atat-web-api-stack.ts
```diff
@@ -59,14 +61,19 @@ export class AtatWebApiStack extends cdk.Stack {
 
   constructor(scope: Construct, id: string, props: AtatWebApiStackProps) {
     super(scope, id, props);
+
     const { environmentName, network } = props;
+    const sharedAtatKmsKey = new kms.Key(this, "AtatKey", {
+      enableKeyRotation: true,
+    });
```
juliusfitzhugh-ccpo, line 68: I know we mentioned rotating keys using lambda, is this the same or different from that?

lib/atat-web-api-stack.ts
```diff
@@ -81,7 +88,16 @@ export class AtatWebApiStack extends cdk.Stack {
       const loadBalancer = new elbv2.ApplicationLoadBalancer(this, "LoadBalancer", {
         vpc: network.vpc,
         internetFacing: false,
+        deletionProtection: true,
```
juliusfitzhugh-ccpo, line 91: Nice add!

juliusfitzhugh-ccpo (approved):
Looks good! Thanks for providing reasons for the suppressions, it helps give context of other things that are in play outside of AWS.

lib/atat-web-api-stack.ts
kylelaker-ccpo: This is different -- I could try to explain it but honestly I think the graphics at https://docs.aws.amazon.com/kms/latest/developerguide/rotate-keys.html explain it a little better than I can. This is purely for the KMS Key, not for any sort of larger definition of "key"

lib/atat-web-api-stack.ts
kylelaker-ccpo: Yeah, it'll be nice to make sure we don't change the DNS name unexpectedly. This is actually one of the `cdk-nag` rules.

jeffsegal-ccpo (approved):

PR #1088, opened 2022-08-01
Add X-Ray Tracing Config for all Lambda functions

kylelaker-ccpo:
This feature is now supported in AWS GovCloud (US) so we can remove our
Aspect config for this and add the TracingConfig to all of our
functions!

See https://aws.amazon.com/about-aws/whats-new/2022/07/aws-lambda-supports-aws-x-ray-govcloud-us-regions/

Additional commits will add the AWS Lambda PowerTools features to all
our functions and patch our usage of the AWS SDK.

### DIFF
--- a/api/cost/cost-request-fn.ts
+++ b/api/cost/cost-request-fn.ts
@@ -1,15 +1,17 @@
-import middy from "@middy/core";
-import { sqsClient } from "../../utils/aws-sdk/sqs";
-import { SQSEvent, SQSRecord } from "aws-lambda";
-import { logger } from "../../utils/logging";
-import { CostRequest, CspRequest } from "../../models/cost-jobs";
-import { SendMessageCommand } from "@aws-sdk/client-sqs";
 import { injectLambdaContext } from "@aws-lambda-powertools/logger";
-import inputOutputLogger from "@middy/input-output-logger";
+import { captureLambdaHandler } from "@aws-lambda-powertools/tracer";
+import { SendMessageCommand } from "@aws-sdk/client-sqs";
+import middy from "@middy/core";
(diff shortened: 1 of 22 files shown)


api/cost/cost-request-fn.ts
```diff
@@ -1,15 +1,17 @@
+import { injectLambdaContext } from "@aws-lambda-powertools/logger";
+import { captureLambdaHandler } from "@aws-lambda-powertools/tracer";
+import { SendMessageCommand } from "@aws-sdk/client-sqs";
 import middy from "@middy/core";
-import { sqsClient } from "../../utils/aws-sdk/sqs";
+import errorLogger from "@middy/error-logger";
+import inputOutputLogger from "@middy/input-output-logger";
 import { SQSEvent, SQSRecord } from "aws-lambda";
-import { logger } from "../../utils/logging";
+import JSONErrorHandlerMiddleware from "middy-middleware-json-error-handler";
```
juliusfitzhugh-ccpo, line 8: Did you reorder the imports, and if so is there a way you normally organize?

juliusfitzhugh-ccpo (approved):
Thanks for adding in X-Ray. I didn't see any functions missing `tracer`, so should be good. I'm excited to be getting traces finally (and learning how to read/parse them better).

[Tracer documentation](https://awslabs.github.io/aws-lambda-powertools-typescript/latest/core/tracer/#lambda-handler) for those interested.

api/cost/cost-request-fn.ts
kylelaker-ccpo: I did reorder the imports. I used the default ordering from the TypeScript Language Server, `tsserver`, (so what would happen if you used the VS Code command)

PR #1090, opened 2022-08-02
Fix allowed scopes for Cost Request Function

kylelaker-ccpo:
Per the API spec, this is only permitted to read cost.

### DIFF
--- a/lib/constructs/cost-api-implementation.ts
+++ b/lib/constructs/cost-api-implementation.ts
@@ -79,9 +79,7 @@ export class CostApiImplementation extends Construct implements ICostApiImplemen
 
     // secrets permissions
     cspConfig.grantRead(this.costRequestFn);
-    props.idp?.addClient(new idp.IdentityProviderLambdaClient("CspWriteCost", this.costRequestFn), [
-      "atat/write-portfolio",
-    ]);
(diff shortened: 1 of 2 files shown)


lib/constructs/cost-api-implementation.ts
```diff
@@ -79,9 +79,7 @@ export class CostApiImplementation extends Construct implements ICostApiImplemen
 
     // secrets permissions
     cspConfig.grantRead(this.costRequestFn);
-    props.idp?.addClient(new idp.IdentityProviderLambdaClient("CspWriteCost", this.costRequestFn), [
-      "atat/write-portfolio",
-    ]);
+    props.idp?.addClient(new idp.IdentityProviderLambdaClient("CspWriteCost", this.costRequestFn), ["atat/read-cost"]);
```
juliusfitzhugh-ccpo, line 82: Should `CspWriteCost` also be changed?
```suggestion
    props.idp?.addClient(new idp.IdentityProviderLambdaClient("CspReadCost", this.costRequestFn), ["atat/read-cost"]);
```

juliusfitzhugh-ccpo (approved):
Thanks for the fix! One comment, but should be good to go.

lib/constructs/cost-api-implementation.ts
kylelaker-ccpo: Hmm curious to see what this does in CloudFormation -- let's roll with it!

PR #1092, opened 2022-08-04
Update dependencies

juliusfitzhugh-ccpo:


kylelaker-ccpo:
Oof not sure what's up with that build failure

juliusfitzhugh-ccpo:
@kylelaker-ccpo okay, so it's not just me I was trying to wrap my mind around the issue, no luck either.

kylelaker-ccpo:
It looks like maybe `npm update` bumped the version of `yargs-parser`? Maybe that's (part of) the issue? Not really sure. But it definitely seems related to the error message.

kylelaker-ccpo:
Ahh yeah https://github.com/yargs/yargs-parser/issues/452 seems related

juliusfitzhugh-ccpo:
yeah, that definitely is the same error we're getting. Do we just wait for that issue to be fixed? Or is there something else we can do?

kylelaker-ccpo:
How did I end up with merge conflicts when I did that? Okay... uhh lemme work that out.

juliusfitzhugh-ccpo:
Thanks for the additions! Looks like the issue was fixed for `yargs-parser`. I would approve, but not allowed 😆.
@jeffsaunders-hacc @zachclark-ccpo can one of you approve this PR?

kylelaker-ccpo (approved):
I'm good with just approving. It's dependencies only and if we're both good with the changes, I think that's sufficient.

PR #1098, opened 2022-08-09
Update provisioning schema and HOTH spec

juliusfitzhugh-ccpo:
Update provisioning schema and HOTH specification to account for the changes in operators 
used for provisioning jobs, which aligns with the ATAT spec  version `0.3.0`. 
- Update provisioning job model and HOTH spec
- Update tests to reflect operator changes

Ticket: AT-7804

### DIFF
--- a/hoth_api.yaml
+++ b/hoth_api.yaml
@@ -272,11 +272,10 @@ components:
           items:
             $ref: "#/components/schemas/FundingSource"
         operators:
+          description: Array of operator objects
           type: array
-          description: Array of operator email addresses
           items:
-            type: string
-            example: "user@mail.mil"
+            $ref: "#/components/schemas/Operator"
         csp_id:
           type: string
           readOnly: true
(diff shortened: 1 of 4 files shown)


hoth_api.yaml
```diff
@@ -299,11 +298,21 @@ components:
...
+      description: Operator data for provisioning resources in a CSP
+      type: object
+      properties:
+        email:
+          type: string
+          example: "user@mail.mil"
+        dod_id:
+          type: string
+        needs_reset:
+          type: boolean
```
kylelaker-ccpo, line 315: Are none of these required?

kylelaker-ccpo (commented):
Great stuff! Just have a question to make sure we didn't missed a `required` field

hoth_api.yaml
jasonburkert-ccpo: In the database schema I marked email and dod_id as mandatory, so perhaps do the same here

juliusfitzhugh-ccpo: oh yeah, they are I'll fix that. Thanks! 👍

juliusfitzhugh-ccpo: @jasonburkert-ccpo I made `needsReset` required too since version `0.3.0` of ATAT schema has it with a default of `false`.

juliusfitzhugh-ccpo: Added in https://github.com/dod-ccpo/atat-web-api/pull/1098/commits/778c44a2da8872ab06991c8e176ab4112c552b96

kylelaker-ccpo (approved):

PR #1105, opened 2022-08-12
Implement IGCE document generatin fn

juliusfitzhugh-ccpo:
Summary of changes
- Add schema and models for IGCE
- Update naming to use camelCase for object properties in schema
- Add util `convertPeriodToMonths` fn
- Update document generation handler
- Impl `igce-document` fn

#### Period sheets
- populate line items on each period sheet (top half)
- group line items based on IDIQ CLIN on period sheet (lower half)
  - offset IDIQ CLIN because of how the summary sheet calculations reference costs
  - duplicated rows if needed
#### Summary sheet
- populate `TO` and `CLIN` columns on the summary sheet
- update formulas that get costs from period sheets
  - `J` column formula update
  - formulas update if period extends beyond initial row 41
  - duplicate rows if more needed to account for TO and CLIN
- update formulas used to calculate totals if additional rows added to the summary sheet 
- update formulas to calculate grand total 

#### Validation of document schema
- using the `oneOf` property for validation was not working originally and would fail or allow either documents through
  - adding the `additionalProperties: false` to the document schema object helped get things to work
#### IGCE document generation
- updating formulas with `exceljs` requires starting from the sheet object
  - it appears that updating formulas has to be done from the sheet level
  - found inspiration https://github.com/exceljs/exceljs/issues/230#issuecomment-314849847
```typescript
// link this
periodSheet.getCell("C21").value = { formula: "=Sum(A1:B5)" }

// rather than 
const selectedRows = periodSheet.getRows(3, 5)
selectedRows.forEach((row, rowNumber) => {
  row.getCell("C").value = { formula: "=Sum(A1:B5)" }
}
```
- `.eachRow` vs `.forEach`
  - while there are methods like `eachRow` and `eachCell` available they iterate over most or all rows/cells rather than a selection so `getRows` was mainly used to select specific sections and iterate over them with `forEach`
  - updating a value rather than an array of values
    - probably could have been done better by using values, but based on how the values for the rows or columns are aligned on the sheets and having to guess which cells are empty or trying to find the value index that aligns with a particular value in order to construct an array of values to update on a given row
    - it seemed better selecting specific sections of rows and going for specific cells, then updating their values and iterating over multiple columns to modify specific cell values in the row
- [Shared Formulas](https://github.com/exceljs/exceljs#shared-formula) error caused by original formatting of the IGCE document
  - removed the formulas causing the issue and reapplied formulas
  - found inspiration https://github.com/exceljs/exceljs/issues/370#issuecomment-773162088

Ticket: AT-7759


document-generation/generate-document.ts
```diff
@@ -6,64 +6,68 @@ import inputOutputLogger from "@middy/input-output-logger";
...
+  logger.info("Generating document", { documentType });
+  const documentTemplatePaths: TemplatePaths = {
+    [DocumentType.DESCRIPTION_OF_WORK]: {
+      html: "/opt/dow-template.html",
+      css: "/opt/dow-style.css",
+    },
+    [DocumentType.INDEPENDENT_GOVERNMENT_COST_ESTIMATE]: {
+      excel: "/opt/igce-template.xlsx",
+    },
+  };
```
kylelaker-ccpo, line 43: Oh this would potentially be a really bad situation. Lambda runtimes can be reused and `/opt/` won't be emptied between. I don't _think_ we'd see concurrent executions in the same runtime. But if we did, that'd be even more catastrophic. And I _really_ don't want to end up in situations where we rewrite to the same file(s). Honestly, I'd guess we were making this mistake before but it's far more obvious now.

So I think that what we'd probably want to use is invoke [`fs.mkdtemp`](https://nodejs.org/docs/latest-v16.x/api/fs.html#fsmkdtempsyncprefix-options). Within that directory, we can use static filenames. But we've gotta use a different temp directory each time. And while I am being pedantic about this, we should use [`path.join`](https://nodejs.org/docs/latest-v16.x/api/path.html#pathjoinpaths) too.

```suggestion
  const baseDirectory = fs.mkdtemp("doc-gen");
  const documentTemplatePaths: TemplatePaths = {
    [DocumentType.DESCRIPTION_OF_WORK]: {
      html: path.join(baseDirectory, "dow", "template.html"),
      css: path.join(baseDirectory, "dow", "style.css"),
    },
    [DocumentType.INDEPENDENT_GOVERNMENT_COST_ESTIMATE]: {
      excel: path.join(baseDirectory, "igce", "template.xlsx"),
    },
  };
```

document-generation/generate-document.ts
```diff
@@ -6,64 +6,68 @@ import inputOutputLogger from "@middy/input-output-logger";
...
-    [DocumentType.DESCRIPTION_OF_WORK as string]: {
-      html: "/opt/dow-template.html",
-      css: "/opt/dow-style.css",
-    },
-  };
+  if (documentType === DocumentType.DESCRIPTION_OF_WORK) {
+    // get files to generate documents
+    html = fs.readFileSync(documentTemplatePaths[documentType].html, "utf-8");
+    css = fs.readFileSync(documentTemplatePaths[documentType].css, "utf-8");
+  }
```
kylelaker-ccpo, line 69: This is really unfortunate for a situation that (currently) can never be false. I'd almost rather break this out into it's own function; and perhaps with a `default` that throws an error:

```ts
let html = "";
let css = "";
switch (documentType) {
  case DocumentType.DESCRIPTION_OF_WORK:
    html = fs.readFileSync(documentTemplatePaths[documentType].html, "utf-8");
    css = fs.readFileSync(documentTemplatePaths[documentType].css, "utf-8");
  default:
    throw new Error(`Unsupported PDF generation type: "${documentType}"`
}
```

document-generation/generate-document.ts
```diff
@@ -80,11 +84,21 @@ async function generatePdf(event: RequestEvent<GenerateDocumentRequest>): Promis
...
+async function generateXlsx(
+  event: RequestEvent<GenerateDocumentRequest>,
+  documentTemplatePaths: TemplatePaths
+): Promise<ApiBase64SuccessResponse> {
+  const { documentType, templatePayload } = event.body;
+  let excelTemplatePath = "";
+
+  if (documentType === DocumentType.INDEPENDENT_GOVERNMENT_COST_ESTIMATE) {
+    excelTemplatePath = documentTemplatePaths[documentType].excel;
+  }
```
kylelaker-ccpo, line 100: Same concerns about using a `switch`.

document-generation/igce-document.ts
```diff
@@ -0,0 +1,270 @@
...
+    return INTERNAL_SERVER_ERROR;
+  }
+
+  // sort the base period and option periods
+  const basePeriodLineItems = payload.periodsEstimate.filter((periodEstimate: IPeriodEstimate) => {
+    return periodEstimate.period.periodType === PeriodType.BASE;
+  });
+  const optionPeriodsLineItems = payload.periodsEstimate.filter((periodEstimate: IPeriodEstimate) => {
+    return periodEstimate.period.periodType === PeriodType.OPTION;
+  });
```
kylelaker-ccpo, line 23: We should probably actually sort these by `optionOrder`.

```suggestion
  const basePeriodLineItems = payload.periodsEstimate.filter((periodEstimate: IPeriodEstimate) => {
    return periodEstimate.period.periodType === PeriodType.BASE;
  });
  const optionPeriodsLineItems = payload.periodsEstimate.filter((periodEstimate: IPeriodEstimate) => {
    return periodEstimate.period.periodType === PeriodType.OPTION;
  }).sort((a, b) => a.optionOrder - b.optionOrder);
```

document-generation/igce-document.ts
```diff
@@ -0,0 +1,270 @@
...
+    periodSheet.getCell("C2").value = pop;
+    periodSheet.getCell("C3").value = fundingDocumentNumber;
+
+    // add each line item to the period sheet (top half)
+    const numberOfItems = periodLineItems.length;
+    const lineItemRows = periodSheet.getRows(periodSheetTopHalfInitialStartRow, numberOfItems);
+    lineItemRows?.forEach((row, index) => {
+      if (!uniqueIdiqClins.includes(periodLineItems[index].idiqClin)) {
+        uniqueIdiqClins.push(periodLineItems[index].idiqClin);
+      }
```
kylelaker-ccpo, line 66: This may be a little cleaner? Not sure.
```suggestion
    const uniqueIdiqClins = Array.from(
      new Set(periodLineItems.map((lineItem) => lineItem.idiqClin))
    );
    lineItemRows?.forEach((row, index) => {
```

document-generation/igce-document.ts
```diff
@@ -0,0 +1,270 @@
...
+    periodIdiqClinLines?.forEach((row, index) => {
+      row.getCell("C").value = uniqueIdiqClins[index];
+      allPeriodsIdiqClins.push(uniqueIdiqClins[index]);
+      summarySheetIdiqClinCounter++;
+    });
+  }
+
+  // populate period sheets
+  basePeriodLineItems.forEach(populatePeriodLineItems);
+  optionPeriodsLineItems.forEach(populatePeriodLineItems);
```
kylelaker-ccpo, line 119: I like it

document-generation/igce-document.ts
```diff
@@ -0,0 +1,270 @@
...
+      let updatedFormula;
+      if (index === 0) {
+        updatedFormula = `=SUM(${col}6:${col}${row.number - 3})`;
+      } else if (index === 1) {
+        updatedFormula = `=${col}${row.number - 1}*.0225`;
+      } else if (index === 2) {
+        updatedFormula = `=SUM(${col}${row.number - 2}:${col}${row.number - 1})`;
+      } else {
+        updatedFormula = "";
+      }
```
kylelaker-ccpo, line 243: This would also make a good `switch`

document-generation/generate-document.ts
kylelaker-ccpo: btw, I know the linter will hate this (there's only 1 `case`) but I'd be cool with an `eslint-disable` for this.

document-generation/generate-document.ts
juliusfitzhugh-ccpo: Oh yeah that's an excellent point, those would be extremely bad 😵. It's easy to overlook the reuse of runtimes in the lambda, it didn't even cross my mind.

@kylelaker-ccpo so to be clear on your suggestion make a temp folder, copy the template files to that temp folder, and then have the lambda use a different temp folder for each execution?

document-generation/generate-document.ts
juliusfitzhugh-ccpo: yeah, that's true. I think I was trying to keep Typescript happy otherwise it complains about the `html` or `css` properties.

I'll break it into it's own function.

document-generation/generate-document.ts
kylelaker-ccpo: Ohh wait this is for the _inputs_?! That's my bad

kylelaker-ccpo: No I don't care about the input templates. Sorry. I shouldn't review when I'm tired.

document-generation/igce-document.ts
juliusfitzhugh-ccpo: yeah, I think so.

document-generation/igce-document.ts
juliusfitzhugh-ccpo: I wanted to break `populatePeriodLineItems` out on it's own, but it would require additional arguments and I don't think fits well with `forEach` 😢.

document-generation/generate-document.test.ts
```diff
@@ -34,6 +53,10 @@ describe("Successful generate-document handler", () => {
       text-align: center;
       margin-top: 0;
     }`;
+    const dowHeaders = {
+      "Content-Type": "application/pdf",
+      "Content-Disposition": `attachment; filename=DescriptionOfWork.pdf`,
```
zachclark-ccpo, line 58:
```suggestion
      "Content-Disposition": `attachment; filename=IGCE.xlsx`,
```

document-generation/igce-document.ts
kylelaker-ccpo: If you're doing `===` comparisons against literals or an `enum`, then you probably want a `switch`. If you're doing complex logic based on function calls or whatever, `if` is probably better. And I think in our cases here, we've mostly got literals and enums

document-generation/generate-document.ts
juliusfitzhugh-ccpo: Yeah, the templates are the files being read in by the generating function as a base and then worked on inside of those functions. Only buffer are output and sent as the response, nothing should be saved as an output. Nothing is written to the original template file I believe.

It's all good, I appreciate the feedback. And now I'll be considering the runtime more.

zachclark-ccpo:
I was doing some testing with the example schema provided in the ticket, did we change the schema recently?
The example is in camelCase while the schema still reads with underscores?

zachclark-ccpo:
> I was doing some testing with the example schema provided in the ticket, did we change the schema recently? The example is in camelCase while the schema still reads with underscores?

It looks like the schema for IGCE is still using underscores in `hoth_api.yaml`, function should be updated to use underscores instead of camelCase for the templatePayload

document-generation/utils/sampleTestData.ts
```diff
@@ -0,0 +1,410 @@
...
+          },
+          {
+            clin: "2000",
+            idiqClin: "2000 Cloud SECRET",
+            dowTaskNumber: "4.2.2.2",
+            serviceOffering: "Special custom built app - not SNOW",
+            itemDescriptionOrConfigSummary: "To handle special cases that can not be done in SNOW",
+            monthlyPrice: 799.0,
+            monthsInPeriod: 36,
+          },
```
zachclark-ccpo, line 282: This doesn't match the schema, which uses underscores. If we are switching to camelCase we need to update `hoth_api.yaml`.

zachclark-ccpo (changes requested):
Need to update `hoth_api.yaml` since we are changing from underscores to camelCase.

document-generation/utils/sampleTestData.ts
juliusfitzhugh-ccpo: > I was doing some testing with the example schema provided in the ticket, did we change the schema recently?
The example is in camelCase while the schema still reads with underscores?

I changed the implementation schema in this ticket, but not in the `hoth_api.yaml` since I thought it was okay. Initially, I think most of the implementation was done with `camelCase` even though the spec was in `snake_cake` (slightly loose from the spec).

Very good point though, but it does make sense that they should match up.

document-generation/generate-document.test.ts
juliusfitzhugh-ccpo: After a second look, this is the correct `filename` since that is the header for the DoW. The `filename` for the IGCE doc is in the `igce-document.ts` and is `filename=IndependentGovernmentCostEstimate.xlsx`.

document-generation/utils/sampleTestData.ts
juliusfitzhugh-ccpo: Will address in a separate PR.

zachclark-ccpo (approved):

jeffsegal-ccpo (approved):

PR #1106, opened 2022-08-12
Add required notification alerts for SSO service

kylelaker-ccpo:
This basically copies the setup used for notifying on changes to IAM
users to the SSO Directory service. This is a compliance requirement.


kylelaker-ccpo:
Actually... I think we need to build this in another account...

PR #1137, opened 2022-08-25
Update dependencies

juliusfitzhugh-ccpo:
`jest` had a major version update to [29.0.0](https://github.com/facebook/jest/blob/main/CHANGELOG.md#2900). There were things that seem like they might break things, so I left it out of this decency update to do it separately.


kylelaker-ccpo:
> I left it out of this decency update to do it separately

I think that's a good idea. Lots to go wrong there even with just `peerDependencies` versions.

kylelaker-ccpo (approved):

PR #1160, opened 2022-09-02
Remove version reporting for CDK commands

kylelaker-ccpo:
This makes templates more readable and removes dead code that never
executes in GovCloud anyway (the conditions never evaluate to true).

### DIFF
--- a/cdk.json
+++ b/cdk.json
@@ -1,5 +1,6 @@
 {
   "app": "npx ts-node --prefer-ts-exts bin/atat-web-api.ts",
+  "versionReporting": false,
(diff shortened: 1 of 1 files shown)


juliusfitzhugh-ccpo (approved):
I didn't even realize it was reporting or had extra code. I guess it is only noticeable if you look at the CF templates?

kylelaker-ccpo:
> I guess it is only noticeable if you look at the CF templates?

Pretty much! It's the really long `Condition` and a lot of the `CDKMetadata` stuff.

PR #1177, opened 2022-09-09
Update AWS CDK to v2.41.0 to put AwsCustomResource in VPC

kylelaker-ccpo:
The new version of the CDK adds support for putting `AwsCustomResource`
resources into a VPC. Because of the underlying ways that these are
created, we just need to create one that ends up having the base
configuration shared across all AwsCustomResource constructs created by
us and elsewhere in the tree, so long as this one is created first
(which is why it's at the top of the `Stack`).

This _only_ makes the changes in the `atat-web-api` stack because that's
the only place where we know the VPC is ready (for example, we can't do
this in the VPC stack because the most important custom resource is
created before the VPC is).


lib/atat-web-api-stack.ts
```diff
@@ -44,6 +45,30 @@ export class AtatWebApiStack extends cdk.Stack {
...
+    // also means that if we create this first in the stack, it will be reused any time that
+    // `AwsCustomResource` is used in our constructs or within the CDK's inner constructs across
+    // this stack.
+    // sts:GetCallerIdentity is the most nop API action that one can think of and it's the only
+    // API call that can never return a permission error so long as you've got valid credentials
+    // so it seems like a pretty reasonable choice for this.
+    // Because we don't control all the places where an `AwsCustomResource` is created and it's
+    // not viable for us to reimplement all of them ourselves to avoid CDK-created ones, this is
+    // the best possible workaround.
+    if (props.network) {
```
jeffsegal-ccpo, line 60: Thanks for the explanation

jeffsegal-ccpo (approved):

PR #1252, opened 2022-10-07
Support building permission sets without permissions boundary

kylelaker-ccpo:
Not all AWS accounts will necessarily have a permissions boundary. We
need to support building our permission sets in such an environment. We
can do this by conditionally providing that attribute.

To avoid awkward issues with potential default values etc, we use the
value "!" as the "policy name" that causes us to ignore applying the
permissions set. We use "!" because it is not valid in a policy name and
because passing the empty string is awkward in the CloudFormation
console and CLI.

This also turns `CentralAccountRolePermissionsBoundary` into a required
parameter instead of an optional parameter with a default value; this is
likely the better long-term decision anyway.


zachmcclurg-ccpo (approved):
approved changes made.

PR #1317, opened 2022-11-02
Update dependencies

juliusfitzhugh-ccpo:
## Summary
- Update dependencies
- Excluded:
  - `@aws-cdk/aws-*-alpha`
  - `puppeteer-core` and `@types/puppeteer`
  - `axios`
- Update logger type to `any` for sdk clients to prevent tests from failing

**Note:** Planning to update `puppeteer` and `axios` after a more recent dependences update 
and to incorporate the temp fix for the logger issues discussed below. 

### Logger issue
There appears to be a difference between the logger type for
sdk clients in `@aws-sdk/type` and `@aws-lambda-powertools/logger`
that causes tests using one of our sdk clients to fail.

I was not able to come across others reporting the same/similar issue. 
The clients `@aws-sdk/client-*` were updated from `v3.193.0` to `v3.201.0`
in this PR (not sure if the error started happening here or for some other reason).

For a temporary fix `any` is used until the issue is resolved, and the fix can
be reverted. The logger appears to be working fine with the fix when testing
against a HOTH API endpoint (provisioning-jobs).

See files:
- `node_modules/@aws-sdk/types/dist-types/logger.d.ts`
- `node_modules/@aws-lambda-powertools/logger/lib/Logger.d.ts`


jeffsegal-ccpo (approved):
Thanks for the workaround and hopefully we can get the real fix soon.

PR #1321, opened 2022-11-04
Fix CSP response

juliusfitzhugh-ccpo:
Previously the `$metadata.request` returned the axios `response.request`
object, which is not the request that is sent to the CSP. To fix this
the original request sent to the CSP is passed into the transformation
functions as an argument.


jeffsegal-ccpo (approved):
Thanks, this should be what we need.

PR #1363, opened 2022-11-23
AT-8376: Added first cut of EvalPlan object

jeffsegal-ccpo:
Adds `EvalPlan` schema to support _Evaluation Plan_ document generation

Ticket: AT-8376

### DIFF
--- a/hoth_api.yaml
+++ b/hoth_api.yaml
@@ -394,13 +394,13 @@ components:
       description: "Determines which document template the API will generate"
       type: object
       properties:
-        document_type:
+        documentType:
           type: string
           enum:
             - EVAL_PLAN
             - DESCRIPTION_OF_WORK
             - INDEPENDENT_GOVERNMENT_COST_ESTIMATE
-        template_payload:
+        templatePayload:
           type: object
           oneOf:
             - $ref: "#/components/schemas/EvalPlan"
--- a/hoth_api.yaml
+++ b/hoth_api.yaml
@@ -394,22 +394,23 @@ components:
       description: "Determines which document template the API will generate"
       type: object
       properties:
-        documentType:
+        document_type:
           type: string
           enum:
+            - EVAL_PLAN
             - DESCRIPTION_OF_WORK
             - INDEPENDENT_GOVERNMENT_COST_ESTIMATE
-        templatePayload:
+        template_payload:
           type: object
           oneOf:
             - $ref: "#/components/schemas/EvalPlan"
             - $ref: "#/components/schemas/DescriptionOfWork"
             - $ref: "#/components/schemas/IndependentGovernmentCostEstimate"
-   EvalPlan:
-     description: Provides the basis of evaluation for an Aquisition Package
-     type: object
-     properties:
-        source_selection:
+    EvalPlan:
+      description: Provides the basis of evaluation for an Aquisition Package
+      type: object
+      properties:
+        sourceSelection:
           type: string
           enum:
             - NO_TECH_PROPOSAL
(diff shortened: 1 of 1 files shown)


hoth_api.yaml
```diff
@@ -402,9 +402,44 @@ components:
...
+            type: string
+        standard_differentiators:
+          type: array
+          items:
+            type: string
+        custom_differentiators:
+          type: array
+          items:
+            type: string
+   IndependentGovernmentCostEstimate:
```
jasonburkert-ccpo, line 442: These lines need to shift by a space

Easier to see in this view

image


```suggestion
    EvalPlan:
      description: Provides the basis of evaluation for an Aquisition Package
      type: object
      properties:
         source_selection:
           type: string
           enum:
             - NO_TECH_PROPOSAL
             - TECH_PROPOSAL
             - SET_LUMP_SUM
             - EQUAL_SET_LUMP_SUM
         method:
           type: string
           enum:
             - LPTA
             - BVTO
             - BEST_USE
             - LOWEST_RISK
         standard_specifications:
           type: array
           items:
             type: string
         custom_specifications:
           type: array
           items:
             type: string
         standard_differentiators:
           type: array
           items:
             type: string
         custom_differentiators:
           type: array
           items:
             type: string
    IndependentGovernmentCostEstimate:
```

jasonburkert-ccpo (changes requested):
Compared to Eval Plan document.  Looks reasonable.

hoth_api.yaml
```diff
@@ -402,9 +402,44 @@ components:
...
           oneOf:
+            - $ref: "#/components/schemas/EvalPlan"
             - $ref: "#/components/schemas/DescriptionOfWork"
             - $ref: "#/components/schemas/IndependentGovernmentCostEstimate"
-    IndependentGovernmentCostEstimate:
+   EvalPlan:
+     description: Provides the basis of evaluation for an Aquisition Package
+     type: object
+     properties:
+        source_selection:
```
juliusfitzhugh-ccpo, line 412: More of a general question, but do we want to shift to using `camelCase` in the HOTH spect to match our models in Lambda?
I think I have a branch where I was started to update the HOTH spec to be camelCase for the DoW and IGCE, but never opened up a PR.

jasonburkert-ccpo: There seems to be a mixture currently.  GenerateDocumentRequest and CspResponse have camelCase properties, with snake_case used in many other places.

UI team seems to use snake_case in their data models, and ServiceNow internals use snake_case.  That may explain the tendancy to lean that way.

I prefer consistency.

juliusfitzhugh-ccpo (approved):

hoth_api.yaml
juliusfitzhugh-ccpo: Per the team discussion, the short term is to progress with `camelCase`.

jasonburkert-ccpo (approved):

hoth_api.yaml
jasonburkert-ccpo: Fixed in https://github.com/dod-ccpo/atat-web-api/pull/1363/commits/df26bbff2c3a0cf385d523e02ec5a19948007366

hoth_api.yaml
jasonburkert-ccpo: Agreed, per team discussion.  This API spec may be inconsistent in the short term, but we'll begin to lean toward use of `camelCase` property names.  A significant driver of that decision is due to [Middy](https://middy.js.org/) validator use in AWS lambdas.  In the long term this API spec and surrounding implementation should use `camelCase`.

PR #1377, opened 2022-11-30
Create IFP document .docx template

juliusfitzhugh-ccpo:
### Implement IFP generation in HOTH
- Create IFP document `.docx` template
- Create/update models for IFP and update `mediaType`
- Implement IFP document generation `ifp-document.ts`
- Update main `generate-document.ts` function

Ticket: AT-8377


document-generation/ifp-document.ts
```diff
@@ -0,0 +1,34 @@
...
+        // update amounts to have commas
+        estimatedTaskOrderValue: payload.estimatedTaskOrderValue.toLocaleString("en-US"),
+        initialAmount: payload.initialAmount.toLocaleString("en-US"),
+        remainingAmount: payload.remainingAmount.toLocaleString("en-US"),
+        // construct Funding Document information to be used in template
+        fundingDocInfo: getFundingDocInfo(payload.fundingDocument),
+      },
+      cmdDelimiter: ["{", "}"],
+    })
+  );
```
zachclark-ccpo, line 25: Should there be a `try... catch` in case `createReport` results in an error?

juliusfitzhugh-ccpo: yeah, good point. I know the `errorHandlerMiddleware` and `middy` end up capturing most of the errors, so I sometimes leave out the `try/catch` blocks.

zachclark-ccpo: True, middy will probably catch it and return a 500 back to SNOW. No need to add try/catch in that case.

jasonburkert-ccpo (approved):
Agree with @zachclark-ccpo, clean!

PR #1418, opened 2022-12-15
Add a step for encrypting EBS volumes

kylelaker-ccpo:
This keeps coming up in conversations and so we should just start
enabling it. There's no reason not to. We can just use the AWS-managed
key for now because we don't expect anything to use it _and_ that at
least ensures it's encrypted.


juliusfitzhugh-ccpo (approved):
Sounds reasonable to me.

PR #1419, opened 2022-12-16
Add ELB Access Logs

kylelaker-ccpo:
These logs will be helpful from a troubleshooting perspective as well as
required for compliance and security reasons. Enabling them requires
specifying the region on the stack. This is really unfortunate because
it means that our template isn't 100% portable anymore (as in, an
already-built template can no longer be run in any region; though, we
can still build a template that works in any region). The change is a
bit unfortunate as it requires _all_ synthesis to take place with the
`AWS_PROFILE` environment variable set where it points to the desired
target account/region.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -173,7 +173,7 @@
         "filename": "lib/atat-web-api-stack.ts",
         "hashed_secret": "e2c872ec4dc5d6d2ca8f1f22852b4670d039470a",
         "is_verified": false,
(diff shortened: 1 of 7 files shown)


kylelaker-ccpo:
The code that "isn't tested" here is mostly nag suppressions or just applying the bucket to the load balancer. Those get tested as part of the CI tests.

jeffsegal-ccpo (approved):

PR #1420, opened 2022-12-17
Add Impact Level and Deadline for provisioning request

juliusfitzhugh-ccpo:
- Add properties for `targetImpactLevel` and `provisionDeadline` to provisioning request
- Fix error from using `test` from `@jest/globals` by updating tests `import`
- Update secrets.baseline

### DIFF
--- a/api/provision/csp-write-portfolio.ts
+++ b/api/provision/csp-write-portfolio.ts
@@ -72,6 +72,8 @@ async function makeRequest(client: IAtatClient, request: ProvisionRequest): Prom
         clin: undefined,
       })),
     },
+    targetImpactLevel: request.targetImpactLevel,
+    provisionDeadline: request.provisionDeadline,
   };
   try {
     // TODO: remove once mocking is no longer needed (e.g., mocking api implemented or actual csp integration)
--- a/api/provision/csp-write-portfolio.ts
+++ b/api/provision/csp-write-portfolio.ts
@@ -61,6 +61,8 @@ function transformAsynchronousResponse(
 async function makeRequest(client: IAtatClient, request: ProvisionRequest): Promise<ProvisionCspResponse> {
   // This function will always be operating for creating new portfolios; if we have something
   // else, this will be a non-recoverable error anyway.
+  const now = new Date();
+  const provisionDeadline = now.setDate(now.getDate() + 12); // deadline in 12 days
   const payload = request.payload as NewPortfolioPayload;
   const creationRequest: atatApiTypes.AddPortfolioRequest = {
     portfolio: {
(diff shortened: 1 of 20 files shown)


lib/aspects/govcloud-compatibility.test.ts
```diff
@@ -5,6 +5,7 @@ import * as lambda from "aws-cdk-lib/aws-lambda";
 import * as logs from "aws-cdk-lib/aws-logs";
 import { Annotations, Match, Template } from "aws-cdk-lib/assertions";
 import { GovCloudCompatibilityAspect } from "./govcloud-compatibility";
+import { test } from "@jest/globals";
```
kylelaker-ccpo, line 8: I understand this change is necessary to get things working at the moment. Can it be tracked as tech debt? The jest globals should just work and shouldn't need to be imported directly (notice that `describe` doesn't need this).

juliusfitzhugh-ccpo: @kylelaker-ccpo reverted change. I originally thought the test would break when pushed up, so I added the fix. Will capture in tech debt.

api/provision/csp-write-portfolio.ts
```diff
@@ -72,6 +72,8 @@ async function makeRequest(client: IAtatClient, request: ProvisionRequest): Prom
         clin: undefined,
       })),
     },
+    targetImpactLevel: request.targetImpactLevel,
```
jeffsegal-ccpo, line 75: This won't be coming from the request - we'll need to pull it from Secrets Manager. Right now the schema is:

```
{
  "csp_name": {
    "uri": "https://atat-mock-server.azurewebsites.net"
  }
}
```

We'll probably want to modify that to:

```
{
  "csp_name": {
    "uri": "https://atat-mock-server.azurewebsites.net",
    "network": "IL2"
  }
}
```

And we may also want to document this structure in the HOTH API YAML. I can go ahead and update the records in SM if you're good with this.

jeffsegal-ccpo (changes requested):

api/provision/csp-write-portfolio.ts
juliusfitzhugh-ccpo: @jeffsegal-ccpo okay, yeah I forgot we were going to provide that in the secrets manager. That works.

juliusfitzhugh-ccpo: @jeffsegal-ccpo updated so SNOW doesn't know anything about `network` or `deadline` and updated HOTH.

jeffsegal-ccpo (approved):

PR #1433, opened 2022-12-21
Add a custom resource to remove all rules from the VPC default SG

kylelaker-ccpo:
AWS' Best Practices for the CIS Benchmarks and for Foundational Best Practices recommend not having any rules that allow traffic in the Default Security Group for a VPC. This SG by default allows traffic within the SG and all egress traffic. We **do not** place any resources in the Security Group. Nonetheless, for some reason we are required to comply with the CIS rules 4.3 and 5.3 as well as EC2.2.

CloudFormation (and therefore the CDK) do not provide a native way to delete resources as part of resource definitions. So we have to write a custom resource to handle the deletions. This is a gross abuse of Custom Resources and perhaps a hook triggered on `AWS::EC2::VPC` creation would be better, but with this setup we can directly pass the attributes of the VPC to the custom resource to handle the rule deletion.

This deletes all rules from the passed security group with no regard for the content of the rule. REALLY DEFINITELY make sure you're passing in the default security group and don't get wild and start passing in other security groups. There is no obvious way to detect if a Security Group is the default so the best check would be description-based which is pretty weak. So instead, this just doesn't check.

At the same time some VPC endpoints are added that make this custom resource's job a little easier and they also check some other compliance boxes.

Tests are not written because this is a bunch of complicated API calls and custom resources that are not easily tested. We can improve testing of these resources at a later time.

### DIFF
--- a/lib/constructs/vpc-default-sg-rule-remove.ts
+++ b/lib/constructs/vpc-default-sg-rule-remove.ts
@@ -1,0 +1,80 @@
+import * as cdk from "aws-cdk-lib";
+import * as ec2 from "aws-cdk-lib/aws-ec2";
+import * as iam from "aws-cdk-lib/aws-iam";
+import * as lambda from "aws-cdk-lib/aws-lambda";
+import * as nodejs from "aws-cdk-lib/aws-lambda-nodejs";
+import * as cr from "aws-cdk-lib/custom-resources";
+import { Construct } from "constructs";
+
+export interface VpcDefaultSecurityGroupRuleRemoverProps {
+  /**
+   * The VPC that should have is default SG rules removed, as well as where
+   * the backing custom resource should be deployed.
+   */
+  // This has to accept the concrete class because we need the VPC and
+  // attributes only available on the concrete class.
+  readonly vpc: ec2.Vpc;
+}
+
+/**
+ * A Construct for removing all rules from a given VPC's default Security Group.
+ *
+ * This may be necessary in some environments to meet compliance requirements. All the
+ * rules on the VPC's default Security Group will be deleted (Ingress and Egress). Therefore,
+ * any ENI in this Security Group will not be able to send or receive traffic unless granted
+ * by another Security Group.
+ *
+ * Constructs that add rules to the VPC's default Security Group should depend on this if
+ * those rules should be kept; otherwise, this should depend on any constructs that create
+ * rules on the default Security Group that should be deleted.
+ *
+ * This is implemented as a custom resource in a VPC. It will need network access to make
+ * EC2 and S3 API calls, as well as possibly CloudFormation.
+ */
+export class VpcDefaultSecurityGroupRuleRemover extends Construct {
+  constructor(scope: Construct, id: string, props: VpcDefaultSecurityGroupRuleRemoverProps) {
+    super(scope, id);
+
+    const handler = new nodejs.NodejsFunction(this, "VpcSgRuleRemovalHandler", {
+      runtime: lambda.Runtime.NODEJS_16_X,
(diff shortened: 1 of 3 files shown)


zachmcclurg-ccpo:
good to go

lib/custom-resources/delete-default-sg-rules.ts
```diff
@@ -0,0 +1,86 @@
+import { EC2, SecurityGroupRule } from "@aws-sdk/client-ec2";
+import type { OnEventRequest, OnEventResponse } from "aws-cdk-lib/custom-resources/lib/provider-framework/types";
+
+const ec2 = new EC2({ useFipsEndpoint: true });
+
+/**
+ * Handle CloudFormation Custom Resource events using the CDK Custom Resource Provider Framework.
+ *
+ * This will, given a VPC Endpoint ID, determine the full list of IP addresses associated with
```
jeffsegal-ccpo, line 9: Copypasta?

lib/constructs/vpc-default-sg-rule-remove.ts
```diff
@@ -0,0 +1,57 @@
...
+        }),
+        new iam.PolicyStatement({
+          effect: iam.Effect.ALLOW,
+          actions: ["ec2:RevokeSecurityGroupIngress", "ec2:RevokeSecurityGroupEgress"],
+          // Only allow changing rules on the default security group
+          resources: [
+            cdk.Stack.of(this).formatArn({
+              service: "ec2",
+              resource: "security-group",
+              resourceName: props.vpc.vpcDefaultSecurityGroup,
```
jeffsegal-ccpo, line 38: Just making sure since you mentioned the criticality of passing in the right SG - are you saying that we need to make sure that this line never changes to some other SG? Also, I don't see that property in  https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_ec2.VpcProps.html but maybe that's not the right place to be looking?

jeffsegal-ccpo (changes requested):

lib/constructs/vpc-default-sg-rule-remove.ts
kylelaker-ccpo: Yeah so `Prop` objects are usually inputs. The type is defined on https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_ec2.Vpc.html#vpcdefaultsecuritygroup but notably not on `ec2.IVpc` (which is why the `Props` for this construct takes `ec2.Vpc` instead of `ec2.IVpc`).

jeffsegal-ccpo: OK got it, thanks. Not that they have any actual description in the docs but at least it confirms the field name.

kylelaker-ccpo: Yeah... https://github.com/aws/aws-cdk/blob/1485f48c515f81fa585a325e02f1602f56443605/packages/%40aws-cdk/aws-ec2/lib/vpc.ts#L1390 is where the implementation itself is, matching https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ec2-vpc.html#DefaultSecurityGroup-fn::getatt

lib/custom-resources/delete-default-sg-rules.ts
kylelaker-ccpo: Fixed in bab938fd76512a81384c1c2a49727e8cc2e75a35 along with other docs being added.

jeffsegal-ccpo (approved):

PR #1441, opened 2022-12-22
Add a step for encrypting EBS volumes (#1418)

zachmcclurg-ccpo:
This keeps coming up in conversations and so we should just start enabling it. There's no reason not to. We can just use the AWS-managed key for now because we don't expect anything to use it _and_ that at least ensures it's encrypted.


kylelaker-ccpo:
I think this PR was created in the wrong direction? I am going to close this.

PR #1450, opened 2022-12-27
Remove GT&C number from IFP document

juliusfitzhugh-ccpo:
Remove GTC number. Only the Order number is needed for IFP document.


jasonburkert-ccpo (commented):
Looks good

jasonburkert-ccpo (approved):

PR #1473, opened 2023-01-13
Update optionOrder page calculation to reflect optionOrder changes

zachclark-ccpo:
`optionOrder` used to start on 1 for Option Period 1, but due to `optionOrder` starting on 1 for Base Period instead of null, Option Period 1 had an `optionOrder` of 2. 

This fixes that issue.

Test it on the dev PDI with `ACQ0000230`, you will see that Option Period 1 is an empty page, and all option periods are shifted over:
image

Test it on the At838423WebApi sandbox endpoint and you will see Option Period 1 is in its proper location:
image


juliusfitzhugh-ccpo (approved):
That works!

PR #1479, opened 2023-01-19
Fix bugs generating Dow docx template

juliusfitzhugh-ccpo:
- Fixes based on minor bugs when trying to generate dow doc
- Refactor section 5 CDRL table function
- Update dow generation function with updated utils for section 5
  - Upate dow template section 5 table
  - Refactor `getCDRLs` func
  - the duplication in sonar cloud is from this change with using the same structure for `FFP` and `T&M`
- Update dow template to use `instanceName` for training title
- Update `sortInstanceClassificationLevels` and `sortSupportPackagesByLevels` to guard against instances without classification levels. These instances will not be included in the document since it would cause the document generation to fail if included.

Ticket: AT-8380


jasonburkert-ccpo (approved):

jasonburkert-ccpo:
@jeffsegal-ccpo Might need you to override this gate.

PR #1480, opened 2023-01-19
Feature/cloudwatch log group cissp

zachmcclurg-ccpo:
created a log group for c5isr to pull the logs they need.

### DIFF
--- a/lib/atat-web-api-stack.ts
+++ b/lib/atat-web-api-stack.ts
@@ -197,7 +197,7 @@ export class AtatWebApiStack extends cdk.Stack {
     });
 
     // Cloudwatch Logs for C5ISR
-    const logGroup = new logs.LogGroup(this, "Log Group", {
+    const logGroup = new logs.LogGroup(this, "LogGroup", {
       logGroupName: "hacc-cssp-cwl-logs",
       retention: RetentionDays.INFINITE,
     });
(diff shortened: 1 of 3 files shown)


lib/atat-web-api-stack.ts
```diff
@@ -194,6 +196,18 @@ export class AtatWebApiStack extends cdk.Stack {
       );
     });
 
+    // Cloudwatch Logs for C5ISR
+    const logGroup = new logs.LogGroup(this, "LogGroup", {
+      logGroupName: "prod-cssp-cwl-logs",
+      retention: RetentionDays.INFINITE,
```
zachclark-ccpo, line 202: Question: Do you need to set retention as infinite?
Based on the documentation, it seems that retention is set to infinite automatically: https://docs.aws.amazon.com/managedservices/latest/userguide/log-customize-retention.html

zachmcclurg-ccpo: You are right for console configuration but for cdk it has a 2 year default.  https://docs.aws.amazon.com/cdk/api/v1/docs/@aws-cdk_aws-logs.LogGroup.html and under 'retention?'

zachclark-ccpo (approved):

jeffsegal-ccpo (approved):

PR #1481, opened 2023-01-19
Feature/cloudwatch log group cissp updated

zachmcclurg-ccpo:
this will test changing the secrets by changing the name of the secret stored in SM.

### DIFF
--- a/cdk.json
+++ b/cdk.json
@@ -33,7 +33,7 @@
 
     "atat:CspConfigurationName": "config/csp/integration",
     "atat:GitHubPatName": "auth/github/pat",
-    "atat:VersionControlBranch": "develop",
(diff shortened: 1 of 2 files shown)


cdk.json
```diff
@@ -33,7 +33,7 @@
 
     "atat:CspConfigurationName": "config/csp/integration",
     "atat:GitHubPatName": "auth/github/pat",
-    "atat:VersionControlBranch": "develop",
+    "atat:VersionControlBranch": "feature/cloudwatch-log-group-cissp-updated",
```
juliusfitzhugh-ccpo, line 36: Assuming this is for testing, once testing is done might want to update back to `develop`.

jasonburkert-ccpo: Agree this needs a second look

PR #1488, opened 2023-01-20
Feature/prod secerets write update

zachmcclurg-ccpo:
This PR is to test changing the secrets value by changing the name

### DIFF
--- a/lib/constructs/provisioning-sfn-workflow.ts
+++ b/lib/constructs/provisioning-sfn-workflow.ts
@@ -89,7 +89,7 @@ export class ProvisioningWorkflow extends Construct implements IProvisioningWork
       vpc: props.vpc,
       tracing: lambda.Tracing.ACTIVE,
     });
-    props.idp?.addClient(new IdentityProviderLambdaClient("CspWritePortfolioClient", cspWritePortfolioFn), [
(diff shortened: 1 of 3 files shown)


lib/constructs/provisioning-sfn-workflow.ts
```diff
@@ -89,7 +89,7 @@ export class ProvisioningWorkflow extends Construct implements IProvisioningWork
       vpc: props.vpc,
       tracing: lambda.Tracing.ACTIVE,
     });
-    props.idp?.addClient(new IdentityProviderLambdaClient("CspWritePortfolioClient", cspWritePortfolioFn), [
+    props.idp?.addClient(new IdentityProviderLambdaClient("CspWriteToPortfolioClient", cspWritePortfolioFn), [
```
juliusfitzhugh-ccpo, line 92: I noticed besides the change in the Idp Client name, this appears to have the same changes as https://github.com/dod-ccpo/atat-web-api/pull/1481. Are both PRs needed?

zachmcclurg-ccpo: we can ignore 1481. That's something I can put a hold on for now.

juliusfitzhugh-ccpo (approved):
Looks good. Might need a revert to change back only idp client name if a second secret update happens during that revert.

PR #1489, opened 2023-01-20
Increase SFN Task timeout

juliusfitzhugh-ccpo:
- Update tasks default timeout in the SFN to 6 minutes. This value
was selected since the longest lambda timeout in the SFN is 5
minutes and this ensures that the lambda timeout before the SFN.

- Update api-sfn-function to provide a lambda default of 5 minutes
rather than 5 seconds.


zachclark-ccpo (approved):
Looks good to merge, might need code maintainer review.

jeffsegal-ccpo (approved):

PR #1492, opened 2023-01-20
Update puppeteer to v19.6.0

juliusfitzhugh-ccpo:
Update `puppeteer-core@19.6.0` and `@sparticuz/chromium@110`.
Document generation for PDFs is no longer being used at the moment, 
so we might want to consider if the puppeteer dependencies can be removed.


jasonburkert-ccpo:
Good point raised in the PR description.  Created tech debt ticket AT-8723 for follow-up.

jasonburkert-ccpo (approved):

PR #1497, opened 2023-01-23
commenting out already made log group 

zachmcclurg-ccpo:
commenting out log group so pipeline wont fail due to resource previously being made in the initial deployment.


jasonburkert-ccpo:
@zachmcclurg-ccpo Is this ready for review?

PR #1499, opened 2023-01-23
Add workaround for issues with stack updates

kylelaker-ccpo:
It looks like the CDK may have introduced an issue with stack updates to
the `DescribeCognitoUserPoolClient`. There is a possible unmerged patch
upstream. To unblock updates, this adds a temporary workaround
(reimplemented the custom resource internally) until the fix (or an
alternative) is merged.


zachmcclurg-ccpo (approved):
approved

PR #1512, opened 2023-02-01
update flow log desc

zachmcclurg-ccpo:
updated the vpc flow logs to point from s3 bucket to a cloudwatch log group per C5ISR.


juliusfitzhugh-ccpo (approved):
Looks good.

PR #1524, opened 2023-02-06
AT-8742 Description of Work IL6 Mislabeled

tomarnold-ccpo:
Changed label for IL6 from unclassified to secret


luakim-ccpo (approved):

juliusfitzhugh-ccpo (approved):
Oh... good catch!

PR #1544, opened 2023-02-16
Update igce-template.xlsx to remove B26 Ditco Fee Cell default value

zachclark-ccpo:
Update igce-template.xlsx to remove B26 Ditco Fee Cell default value.
B26 is filled in by the igce-docgen script, no need to set a default value for it.

image


jasonburkert-ccpo (approved):
Looks good.  Understand that label in B26 will be provided dynamically.

image

PR #1552, opened 2023-02-24
Feature/cw event kms access

zachmcclurg-ccpo:
grants the Decrypt and GenerateDataKey capabilities for AWS CloudWatch events to trigger SNS topics. Currently SNS it encrypted with a KMs so the "events.amazonaws.com" needs access to that key. 
https://aws.amazon.com/premiumsupport/knowledge-center/sns-not-getting-eventbridge-notification/


jasonburkert-ccpo (approved):
Thanks for the call @zachmcclurg-ccpo

PR #1561, opened 2023-03-10
Refactor Consume Provisioning Job unit test

jasonburkert-ccpo:
- Improves test descriptions for clarity of intent (non-empty queue/empty queue/misbehaving queue)
- Explicitly states expectations for mock responses, especially wrt to `Messages` key
- Simplifies sample message data and removes expectation of a specific number of messages or internal message structure
  - Replaced code used magic number `2` in several locations, which happened to be the length of the sample message data
  - Refactored code allows for arbitrary sample message data length and message complexity; tests are now independent of the test data
  - Replaced sample message data contained `code: 200` and `code: 400`, which I conflated with expected handler behavior upon review; I thought it best to remove this
- Simplifies message content check.
  - Replaced code checked for specific internal message structure (expecting specific object keys); but internal message structure is not relevant to these tests


jasonburkert-ccpo:
### New test output
```
 PASS  api/provision/consume-provisioning-job.test.ts
  Consume Provisioning Job handler - normal SQS behavior
    ✓ should poll non-empty queue and return 200 status and expected number of messages (5 ms)
    ✓ should poll empty queue and return 200 status and no messages (4 ms)
  Consume Provisioning Job handler - abnormal SQS behavior
    ✓ should poll an abnormally behaving queue and return 500 status (10 ms)
```

### Old test output
```
 PASS  api/provision/consume-provisioning-job.test.ts (5.899 s)
  Consumer Provisioning Job handler valid behavior
    ✓ should poll 2 messages from SQS provisioning queue with 200 (9 ms)
    ✓ should poll 0 messages and return empty array with 200 (2 ms)
  Consumer Provisioning Job invalid behavior
    ✓ should throw an internal error if invalid response from queue (11 ms)
```

jeffsegal-ccpo (approved):

PR #1568, opened 2023-03-14
AT-8767: Gap filling

jeffsegal-ccpo:
During initial integration testing, I noticed that a few things were missing from the object model, so I made the following changes:

- Added classification_level and type to Clin object
- Renamed CLIN and TO number fields to align with the ATAT spec


api/client/types.ts
```diff
@@ -205,6 +210,56 @@ export interface GetProvisioningStatusResponse extends AtatResponse {
...
+  administrators: Array<Administrator>;
+}
+
+export interface HothProvisionRequest {
+  jobId: string;
+  userId: string;
+  portfolioId?: string;
+  operationType: ProvisionRequestType;
+  targetCspName: string;
+  payload: NewPortfolioPayload | NewEnvironmentPayload | NewTaskOrderPayload | AdministratorPayload;
```
jasonburkert-ccpo, line 247: It's trivial but is there any reason `AdministratorPayload` doesn't follow the _NewX_ convention?

jasonburkert-ccpo (approved):

PR #1576, opened 2023-03-22
AT-8895: HOTH API: Encode path parameters in CSP requests and update version number header to remove "v"

jeffsegal-ccpo:
- Fixed version number header
- Updated client.ts to encode path parameters (not entire URL or path portion)

### DIFF
--- a/api/client/client.ts
+++ b/api/client/client.ts
@@ -8,6 +8,7 @@ import MockAdapter from "axios-mock-adapter";
 import {
   addEnvironmentRequest,
   CSP_A_TEST_ENDPOINT,
+  CSP_B_STATUS_ENDPOINT,
   CSP_B_TEST_ENDPOINT,
   cspAAddPortfolioRequest,
   TEST_ENVIRONMENT_ID,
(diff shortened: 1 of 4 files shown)


api/client/client.ts
```diff
@@ -100,7 +101,7 @@ export class AtatClient implements IAtatClient {
    * The version of the ATAT CSP API specification that is supported by this version
    * of the client. This should match the `info.version` from the OpenAPI specification.
    */
-  public static supportedApiVersion = "v1.1.0";
+  public static supportedApiVersion = "1.1.0";
```
jasonburkert-ccpo, line 104: 👍

jasonburkert-ccpo: Agrees with [version: 1.1.0](https://github.com/dod-ccpo/atat-csp-orchestration/blob/e0988bc14320befe6a2da6a04120cb8e788c9533/provisioning/atat_provisioning.yaml#L29)

jasonburkert-ccpo (approved):

PR #1586, opened 2023-03-31
Feature/AT-8729

jeffsegal-ccpo:
Added support for building the Table in Section 8 of the Description of Work. Includes a large number of fixes to the TypeScript interfaces in support of ensuring that sample data from SNOW aligns to them. Most of the work consisted of adding convenience functions for calculating values used in the actual DoW docx template so the data can be easily displayed there.

A few snippets from the docx template:

image
image
image


jasonburkert-ccpo (approved):

document-generation/utils/dow.test.ts
```diff
@@ -263,3 +274,115 @@ describe("getCDRLs", () => {
     expect(cdrls).toHaveLength(expectedCdrls.length);
   });
 });
+
+describe("securityRequirements", () => {
```
jasonburkert-ccpo, line 278: Yay for new tests!

jeffsegal-ccpo:
Bypassing merge requirements because we're only slightly under (~2%) and the lines in question are actually old, not new code.

PR #1592, opened 2023-04-06
updated the local dependencies & added IAM user monitoring cf template

zachmcclurg-ccpo:
- updates many local dependencies
  - including typescript to 5.0.3
- downgrades parts of middy to 3.6.2, so all are on that version
- adds IAM user monitoring CloudFormation template


package.json
```diff
@@ -13,63 +13,62 @@
...
+    "@aws-sdk/client-secrets-manager": "^3.306.0",
+    "@aws-sdk/client-sfn": "^3.306.0",
+    "@aws-sdk/client-sqs": "^3.306.0",
+    "@aws-sdk/client-sts": "^3.306.0",
     "@middy/core": "^3.6.2",
-    "@middy/error-logger": "^4.2.7",
+    "@middy/error-logger": "^3.6.2",
     "@middy/http-json-body-parser": "^3.6.2",
-    "@middy/input-output-logger": "^4.2.7",
+    "@middy/input-output-logger": "^3.6.2",
```
jasonburkert-ccpo, line 62: This looks suspect.  Appears to downgrade middy which we just for JSON object validation.

jasonburkert-ccpo (changes requested):
Want to look more closely.  Find the change to middy surprising.

cfn-templates/sso-user-account-activity.yaml
```diff
@@ -41,6 +41,8 @@ Resources:
       EventPattern: 
         detail-type:
           - "AWS API Call via CloudTrail"
+        source:
```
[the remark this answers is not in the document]
zachmcclurg-ccpo: added the source

package.json
zachmcclurg-ccpo: its best to keep middy all the same version. and there were some issues with the latest middy release so we couldn't update them all with out risk of issue.

jasonburkert-ccpo: after looking more closely I understand our current config to have mixed versions, and this PR downgrades add middy libraries to the same version
image

package.json
```diff
@@ -13,63 +13,62 @@
...
     "eslint-plugin-promise": "^6.1.1",
-    "jest": "^29.4.3",
-    "prettier": "^2.8.4",
-    "ts-jest": "^29.0.5",
+    "jest": "^29.5.0",
+    "prettier": "^2.8.7",
+    "ts-jest": "^29.1.0",
     "ts-node": "^10.9.1",
-    "typescript": "~4.9.5"
+    "typescript": "~5.0.3"
```
jasonburkert-ccpo, line 46: This is major version upgrade.  I wonder what the potential impact is.  Maybe it's no big deal.

jasonburkert-ccpo:
Closing PR.  Suggest the following to isolate unrelated changes:
 
- PR to add monitoring CloudFormation template
- PR to downgrade some middy packages to 3.6.2 so all are on same version - [middy 4.x high encourages NodeJS v18](https://middy.js.org/docs/upgrade/3-4/), so I'll add a tech debt ticket for that with appropriate linkage and blocking
- PR to upgrade typescript from major version 4.x to 5.x - 5.x has been available now for not quite two months, will make decision
- PR to upgrade everything else

PR #1594, opened 2023-04-07
AT-8882 Update logic for architectural design

luakim-ccpo:
Removed current env arch design req section
Added additional logic to check for required fields in UI
Updated section to match the rest of the document


luakim-ccpo:
image

jeffsegal-ccpo (approved):

PR #1600, opened 2023-04-10
AT-8929: DoW - More updates to Security Requirements table (Section 8)

jeffsegal-ccpo:
- Updated definition of `isSecurityNeeded`, which drives display of the table in Section 8 of the DoW. 
- Modified Clearance Level & Justification columns for rows 2, 4 and 6.
- Modified Clearance Level column for rows 8, 10, 12, 14, 16 and 18.

Ticket: AT-8929


document-generation/utils/utils.ts
```diff
@@ -84,35 +84,38 @@ interface PDFTemplateFiles {
...
-  [DocumentType.INCREMENTAL_FUNDING_PLAN]: {
-    docx: "/opt/ifp-template.docx",
-  },
-  [DocumentType.EVALUATION_PLAN]: {
-    docx: "/opt/eval-plan-template.docx",
-  },
-  [DocumentType.REQUIREMENTS_CHECKLIST]: {
-    docx: "/opt/requirements-checklist-template.docx",
-  },
+export const getDocumentTemplatePath = (folder?: string): TemplatePaths => {
```
jasonburkert-ccpo, line 87: Nice

jasonburkert-ccpo (approved):

document-generation/utils/utils.ts
jeffsegal-ccpo: I have some code that I use for local testing that I don't want to commit as actual code but might just throw in the README later, but it uses this method and passes in its own folder.

PR #1615, opened 2023-04-24
AT-8940: Added U_TO_S and S_TO_U CDS entries to DoW template

jeffsegal-ccpo:
Added missing rows in CDS Table:

image

Which yields:

image


jasonburkert-ccpo (approved):
Great show and tell in this PR

PR #1619, opened 2023-04-26
Feature/at 8973

zachmcclurg-ccpo:
this adds a "Classification:Unclass" tag to the load balancer access log bucket via the cdk code.

### DIFF
--- a/lib/atat-web-api-stack.ts
+++ b/lib/atat-web-api-stack.ts
@@ -115,7 +115,7 @@ export class AtatWebApiStack extends cdk.Stack {
       objectLockEnabled: true,
       objectLockDefaultRetention: s3.ObjectLockRetention.compliance(cdk.Duration.days(365)),
     });
-    Tags.of(accessLogsBucket).add("Classification", "UNCLASSIFIED");
+    Tags.of(accessLogsBucket).add("Classification", "UNCLASS");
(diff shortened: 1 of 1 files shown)


lib/atat-web-api-stack.ts
```diff
@@ -114,6 +115,7 @@ export class AtatWebApiStack extends cdk.Stack {
       objectLockEnabled: true,
       objectLockDefaultRetention: s3.ObjectLockRetention.compliance(cdk.Duration.days(365)),
     });
+    Tags.of(accessLogsBucket).add("Classification", "UNCLASS");
```
jeffsegal-ccpo, line 118: Are these tags expected by C5ISR? Can we create an enum class that briefly describes how/why they're used?

jeffsegal-ccpo (changes requested):

lib/atat-web-api-stack.ts
zachmcclurg-ccpo: This is used for STIG compliance. Not for C5ISR. this is being applied to Cloud Trail and the cloud trail s3 bucket and the load balancer access log bucket.

zachmcclurg-ccpo: working on the enum now...

jeffsegal-ccpo (approved):

PR #1620, opened 2023-04-26
FIX: Don't alter contracting office fee used in IGCE docgen

jasonburkert-ccpo:
Fixes a latent bug that was made worse by #1614.  That PR spread the bug to DITCO calculations in addition to impacting non-DITCO calculations.

- Removes altering of contracting office fee to two fixed digits; had been negatively impacting fee calculation
  - use of `Number.toFixed()` returns a string which was then placed directly in formula of the generated XLSX file
  - for example: a fee of `3.55%` would be translated to decimal `0.0355`, then process by `Number.toFixed()` resulting in `0.03` which was then used in the fee calculation; incorrectly using `3%` rather than `3.55%`
- Changes fee percentage calculations for similarity with corresponding calculations in https://github.com/dod-ccpo/atat-snow/pull/360/
- Renames variables containing exceljs.Cell objects for clarity

### DIFF
--- a/document-generation/igce-document.ts
+++ b/document-generation/igce-document.ts
@@ -242,17 +242,20 @@ export async function generateIGCEDocument(
   summaryFundingDocCell.value = fundingDocumentNumber;
 
   // Surge Capabilities
+  const surgeFeePct = payload.surgeCapabilities * 0.01;
   const surgeCapabilities = summarySheet.getCell("A23");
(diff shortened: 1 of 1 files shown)


jeffsegal-ccpo:
Wish we had built UTs for this stuff initially - it would have caught this and made it easier for us to maintain everything. Alas.

PR #1627, opened 2023-05-04
AT-8831: First cut of J&A .docx template

jeffsegal-ccpo:
Includes all fields necessary to generate a J&A document with a few notable exceptions (due to the SNOW data model still being in flux):

* Type of funds (O&M, RDT&E, etc.)
* Procurement history
* POC Organizations

It also intentionally leaves a few DITCO inputs in red text.

A few snippets:

image
image
image
image


luakim-ccpo (approved):
Reviewed J&A template and found no issues

PR #1635, opened 2023-05-10
Adjust docgen to support many prior contracts

tomarnold-ccpo:
updated contractInformation to support an array of contracts


jasonburkert-ccpo:
All of this looks appropriate from quick inspection

PR #1687, opened 2023-06-07
AT-9027-Document-styling-change

luakim-ccpo:
DoW:
- remove extra spaces and adjust alignment 
- add underline 
- minor change for following conditions:
- * no travel
- * no packaging/shipping
- * no surge 

Req Checklist:
- remove extra spaces and adjust alignment
- make invisible table for signature block
- changed mission owner to mission partner


luakim-ccpo:
DOW change:
highlighted paragraph is only need it when there is surge
image

Removed highlighted sentence based on ditco request
image

When there is no travel, we have to show text "Reserved" and hide the travel detail table
image

Add underline for instance title 
image

no package/shipping (rather than showing as a bullet list, inline with package title
image

PR #1723, opened 2023-06-29
adjust template adjust to fix 500 response error

luakim-ccpo:
Adjust so that all eval paln info is wrapped in side of {IF !exceptionToFairOpportunity}  statement. As a business requirement, you cannot have method/sourceSelection when the package is exception to fair opportunity. Currently eval memo doc gen is returning 500 error response because it's expecting method/sourceSelection data even through package is exception to the fair opportunity. 
This adjustment will enhance the robustness of the evaluation memo document template. As the user modifies the option (exception to fair opportunity vs evaluation plan), it ensures the accurate display of relevant content.


luakim-ccpo:
image

torinharthcock-ccpo (approved):
Code changes pass document tests, changes to document template appear to work correctly.

PR #1727, opened 2023-06-30
Change Eval Plan validation model to allow 'null' for property 'sourceSelection'

jasonburkert-ccpo:
The Eval Plan model is extended by that for Eval Memo.  This may resolve validation errors during docgen when there is no `sourceSelection` value.  Property `method` already contains this null option and it doesn't seem to cause the same validation error.

Ticket: AT-9120

### DIFF
--- a/models/document-generation.ts
+++ b/models/document-generation.ts
@@ -710,9 +710,11 @@ export const evalPlan = {
         SourceSelection.TECH_PROPOSAL,
         SourceSelection.SET_LUMP_SUM,
         SourceSelection.EQUAL_SET_LUMP_SUM,
+        null,
       ],
     },
     method: {
+      type: "string",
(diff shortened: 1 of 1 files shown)


models/document-generation.ts
```diff
@@ -710,9 +710,11 @@ export const evalPlan = {
         SourceSelection.TECH_PROPOSAL,
         SourceSelection.SET_LUMP_SUM,
         SourceSelection.EQUAL_SET_LUMP_SUM,
+        null,
```
jasonburkert-ccpo, line 713: This should fix the validation errors for Eval Memo docgen

models/document-generation.ts
```diff
@@ -710,9 +710,11 @@ export const evalPlan = {
...
       ],
     },
     method: {
+      type: "string",
```
jasonburkert-ccpo, line 717: This just seemed to be missing

luakim-ccpo (approved):

PR #1728, opened 2023-06-30
At 9007/additional IGCE template change

luakim-ccpo:
Based on comment from the functional testing, adjusted IGCE template.

This changes fix all the comments listed below:
image


luakim-ccpo:
Added If statement to only show $ when there is external ordering agency fee
image

Adjusted the cell type to be accounting to show $ sign when there is contracting office fee

Adjusted the column width size to display the full text
image

jasonburkert-ccpo (approved):
Looks reasonable

PR #1731, opened 2023-06-30
AT-9007 - Fixed ditco fee returning undefined

luakim-ccpo:
With recent change, DITCO user will have empty object for row 25 (External agency ordering fee)

Thus, formula `=(K24 + K25) * ${contractingShopFeePct}` will retrun underifned since K25 is null. 
image

Use the If statement to properly display the contracting office fee
image

### DIFF
--- a/document-generation/igce-document.ts
+++ b/document-generation/igce-document.ts
@@ -262,7 +262,12 @@ export async function generateIGCEDocument(
   const contractingShopFeePct = payload.contractingShop.fee * 0.01;
   if (contractingShopFeePct) {
     const contractingShopFeeCell = summarySheet.getCell("C27");
-    contractingShopFeeCell.value = { formula: `=(K24 + K25) * ${contractingShopFeePct}`, date1904: false };
+    if (payload.contractingShop.name !== "DITCO") {
+      contractingShopFeeCell.value = { formula: `=(K24 + K25) * ${contractingShopFeePct}`, date1904: false };
+    }
+    if (payload.contractingShop.name === "DITCO") {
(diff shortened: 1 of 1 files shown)


document-generation/igce-document.ts
```diff
@@ -262,7 +262,12 @@ export async function generateIGCEDocument(
   const contractingShopFeePct = payload.contractingShop.fee * 0.01;
   if (contractingShopFeePct) {
     const contractingShopFeeCell = summarySheet.getCell("C27");
-    contractingShopFeeCell.value = { formula: `=(K24 + K25) * ${contractingShopFeePct}`, date1904: false };
+    if (payload.contractingShop.name !== "DITCO") {
+      contractingShopFeeCell.value = { formula: `=(K24 + K25) * ${contractingShopFeePct}`, date1904: false };
+    }
+    if (payload.contractingShop.name === "DITCO") {
+      contractingShopFeeCell.value = { formula: `=K24 * ${contractingShopFeePct}`, date1904: false };
+    }
```
AndrewNickerl-ccpo, line 270: Little bit of a nitpick but from a code structure standpoint this could be simplified into
`if (payload.contractingShop.name === "DITCO") {
      contractingShopFeeCell.value = { formula: `=K24 * ${contractingShopFeePct}`, date1904: false };
    } else {
      contractingShopFeeCell.value = { formula: `=(K24 + K25) * ${contractingShopFeePct}`, date1904: false };
    }`

AndrewNickerl-ccpo: Whew that code did not come out pretty in my comment. Basically my point is that for efficiency and readability you can replace the two `if` blocks with an if/else

AndrewNickerl-ccpo (approved):

PR #1733, opened 2023-06-30
changed if/else to ternary operation

luakim-ccpo:
It seem like IGCE doc template tool does not support if/else statement. Looking at other IGCE codes, it uses ternary operation instead of if/else statement.


torinharthcock-ccpo (approved):
Code is equivalent, just changed if/else to ternary. Tests still pass.

PR #1740, opened 2023-07-04
Create SNS notification for pipeline status

zachmcclurg-ccpo:
created a sns topic that notifies atat distro group on when the pipeline is started and when the pipeline is finished.


zachmcclurg-ccpo:
restarting this pr

PR #1742, opened 2023-07-05
Eval Memo improvements

jasonburkert-ccpo:
- adds additional sample payloads and unit tests for Eval Memo (EM) using the two most common variations - with and without exception to fair opportunity.
  - The replaced singular sample payload had a (what I now understand to be) nonsensical combination with `exceptionToFairOpportunity`=_YES_ and values for both `sourceSelection` and `method` which come from Eval Plan (EP)
- adds unit tests for generate-document lambda for EM using the two new sample payloads
- corrects schema models for EP and EM that we pass to [middy](https://middy.js.org/) for validation during docgen
  - this change allowed both of the new unit tests for the generate-document lambda to pass
  - I believe this effectively isolated the validation errors we've seen in DSP Dev environment
  - I witnessed similar oneOf validation errors with index 4 before correcting
  - The then-failing test agrees with behavior that @palaparthi-ccpo has documented during testing, only failed when `exceptionToFairOpportunity`=_YES_ and we won't expect to have an EP record, thus all of those inherited properties will be null or empty array.  The middy validation must be configured to allow this for successful docgen.

### Expanded unit tests for EM document
Prior sample data was at odds with business rules.  These two configurations are more representative of expected combinations.  Had to loosen type model for EP to pass the test (due to model inheritance).

image

The two latter tests generate these files.  To my eye these files contain the desired generated content.

image

[evalmemo-withException.docx](https://github.com/dod-ccpo/atat-web-api/files/11962589/evalmemo-withException.docx)
[evalmemo-withoutException.docx](https://github.com/dod-ccpo/atat-web-api/files/11962590/evalmemo-withoutException.docx)

### Added unit test for generate-document lambda
When first added, one of these tests passed, and the other failed (the w/ Exception to Fair Opportunity flavor).  I corrected errors in the schema models we pass to middy for validation during docgen.  Now both tests pass.

image

Ticket: AT-9120

### DIFF
--- a/document-generation/generate-document.test.ts
+++ b/document-generation/generate-document.test.ts
@@ -16,7 +16,8 @@ import {
   sampleRequirementsChecklistRequest,
   sampleJustificationAndApprovalRequest,
   sampleMarketResearchReportRequest,
-  sampleEvalMemoRequest,
+  sampleEvalMemoRequestWithException,
+  sampleEvalMemoRequestWithoutException,
 } from "./utils/sampleTestData";
 
 const validRequest = {
@@ -219,12 +220,27 @@ describe("Successful generate-document handler", () => {
     expect(response.headers).toEqual(docHeaders.mrr);
   });
 
-  it("should return successful Eval Memo document response", async () => {
+  it("should return successful Eval Memo document response - w/ Exception to Fair Opportunity", async () => {
     // GIVEN / ARRANGE
     const request = {
       ...validRequest,
       body: JSON.stringify({
-        ...sampleEvalMemoRequest,
+        ...sampleEvalMemoRequestWithException,
+      }),
+    };
+
+    // WHEN / ACT
+    const response = await handler(request, {} as Context);
+    // THEN / ASSERT
+    expect(response).toBeInstanceOf(SuccessBase64Response);
+    expect(response.headers).toEqual(docHeaders.evalMemo);
+  });
+  it("should return successful Eval Memo document response - w/o Exception to Fair Opportunity", async () => {
+    // GIVEN / ARRANGE
+    const request = {
+      ...validRequest,
+      body: JSON.stringify({
+        ...sampleEvalMemoRequestWithoutException,
       }),
     };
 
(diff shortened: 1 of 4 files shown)


jasonburkert-ccpo:
### New sample data / unit test  emulates EvalMemo script include behavior
This new sample data
image
looks like the ServiceNow EvalMemo script include behavior, with three nullable properties followed by four empty arrays
image

Had to make these properties nullable to satisfy TypeScript and get the test using this sample data to pass
image

document-generation/generate-document.test.ts
```diff
@@ -219,12 +220,27 @@ describe("Successful generate-document handler", () => {
     expect(response.headers).toEqual(docHeaders.mrr);
   });
 
-  it("should return successful Eval Memo document response", async () => {
+  it("should return successful Eval Memo document response - w/ Exception to Fair Opportunity", async () => {
```
jasonburkert-ccpo, line 223: This unit test was failing after I added it (but the other -- w/o Exception -- was passing).

jasonburkert-ccpo: Believe I fixed it by correcting EM and EP schema models used by middy for validation when docgen lambda executes
https://github.com/dod-ccpo/atat-web-api/pull/1742/commits/83979dced2b31c729dc05e54e91c5a865aff5f8c

document-generation/utils/sampleTestData.ts
```diff
@@ -1807,14 +1807,31 @@ export const sampleMarketResearchReportRequest = {
...
+    // EP properties below
+    taskOrderTitle: null,
+    sourceSelection: null,
+    method: null,
+    standardSpecifications: [],
+    customSpecifications: [],
+    standardDifferentiators: [],
+    customDifferentiators: [],
+  },
+};
```
jasonburkert-ccpo, line 1826: New sample payload number 1

document-generation/utils/sampleTestData.ts
```diff
@@ -1807,14 +1807,31 @@ export const sampleMarketResearchReportRequest = {
...
+
+export const sampleEvalMemoRequestWithoutException = {
```
jasonburkert-ccpo, line 1828: New sample payload number 2

document-generation/utils/sampleTestData.ts
```diff
@@ -1807,14 +1807,31 @@ export const sampleMarketResearchReportRequest = {
...
+  ...sampleEvalMemoRequestWithException,
+  templatePayload: {
+    ...sampleEvalMemoRequestWithException.templatePayload,
+    exceptionToFairOpportunity: false, // EP generated, expect EP record
+    // EP properties below
+    taskOrderTitle: "Amazing Task Order",
```
jasonburkert-ccpo, line 1834: New sample payload number 2

luakim-ccpo (approved):

PR #1758, opened 2023-07-18
AT-9431: Removing xss middleware

jeffsegal-ccpo:
This PR was precipitated by the docgen process failing for certain packages. After deeper investigation, we found that the existence of certain characters in the docgen payload, such as `:`, was causing the entire payload to get mangled. Local debugging led us to the `xssSanitizer` Middy middleware, which was both:
* Incorrectly written in that it yielded malformed JSON in some cases, and 
* Not serving any security purpose, as no browser content is being served from the HOTH API

As such, we are removing it here. Further, we verified that the Vue feature described at https://vuejs.org/guide/best-practices/security.html#what-vue-does-to-protect-you is in fact what we truly need, as it will escape XSS content in user input and prevent it from being executed by client browsers.


jaynewlin-ccpo:
Not providing a formal review of code per se, but chiming in to affirm that we've discussed the impact of this change on our security posture, and Jeff's PR description indicates why there is no change to security posture as a result of this code change. 👍 from me on that standpoint

PR #1783, opened 2023-08-11
updated the aws-cdk and aws-cdk-lib to 2.91

zachmcclurg-ccpo:


jeffsegal-ccpo (approved):
Working in a sandbox env, so approving and merging to test in Dev env.

PR #1787, opened 2023-08-16
added sso viewing permissions for the Amanda

zachmcclurg-ccpo:
added permissions to the sso-permission-set template to allow maanda to view the IAM SSO roles


jessepeirce-ccpo (approved):
looks good, approved

PR #1805, opened 2023-08-31
AT-8454: Loosened event schema

jeffsegal-ccpo:
Setting `additionalProperties` to `true` for the wrapper event schemas


xanderkeele-ccpo (approved):
Approved

PR #1810, opened 2023-09-05
AT-9692: Implement PUT /portfolios/{portfolioId}/task-orders/{taskOrderId}

xanderkeele-ccpo:
Added all code necessary to implement `updateTaskOrder` functionality. All tests pass. Cleaned up some linter problems also.


xanderkeele-ccpo:
Tried to get 80% coverage, but the holdup is testing error conditions that are going to be very difficult, such as causing the code to throw an undefined exception or an exception that isn't of type `AtatApiError`.

Without those negative tests, I can't get above 76.3% coverage.

jeffsegal-ccpo (approved):

PR #1818, opened 2023-09-13
AT-9751: Implement GetPortfolio in HOTH Api

xanderkeele-ccpo:
Created all step-fn code for AWS functions
Created Tests and mocks for the client
Added handling for getting the portfolio based on Id from the CSP.


jeffsegal-ccpo (approved):
We'll address the copypasta in follow-on tickets very soon

PR #1881, opened 2023-10-27
AT-9971 types/lambda updated with three new optional fields

stephenhayes-ccpo:
types/lambda updated with three new optional fields


jeffsegal-ccpo (approved):
Looks good now.

jeffsegal-ccpo (approved):

PR #1884, opened 2023-11-01
Updating chromium to 118.0.0

jeffsegal-ccpo:
Manual attempt at figuring out why sonar scanning is breaking for dependabot PRs


jeffsegal-ccpo:
Closing as this is OBE now that we fixed the dependabot issue

